PR #22, opened 2016-12-22
[PR] One line update

csauvage:
Updating doc in order to update both JS & PHP dependencies in one command.


tgalopin:
Can you write a more explicit commit message? There is no need to add PR ;) .

Thanks by advance!

PR #25, opened 2016-12-24
Adding ESLint (ES 2015 + React) Support & updating Webpack config file

csauvage:
I've added few packages for multiple purposes : 
-  `.eslintrc` file is the default configuration file I use on a daily based when developing with React 
- I've updated `package.json` file with task lint & lint:watch. 
- I've also made some slight fix on webpack config file in order to display useful informations in terminal. 

Merry Xmas 🎅🏻


.eslintrc
```diff
@@ -0,0 +1,69 @@
...
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "jsx": true
+    }
+  },
+  "env": {
+    "es6": true,
+    "browser": true,
+    "node": true,
+    "jquery": true,
```
tgalopin: You can remove this line, we won't use jQuery.

.eslintrc
```diff
@@ -0,0 +1,69 @@
...
+    "react/no-unknown-property": 1,
+    "react/prefer-es6-class": 1,
+    "react/prop-types": 2,
+    "react/react-in-jsx-scope": 1,
+    "react/require-extension": 1,
+    "react/self-closing-comp": 1,
+    "react/sort-comp": 1,
+    "react/wrap-multilines": 1
+  }
+}
```
tgalopin: This file is missing an ending line.

.gitignore
```diff
@@ -20,3 +20,4 @@
 /node_modules/
 /.tmp
 .DS_Store
+/.idea
```
tgalopin: You shouldn't add this here but globally on your system. If we did here, we would have to add every IDE.

webpack.config.dev.js
```diff
@@ -4,6 +4,7 @@ const path = require('path');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 module.exports = {
+    target: 'web',
```
tgalopin: Not needed, it's the default value.

package.json
```diff
@@ -27,6 +29,24 @@
     "sass-loader": "^4.0.0",
     "style-loader": "^0.13.1",
     "webpack": "^2.1.0-beta.4",
-    "webpack-md5-hash": "^0.0.5"
+    "webpack-md5-hash": "^0.0.5",
+    "toastr": "2.1.2"
+  },
+  "devDependencies": {
```
tgalopin: Can you explain a bit why you add all these dependencies? From my perspective, eslint and its plugins is enough to lint.

.eslintrc
```diff
@@ -0,0 +1,69 @@
...
+  },
+  "env": {
+    "es6": true,
+    "browser": true,
+    "node": true,
+    "jquery": true,
+    "mocha": true
+  },
+  "rules": {
+    "quotes": 0,
```
tgalopin: Does `0` means `off`? If it does, can you use `off` instead?

I will review these rules asap, I don't have much time right now :) .

csauvage: 0 means no alerts
1 means warning
2 means errors. 

I'd like to keep this semantic, firstable for the file weight `"off"` is 4 bytes `0` is one, and then for the community as I never saw an "off" online, moreover, any programmer knows that `0 == false` or in our case silenced rule 😉

package.json
tgalopin: I'm OK with all this but toastr: is it that important? I don't think it's worth adding it.

webpack.config.dev.js
tgalopin: This is still removable :) .

.eslintrc
```diff
@@ -0,0 +1,68 @@
...
+      "jsx": true
+    }
+  },
+  "env": {
+    "es6": true,
+    "browser": true,
+    "node": true,
+    "mocha": true
+  },
+  "rules": {
```
tgalopin: Instead of using your rules (which is quite personnal in my opinion :) ), I think we should stick to a standard: can you add the AirBnB eslintrc configuration (https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb) instead of your own?

PR #56, opened 2017-01-06
Force PHP 7.0 as the minimum required runtime version.

hhamon:


composer.json
```diff
@@ -18,7 +18,7 @@
         }
     },
     "require": {
-        "php": ">=5.5.9",
+        "php": ">=7.0",
```
HeahDude: I'd rather like having the same version for everyone, what about `"php": "7.0"`?

hhamon: Make sense!

PR #77, opened 2017-01-11
Added unit tests for the Adherent and AdherentFactory classes.

hhamon:


src/AppBundle/Membership/MembershipRequest.php
```diff
@@ -128,6 +128,16 @@ public static function createWithCaptcha(string $recaptchaAnswer = null)
         return $dto;
     }
 
+    public function setEmailAddress($emailAddress)
```
salahm: (string $emailAddress)

hhamon: Thanks.

src/AppBundle/Entity/Adherent.php
```diff
@@ -176,4 +176,54 @@ public function getUsername(): string
...
+    }
+
+    public function getLastName()
+    {
+        return $this->lastName;
+    }
+
+    public function getBirthdate()
+    {
+        return $this->birthdate;
```
HeahDude: I don't think it's needed to clone the instance [here](https://github.com/EnMarche/en-marche.fr/pull/77/files#diff-78fe6f3e331a599f3c301c3ee3626789R27) since this getter still returns a mutable object, what about using `\DatetimeImmutable` for this property?

src/AppBundle/Membership/MembershipRequest.php
```diff
@@ -120,14 +120,24 @@ public function __construct()
...
     {
         $dto = new self();
         $dto->recaptcha = $recaptchaAnswer;
 
         return $dto;
     }
 
+    public function setEmailAddress(string $emailAddress)
+    {
+        $this->emailAddress = mb_strtolower($emailAddress);
```
tgalopin: Are we sure all mail providers are case insensitive?

HeahDude: @tgalopin I think so, it's more about the protocol, see the first result I got for searching "e-mail lower case" https://www.lifewire.com/are-email-addresses-case-sensitive-1171111.

tgalopin: Can we still save the original data, just in case? It does not cost much IMO.

HeahDude: Sounds pointless to me, why for?

tgalopin: Okay, let's do it like this.

hhamon: The local part (the string before @) is case sensitive but not the domain name.

```
The local-part of a mailbox MUST BE treated as case sensitive. Therefore, SMTP implementations MUST take care to preserve the case of mailbox local-parts. In particular, for some hosts, the user "smith" is different from the user "Smith". However, exploiting the case sensitivity of mailbox local-parts impedes interoperability and is discouraged. Mailbox domains follow normal DNS rules and are hence not case sensitive.
```

hhamon: So I think we need to store the canonical email address in a separate field when checking for the uniqueness of the user account.

PR #93, opened 2017-01-17
Implement administration security firewall

tgalopin:


src/AppBundle/Admin/AdministratorAdmin.php
```diff
@@ -0,0 +1,102 @@
...
+        parent::preUpdate($admin);
+
+        if ($admin->plainPassword) {
+            $admin->setPassword($this->encoders->getEncoder($admin)->encodePassword($admin->plainPassword, null));
+        }
+    }
+
+    protected function configureFormFields(FormMapper $formMapper)
+    {
+        $isCreation = $this->getSubject()->getId() === null;
```
hhamon: Yoda style.

src/AppBundle/Entity/Administrator.php
```diff
@@ -0,0 +1,155 @@
...
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+
+    /**
+     * @var string|null
+     *
+     * @ORM\Column
+     *
+     * @Assert\Email()
```
hhamon: Empty parenthesis can be removed.

src/AppBundle/Entity/Administrator.php
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * @param null|string $emailAddress
+     *
+     * @return Administrator
+     */
+    public function setEmailAddress($emailAddress): Administrator
+    {
+        $this->emailAddress = $emailAddress;
+
+        return $this;
```
hhamon: I'm not a big fan of returning $this in an entity setter.

src/AppBundle/Admin/AdministratorAdmin.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $formMapper
+            ->add('emailAddress', null, [
```
HeahDude: Should be `EmailType::class`.

src/AppBundle/Entity/Administrator.php
```diff
@@ -0,0 +1,155 @@
...
+    private $password;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column
+     *
+     * @Assert\NotBlank()
+     */
+    private $role;
```
HeahDude: Are you sure an admin won't have many roles?

src/AppBundle/Form/LoginType.php
```diff
@@ -53,6 +53,6 @@ public function configureOptions(OptionsResolver $resolver)
      */
     public function getBlockPrefix()
     {
-        return 'app_login';
+        return 'app_admin_login';
```
HeahDude: Should be reverted.

src/AppBundle/Admin/AdministratorAdmin.php
```diff
@@ -0,0 +1,102 @@
...
+    private $encoders;
+
+    /**
+     * @param Administrator $admin
+     */
+    public function prePersist($admin)
+    {
+        parent::prePersist($admin);
+
+        $admin->setPassword($this->encoders->getEncoder($admin)->encodePassword($admin->plainPassword, null));
```
HeahDude: I don't think it's a good idea to have a public property for the plain password, it may change in this listener at any time.

The password should be encoded in the factory or a manager only. And this listener should only force `null` on this public property, not encode it.

What do you think? /cc @hhamon

src/AppBundle/Entity/Administrator.php
tgalopin: Yes, we will use the roles hierarchy.

src/AppBundle/Form/LoginType.php
tgalopin: Indeed

src/AppBundle/Entity/Administrator.php
HeahDude: You should allow many roles here though.

src/AppBundle/Admin/AdministratorAdmin.php
tgalopin: I'm not sure to understand what you mean, and how I could integrate this in Sonata. Could you give me more informations?

HeahDude: Ok, I  understand. Still, having this public property being changed is enough to persist a new encoded password, it is not very safe, right?

The best way IMHO is to use a model transformer instead of both listener:
```php
protected function configureFormFields(FormMapper $formMapper)
    {
        $isCreation = null === $this->getSubject()->getId();

        $formMapper
            // ...
            ->add(
                // Use the encoded property for mapping, even get rid of the plain one
                $this->create('password', RepeatedType::class, [
                    'first_options' => [
                        'label' => 'Mot de passe',
                    ],
                    'second_options' => [
                        'label' => 'Confirmation',
                    ],
                    'type' => PasswordType::class,
                    'required' => $isCreation,
                ])
                ->addModelTransformer(new CallbackTransformer(
                    function ($encoded) {
                        return ''; // or ****
                    },
                    function ($plain) {
                        return is_string($plain) && '' !== $plain
                            ? $this->encoders->getEncoder($admin)->encodePassword($admin->plainPassword, null)
                            : null
                    ;
            )
    ;
}

File: src/AppBundle/Form/LoginType.php (hunk 6 above)
Comment (HeahDude) [inline src/AppBundle/Form/LoginType.php]: If you need to customize the form name (html attribute and form/data tree key), to ease theming, you can use `block_name` option from the controller, see http://symfony.com/doc/current/reference/forms/types/form.html#block-name.

File: src/AppBundle/Admin/AdministratorAdmin.php (hunk 7 above)
Comment (tgalopin) [inline src/AppBundle/Admin/AdministratorAdmin.php]: I understand what you mean, that's a good idea. I'll change this.

**Comment by tgalopin** [comment]
Updated, is ot ok for you @HeahDude ?

File: src/AppBundle/Admin/AdministratorAdmin.php [hunk 8]
```diff
@@ -0,0 +1,96 @@
...
+                    'required' => $isCreation,
+                ])
+                    ->addModelTransformer(new CallbackTransformer(
+                        function ($encoded) {
+                            return ''; // or ****
+                        },
+                        function ($plain) use ($admin) {
+                            return is_string($plain) && '' !== $plain
+                                ? $this->encoders->getEncoder($admin)->encodePassword($plain, null)
+                                : $admin->getPassword();
```
HeahDude: Yeah my example was bad, sorry. 👍

src/AppBundle/Admin/AdministratorAdmin.php
```diff
@@ -0,0 +1,96 @@
...
+                        }
+                    ))
+            );
+    }
+
+    protected function configureDatagridFilters(DatagridMapper $datagridMapper)
+    {
+        $datagridMapper
+            ->add('emailAddress', null, [
```
HeahDude: I don't know Sonata very well, but using `null` instead of an explicite type will trigger guessing and this should be done wisely, even avoided for micro optimization.

src/AppBundle/Entity/Administrator.php
```diff
@@ -0,0 +1,145 @@
...
+     * @ORM\Column
+     *
+     * @Assert\NotBlank
+     */
+    private $role;
+
+    /**
+     * @var string|null
+     */
+    public $plainPassword;
```
HeahDude: Do you still need it?

src/AppBundle/Admin/AdministratorAdmin.php
tgalopin: It's recommended in the documentation and for the admin I don't think it's a big issue. I'll keep it like that for the moment.

PR #95, opened 2017-01-17
Handled integration of registration page

HeahDude:
Closes [EM-9](http://jira.sensiolabs.net/browse/EM-9).

### DIFF
--- a/app/Resources/views/membership/register.html.twig
+++ b/app/Resources/views/membership/register.html.twig
@@ -29,112 +29,132 @@
 
 {% block content %}
     <section class="l__wrapper--medium">
-        <div class="donation">
+        <div class="register">
             <h2>J'adhère au mouvement En Marche !</h2>
 
-            <div class="l__row l__row--top l__row--h-stretch">
-                <div class="donation__form">
-                    <form method="post" id="app-membership" class="form">
-                        {{ form_errors(form) }}
+            <p>
+                En Marche ! est un collectif citoyen qui part d'une conviction : les Français doivent être au cœur
+                de la vie politique et non son décor. Ce collectif n'est pas l'addition d'individus, mais le rassemblement
+                de personnes engagées qui partagent un socle de valeurs.
+                <a href="#" class="color--blue">Télécharger l'intégralité de la charte et des règles de fonctionnement.</a>
+            </p>
 
-                        <div class="form__row">
-                            {{ form_errors(form.gender) }}
-                            {{ form_widget(form.gender) }}
-                        </div>
+            <div class="register__form">
+                {{ form_start(form) }}
+                    {{ form_errors(form) }}
 
-                        <div class="form__row">
-                            {{ form_label(form.firstName, 'Prénom') }}
+                    <div class="form__row">
+                        {{ form_errors(form.gender) }}
+                        {{ form_widget(form.gender, { attr: { class: 'register__form__gender' } }) }}
+                    </div>
+
+                    <div class="form__row l__row">
+                        <div class="l__col form--mid">
                             {{ form_errors(form.firstName) }}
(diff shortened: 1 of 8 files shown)


app/Resources/views/membership/register.html.twig
```diff
@@ -29,112 +29,132 @@
...
-                </div>
+                    <div class="form__row text--center b__nudge--top">
+                        <button type="submit" id="become-adherent" class="btn--green form btn">J'adhère</button>
+                    </div>
 
+                {{ form_end(form) }}
             </div>
+
+            <p class="register--mentions">
+                Les informations que vous nous communiquer sont nécessaires à la gestion de vos dons et de nos
```
hhamon: `/communiquer/communiquez/`

app/Resources/views/membership/register.html.twig
```diff
@@ -29,112 +29,132 @@
...
     <section class="l__wrapper--medium">
-        <div class="donation">
+        <div class="register">
             <h2>J'adhère au mouvement En Marche !</h2>
 
-            <div class="l__row l__row--top l__row--h-stretch">
-                <div class="donation__form">
-                    <form method="post" id="app-membership" class="form">
-                        {{ form_errors(form) }}
+            <p class="register--intro">
```
tgalopin: If you really want to do integration, we use the BEM convention with the following synthax: `block__subblock--variant` :) .

app/Resources/views/membership/register.html.twig
```diff
@@ -29,112 +29,132 @@
...
+                        <button type="submit" id="become-adherent" class="btn--green form btn">J'adhère</button>
+                    </div>
 
+                {{ form_end(form) }}
             </div>
+
+            <p class="register--mentions">
+                Les informations que vous nous communiquer sont nécessaires à la gestion de vos dons et de nos
+                relations. Elles sont exclusivement réservées à l'usage de En Marche et en retournant ce formulaire, vous
+                autorisez celle-ci à utiliser vos donner pour des opérations de communication politique et de dons. Vos
```
hhamon: `/donner/données/`

front/style/_forms.scss
```diff
@@ -45,13 +45,29 @@
         margin-bottom: 15px;
     }
 
+    //&__blue_row {
```
tgalopin: I don't think we should add commented code into master.

app/Resources/views/membership/register.html.twig
HeahDude: Fixed.

app/Resources/views/membership/register.html.twig
HeahDude: Fixed.

app/Resources/views/membership/register.html.twig
HeahDude: Yeah, I don't know about that, and the "real" integration is not my job :)

front/style/_forms.scss
HeahDude: This code is a question (I forgot to ask) because I could not handle blue fields like in your mockup.

So it's up to you, either I leave it as is or just remove it, but I won't go further on the integration work.

app/Resources/views/membership/register.html.twig
tgalopin: So please remove this class, we won't use it.

front/style/_forms.scss
tgalopin: Remove it, we will do the integration ourselves.

app/Resources/views/membership/register.html.twig
HeahDude: Done.

front/style/_forms.scss
HeahDude: Done.

PR #100, opened 2017-01-19
Implement Two Factor Authentication for administrators

tgalopin:
Based on #93.

### DIFF
--- a/src/AppBundle/Admin/AdministratorAdmin.php
+++ b/src/AppBundle/Admin/AdministratorAdmin.php
@@ -3,6 +3,7 @@
 namespace AppBundle\Admin;
 
 use AppBundle\Entity\Administrator;
+use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Google\GoogleAuthenticator;
 use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Datagrid\DatagridMapper;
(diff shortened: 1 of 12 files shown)


src/AppBundle/Admin/AdministratorAdmin.php
```diff
@@ -61,6 +77,12 @@ function ($plain) use ($admin) {
                         }
                     ))
             );
+
+        if (!$isCreation) {
+            $formMapper->add('googleAuthenticatorSecret', null, [
```
HeahDude: Same comment as in the previous PR `null` as second argument is preferable.

src/AppBundle/Controller/Security/AdminSecurityController.php
```diff
@@ -56,4 +57,15 @@ public function loginCheck()
...
     {
     }
+
+    /**
+     * QR-code generator to be used with Google Authenticator.
+     *
+     * @Route("/qr-code/{id}", name="app_admin_qrcode")
+     * @Method("GET")
+     */
+    public function qrcodeAction(Administrator $administrator): Response
```
HeahDude: `qrCodeAction`?

src/AppBundle/Controller/Security/AdminSecurityController.php
```diff
@@ -56,4 +57,15 @@ public function loginCheck()
     public function logoutAction()
     {
     }
+
+    /**
+     * QR-code generator to be used with Google Authenticator.
+     *
+     * @Route("/qr-code/{id}", name="app_admin_qrcode")
```
HeahDude: `app_admin_qr_code`?

src/AppBundle/Entity/Administrator.php
```diff
@@ -44,6 +45,13 @@ class Administrator implements UserInterface
     private $password;
 
     /**
+     * @var string|null
+     *
+     * @ORM\Column(nullable=true)
+     */
+    private $googleAuthenticatorSecret;
```
HeahDude: Does it really need to be persisted, or just to be checked once to trust the user?

Maybe saving a date would be more useful, i.e to handle expiration before asking again.

tgalopin: This is not the code, this is the secret used to generate the code. A bit like a private key. It's stored on the server and on the application only once.

HeahDude: Understood, thanks.

PR #132, opened 2017-01-26
Prevent non geocodable addresses for adherent and committees

hhamon:


app/config/config_prod.yml
```diff
@@ -2,6 +2,10 @@ imports:
     - { resource: config.yml }
     - { resource: services/sentry.xml }
 
+#services:
+#    app.geocoder.cache_adapter:
+#        class: 'Doctrine\Common\Cache\RedisCache'
```
tgalopin: I think we should use the 3.2 Cache component, and I don't like commiting commented code.

PR #135, opened 2017-01-26
Add bash task into Makefile

tgalopin:


Makefile
```diff
@@ -16,6 +16,9 @@ start:          ## Start the Docker containers
 stop:           ## Stop the Docker containers
 	docker-compose down
 
+bash:           ## Start a bash access to the application container
+	docker-compose exec app bash
```
HeahDude: I'm not fan of this, let's have a `tools`container first with:
```
bash:           ## Start a bash access to the application container
	docker-compose run --rm tools bash
```

tgalopin: That's true. Let's think about this tomorrow.

PR #138, opened 2017-01-26
Handled adherent changing password

HeahDude:
Closes [EM-23](http://jira.sensiolabs.net/browse/EM-23).

### DIFF
--- a/app/Resources/translations/messages.fr.yml
+++ b/app/Resources/translations/messages.fr.yml
@@ -30,6 +30,7 @@ adherent.reset_password.expired_key: Votre demande de réinitialisation de mot d
 
 adherent.update_profile.success: Vos informations ont été mises à jour avec succès.
 adherent.update_interests.success: Vos centres d'intérêts ont bien été mis à jour.
+adherent.update_password.success: Vôtre mot de passe a bien été mis à jour.
 
 #
 # Committee
--- a/app/Resources/views/adherent/change_password.html.twig
+++ b/app/Resources/views/adherent/change_password.html.twig
@@ -4,4 +4,22 @@
 
     <h3>Mot de passe</h3>
 
+    {{ form_start(form) }}
+        {{ form_errors(form) }}
+
+        {{ form_label(form.old_password, 'Mot de passe actuel') }}
+        {{ form_errors(form.old_password) }}
+        {{ form_widget(form.old_password, { attr: { placeholder: 'Entrez votre mot de passe actuel' } }) }}
+
+        {{ form_label(form.password.first, 'Nouveau mot de passe') }}
+        {{ form_errors(form.password.first) }}
+        {{ form_widget(form.password.first, { attr: { placeholder: 'Entrez votre nouveau mot de passe' } }) }}
+
+        {{ form_label(form.password.second, 'Nouveau mot de passe une deuxième fois') }}
+        {{ form_widget(form.password.second, { attr: { placeholder: 'Ré-entrez votre nouveau mot de passe' } }) }}
+
+        {{ form_widget(form.submit, { label: 'Enregistrer ce nouveau mot de passe' }) }}
+
+    {{ form_end(form) }}
+
 {% endblock %}
--- a/src/AppBundle/Entity/Adherent.php
+++ b/src/AppBundle/Entity/Adherent.php
@@ -221,6 +221,11 @@ class Adherent implements UserInterface, GeocodableInterface
         return $this->activatedAt;
     }
 
+    public function changePassword(string $newPassword)
+    {
+        $this->password = $newPassword;
+    }
+
     /**
      * Activates the Adherent account with the provided activation token.
      *
--- a/src/AppBundle/Form/AdherentChangePasswordType.php
+++ b/src/AppBundle/Form/AdherentChangePasswordType.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace AppBundle\Form;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\PasswordType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Security\Core\Validator\Constraints\UserPassword;
+
+class AdherentChangePasswordType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder->add('old_password', PasswordType::class, [
+            'mapped' => false,
+            'constraints' => new UserPassword(['message' => 'adherent.wrong_password']),
+        ]);
+    }
+
+    public function getParent()
+    {
+        return AdherentResetPasswordType::class;
+    }
+}
--- a/src/AppBundle/Membership/AdherentChangePasswordHandler.php
+++ b/src/AppBundle/Membership/AdherentChangePasswordHandler.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace AppBundle\Membership;
+
+use AppBundle\Entity\Adherent;
+use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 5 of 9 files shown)


app/Resources/views/adherent/change_password.html.twig
```diff
@@ -4,4 +4,22 @@
...
+
+        {{ form_label(form.old_password, 'Mot de passe actuel') }}
+        {{ form_errors(form.old_password) }}
+        {{ form_widget(form.old_password, { attr: { placeholder: 'Entrez votre mot de passe actuel' } }) }}
+
+        {{ form_label(form.password.first, 'Nouveau mot de passe') }}
+        {{ form_errors(form.password.first) }}
+        {{ form_widget(form.password.first, { attr: { placeholder: 'Entrez votre nouveau mot de passe' } }) }}
+
+        {{ form_label(form.password.second, 'Nouveau mot de passe une deuxième fois') }}
```
romain-pierre: Je vois plus souvent "Confirmation", voir même aucun label pour ce champs.

src/AppBundle/Membership/AdherentChangePasswordHandler.php
```diff
@@ -0,0 +1,29 @@
...
+use AppBundle\Entity\Adherent;
+use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Component\Security\Core\Encoder\EncoderFactoryInterface;
+
+class AdherentChangePasswordHandler
+{
+    private $manager;
+    private $encoderFactory;
+
+    public function __construct(ObjectManager $manager, EncoderFactoryInterface $encoderFactory)
```
romain-pierre: Why using EncoderFactory instead of PasswordEncoder?

tests/AppBundle/Controller/AdherentControllerTest.php
```diff
@@ -207,6 +207,50 @@ public function testEditAdherentInterests()
...
+    {
+        $this->authenticateAsAdherent($this->client, 'carl999@example.fr', 'secret!12345');
+
+        $crawler = $this->client->request(Request::METHOD_GET, '/espace-adherent/mon-profil/changer-mot-de-passe');
+
+        $this->assertCount(1, $crawler->filter('input[name="adherent_change_password[old_password]"]'));
+        $this->assertCount(1, $crawler->filter('input[name="adherent_change_password[password][first]"]'));
+        $this->assertCount(1, $crawler->filter('input[name="adherent_change_password[password][second]"]'));
+
+        // Submit the profile form with invalid data
```
romain-pierre: Please can you add for this step (invalid data) an assert to be sure that the password have not been updated? With `$this->authenticateAsAdherent')` or directly from the database?

app/Resources/views/adherent/change_password.html.twig
HeahDude: I agree, but cf user stories

src/AppBundle/Membership/AdherentChangePasswordHandler.php
HeahDude: Because the factory already holds every instances, no need to create an extra service as done in other places.

src/AppBundle/Entity/Adherent.php
```diff
@@ -221,6 +221,11 @@ public function getActivatedAt()
         return $this->activatedAt;
     }
 
+    public function changePassword($newPassword)
```
tgalopin: String typehint?

tests/AppBundle/Controller/AdherentControllerTest.php
HeahDude: How could it be possible? It's the same client redirected that we're handling all along the test here.

src/AppBundle/Entity/Adherent.php
HeahDude: Good catch, thx.

HeahDude: Fixed.

tests/AppBundle/Controller/AdherentControllerTest.php
romain-pierre: The client uses a cookie to keep its connection state. So, even if your password is updated (in the database), you will be able to perform some new requests. Ok, if the password is updated, you will be unable to make the password change in the next step...

What do you think about splitting your test in parts:
- Submit empty form
- Submit valid new passwords but wrong old
- Submit valid old password but not same new ones
- Submit valid form

So, just add two steps in your case. This is a security feature, this is why I like to have wide tests.

src/AppBundle/Membership/AdherentChangePasswordHandler.php
romain-pierre: Ok, its up to you. Personally I prefer to use the `PasswordEncoder` (available at `security.password_encoder`) because it's just do what you do (so more convenient) :)

HeahDude: The service you mention just calls the factory, so no need for this level of indirection IMHO (see https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/Resources/config/security.xml#L53).

tests/AppBundle/Controller/AdherentControllerTest.php
HeahDude: I don't understand the need, here the test have a response with errors (due to empty string), how could the password change then (error means no persistence possible). Testing otherwise would mean we would test symfony flow, weird.

app/Resources/translations/messages.fr.yml
```diff
@@ -30,6 +30,7 @@ adherent.reset_password.expired_key: Votre demande de réinitialisation de mot d
 
 adherent.update_profile.success: Vos informations ont été mises à jour avec succès.
 adherent.update_interests.success: Vos centres d'intérêts ont bien été mis à jour.
+adherent.update_password.success: Vos centres d'intérêts ont bien été mis à jour.
```
hhamon: Error message does not message the key name ^^

src/AppBundle/Form/AdherentChangePasswordType.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Security\Core\Validator\Constraints\UserPassword;
+
+class AdherentChangePasswordType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder->add('old_password', PasswordType::class, [
+            'mapped' => false,
+            'constraints' => new UserPassword(['message' => 'adherent.wrong_password']),
```
hhamon: Missing `NotBlank` constraint as well?

src/AppBundle/Membership/AdherentChangePasswordHandler.php
hhamon: Well I agree with Romain as we already have the user so we can use the `security.password_encoder` service and inject the `UserPasswordEncoderInterface` service. Probably easier to mock in unit tests.

app/Resources/translations/messages.fr.yml
HeahDude: Good catch, thanks!

src/AppBundle/Form/AdherentChangePasswordType.php
HeahDude: Not needed with this one.

src/AppBundle/Membership/AdherentChangePasswordHandler.php
HeahDude: Fixed.

app/Resources/translations/messages.fr.yml
HeahDude: Fixed.

PR #142, opened 2017-01-27
[Makefile] added new db-fixtures command.

hhamon:


Makefile
```diff
@@ -32,6 +32,9 @@ db-create:      ## Create the database and load the fixtures in it
 	$(CONSOLE) doctrine:schema:create
 	$(CONSOLE) doctrine:fixtures:load -n
 
+db-fixtures:    ## Reloads the data fixtures for the dev environment
+	$(CONSOLE) doctrine:fixtures:load
```
tgalopin: `-n` ?

hhamon: true that! fixed! thanks.

Makefile
```diff
@@ -32,6 +32,9 @@ db-create:      ## Create the database and load the fixtures in it
 	$(CONSOLE) doctrine:schema:create
 	$(CONSOLE) doctrine:fixtures:load -n
 
+db-fixtures:    ## Reloads the data fixtures for the dev environment
```
HeahDude: Please add it to the `install` task then.

HeahDude: Ah ok you didn't remove it from db-create, so you should :)

PR #144, opened 2017-01-27
Fixes missing adherent address geocoding when user registers.

hhamon:


HeahDude:
Thanks for fixing this bug @hhamon.

PR #167, opened 2017-01-30
Added controller and template skeletons for the add event page.

hhamon:
Closes [EM-42](http://jira.sensiolabs.net/browse/EM-42).

### DIFF
--- a/app/Resources/views/committee/add_event.html.twig
+++ b/app/Resources/views/committee/add_event.html.twig
@@ -1,0 +1,17 @@
+{% extends 'base.html.twig' %}
+
+{% block title 'Nouvel événement' %}
+
+{% block banner '' %}
+
+{% block sidebar %}
+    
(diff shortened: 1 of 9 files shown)


app/Resources/views/committee/add_event.html.twig
```diff
@@ -0,0 +1,17 @@
+{% extends 'base.html.twig' %}
+
+{% block title 'Nouvel événement' %}
+
+{% block banner '' %}
+
+{% block sidebar %}
```
HeahDude: What is this empty block?

HeahDude:
👍

app/Resources/views/committee/add_event.html.twig
hhamon: For the "hosts" block on the right. It will probably be dropped.

PR #176, opened 2017-01-30
Improve front structure

tgalopin:


.gitignore
```diff
@@ -13,3 +14,4 @@
 /phpunit.xml
 /supervisord.log
 /supervisord.pid
+/yarn-error.log
```
HeahDude: Can you please add this one in the clean task of the makefile?

src/AppBundle/Donation/DonationFactory.php
```diff
@@ -12,6 +12,7 @@ class DonationFactory
     public function createDonationFromRequest(Request $request): Donation
     {
         $donation = new Donation();
+        $donation->setAmount(50);
```
HeahDude: You could set this one as argument of the constructor with a default value.

PR #185, opened 2017-01-31
Added a CommitteeFeedMessage entity

HeahDude:
Closes [EM-46](http://jira.sensiolabs.net/browse/EM-46).

PR #184, opened 2017-01-31
Handled committee page blocks per authorization

HeahDude:
Closes [EM-32](http://jira.sensiolabs.net/browse/EM-32).

PR #174, opened 2017-01-30
Handled committee page

HeahDude:
Closes [EM-40](http://jira.sensiolabs.net/browse/EM-40).

PR #189, opened 2017-01-31
Handled host creating committee message

HeahDude:
Closes [EM-47](http://jira.sensiolabs.net/browse/EM-47).

### DIFF
--- a/src/AppBundle/Controller/CommitteeController.php
+++ b/src/AppBundle/Controller/CommitteeController.php
@@ -2,13 +2,16 @@
 
 namespace AppBundle\Controller;
 
+use AppBundle\Committee\CommitteePermissions;
 use AppBundle\Committee\Event\CommitteeEventCommand;
 use AppBundle\Entity\Committee;
 use AppBundle\Form\CommitteeEventCommandType;
+use AppBundle\Form\CommitteeFeedMessageType;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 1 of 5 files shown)


HeahDude:
Until #174, #184, and #185, please only review bbdd77470fca5a7e7104c89743cd3e00e2881d71.

src/AppBundle/Committee/CommitteeFeedHandler.php
```diff
@@ -0,0 +1,28 @@
...
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\Committee;
+use AppBundle\Entity\CommitteeFeedMessage;
+use Doctrine\Common\Persistence\ManagerRegistry;
+
+class CommitteeFeedHandler
+{
+    private $manager;
+
+    public function __construct(ManagerRegistry $doctrine)
```
hhamon: Should be `ObjectManager` instead of `ManagerRegistry`.

src/AppBundle/Twig/CommitteeExtension.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return $this->authorizationChecker->isGranted(CommitteePermissions::CREATE, $committee);
+    }
+
+    public function canSee(Committee $committee): bool
+    {
+        return $this->authorizationChecker->isGranted(CommitteePermissions::SHOW, $committee);
+    }
+
+    public function getPath(string $routeName, Committee $committee): bool
```
hhamon: Return type must be `string`

src/AppBundle/Twig/CommitteeExtension.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return $this->urlGenerator->generate($routeName, [
+            'uuid' => $committee->getUuid()->toString(),
+            'slug' => $committee->getSlug(),
+        ]);
+    }
+
+    public function getUrl(string $routeName, Committee $committee): bool
```
hhamon: Return type must be `string`

src/AppBundle/Controller/CommitteeController.php
```diff
@@ -2,36 +2,67 @@
...
+     * @Method("GET|POST")
      * @Security("is_granted('SHOW_COMMITTEE', committee)")
      */
-    public function showAction(Committee $committee): Response
+    public function showAction(Committee $committee, Request $request): Response
     {
+        $form = null;
+        if ($this->isGranted(CommitteePermissions::HOST, $committee)) {
+            $form = $this->createForm(CommitteeFeedMessageType::class)
+                ->add('publish', SubmitType::class, ['label' => 'Publier']);
```
hhamon: Can't you have this submit button in the template instead?

HeahDude: Why? This prevents checking if it's been clicked.

hhamon: Because a submit button is not a controller or model thing. It's a view related thing. If you don't need to check for the button to be clicked, there are absolutely no reasons to add it to the form object.

HeahDude: I do not agree, the button is a form thing that can have form themes as well it should either be optional in a type or in a controller. This is done every where else, why should this be a blocker here. I would accept a comment, not a request for change.

hhamon: You can leave it if you want but it makes the controller bigger for nothing. Anyway see my previous comments above that must be fixed.

src/AppBundle/Committee/CommitteeFeedHandler.php
HeahDude: Fixed in the previous PR.

src/AppBundle/Twig/CommitteeExtension.php
HeahDude: Fixed in the previous PR.

src/AppBundle/Twig/CommitteeExtension.php
HeahDude: Fixed in the previous PR.

src/AppBundle/Controller/CommitteeController.php
HeahDude: Fixed in the previous PR.

PR #192, opened 2017-02-01
[RFR] Various fixes

marineb:


tgalopin:
Isn't is an issue to have fixed images for CTA?

PR #201, opened 2017-02-02
Wire Redis for cache and sessions

tgalopin:


app/config/config.yml
```diff
@@ -160,3 +162,12 @@ bazinga_geocoder:
             locale:  '%locale%'
             use_ssl: true
             api_key: '%env(GMAPS_API_KEY)%'
+
+snc_redis:
+    clients:
+        default:
+            type: predis
+            alias: default
+            dsn: "redis://%env(REDIS_HOST)%"
```
salahm: Y U NO use default value ? (or update docker to handle this env var)

tgalopin: What do you mean? I need this env var for staging (in Kubernetes, Redis is not local).

romain-pierre: Good idea to set the host in a parameter! This also allow an easy switch between docker and local Redis (in dev).

romain-pierre:
If you wanna use Redis for sessions management in all environments, you could also remove `var/sessions/.gitkeep`

tgalopin:
I removed var/sessions and the line for it in docker-compose.yml.

PR #209, opened 2017-02-02
Refactored social networks links.

hhamon:


tgalopin:
Couldn't we create a Domain Object for this, in order to use in Events too?

hhamon:
It may make sense yes.

PR #229, opened 2017-02-04
Map page with committees list

atamano:


app/Resources/views/page/la-carte.html.twig
```diff
@@ -0,0 +1,23 @@
...
+            <header class="space--60-0 l__wrapper--slim">
+                <h1 class="text--large b__nudge--bottom-small">
+                    Comités à proximité
+                </h1>
+            </header>
+
+            <article class="list__links list__links--col list__links--no-decor">
+              <ul>
+                {% for committee in committees %}
+                    <li class="text--on-white">{{ committee.name }} <br/> {{ committee.getCityName }}</li>
```
hhamon: `committee.cityName` is enough

src/AppBundle/Controller/PageController.php
```diff
@@ -93,6 +94,19 @@ public function mouvementOrganisationAction()
     }
 
     /**
+     * @Route("/le-mouvement/la-carte", name="page_le_mouvement_la_carte")
+     * @Method("GET")
+     */
+    public function mouvementCarteAction()
+    {
+        return $this->render('page/la-carte.html.twig', [
+            'committees' => $this->getDoctrine()->getRepository(Committee::class)->findBy([
```
hhamon: it's better to have an explicit method in the repository class to fetch approved committees rather than using Doctrine built-in finders.

tests/AppBundle/Controller/PageControllerTest.php
```diff
@@ -21,6 +21,14 @@ public function testPages($path)
         $this->assertResponseStatusCode(Response::HTTP_OK, $response = $this->client->getResponse());
     }
 
+    public function testMap()
+    {
+        $crawler = $this->client->request(Request::METHOD_GET, '/le-mouvement/la-carte');
+
+        $this->assertResponseStatusCode(Response::HTTP_OK, $response = $this->client->getResponse());
```
hhamon: No need to create the `$response` variable here as it's not used later.

src/AppBundle/Controller/PageController.php
```diff
@@ -93,6 +94,17 @@ public function mouvementOrganisationAction()
     }
 
     /**
+     * @Route("/le-mouvement/la-carte", name="page_le_mouvement_la_carte")
+     * @Method("GET")
+     */
+    public function mouvementCarteAction()
+    {
+        return $this->render('page/la-carte.html.twig', [
+            'committees' => $this->getDoctrine()->getRepository(Committee::class)->findByStatus(Committee::APPROVED),
```
hhamon: Better but it's even better with a method like `->findApprovedCommittees()` or `->findForInteractiveMap()`.

tgalopin: +1 for findApprovedCommittees

hhamon:
Can you squash your commits to make only one please?

atamano:
sure !

PR #242, opened 2017-02-05
Improve ui of the flash messages

marineb:


front/style/_register.scss
```diff
@@ -1,98 +1,175 @@
...
+            }
+
+            .form {
+                font-family: $font-roboto;
+            }
+        }
+    }
+}
+
+.flash {
```
tgalopin: This should probaby be in `_forms` as it's not limited to registration.

marineb: Yes, good point. Updated!

PR #253, opened 2017-02-06
Display the members list for a committee

romain-pierre:
Closes [EM-60](http://jira.sensiolabs.net/browse/EM-60).

### DIFF
--- a/app/Resources/views/committee/members.html.twig
+++ b/app/Resources/views/committee/members.html.twig
@@ -6,6 +6,64 @@
 
 {% block banner '' %}
 
+{% block javascripts %}
+    <script type="application/javascript">
+        var allCheckbox = dom('#members-check-all');
+        var selectionCheckboxes = document.getElementsByName('members[]');
+        var contactButton = dom('#members-contact-button');
+        var exportButton = dom('#members-export-button');
+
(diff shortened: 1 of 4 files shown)


app/Resources/views/committee/members.html.twig
```diff
@@ -6,6 +6,44 @@
 
 {% block banner '' %}
 
+{% block javascripts %}
+    <script type="application/javascript">
+        var allCheckbox = dom('#members-check-all');
```
hhamon: `allCheckboxesContainer` ?

app/Resources/views/committee/members.html.twig
```diff
@@ -6,6 +6,44 @@
...
+        var membersCheckbox = dom("[name='members[]']");
```
hhamon: `membersCheckboxes` ?

app/Resources/views/committee/members.html.twig
romain-pierre: Why container? This is an unique field.

hhamon: Nevermind... It targets the checkbox in the header section. I thought it meant "all checkboxes" elements.

PR #261, opened 2017-02-07
Hide committee timeline message body for non authenticated users.

hhamon:
Close EM-38.

### DIFF
--- a/tests/AppBundle/Controller/CommitteeControllerTest.php
+++ b/tests/AppBundle/Controller/CommitteeControllerTest.php
@@ -3,11 +3,15 @@
 namespace Tests\AppBundle\Controller;
 
 use AppBundle\DataFixtures\ORM\LoadAdherentData;
+use AppBundle\DataFixtures\ORM\LoadCommitteeEventData;
 use AppBundle\Entity\CommitteeEvent;
 use AppBundle\Entity\CommitteeFeedItem;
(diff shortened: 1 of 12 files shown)


tests/AppBundle/Controller/CommitteeControllerTest.php
```diff
@@ -383,7 +393,7 @@ public function testAnonymousGuestCanShowCommitteePage()
         $this->assertTrue($this->seeHosts($crawler, 2), 'The guest should see the hosts');
         $this->assertTrue($this->seeHostsContactLink($crawler, 2), 'The guest should see the hosts contact link');
         $this->assertFalse($this->seeHostNav($crawler), 'The guest should not see the host navigation');
-        $this->assertSeeSocialLinks($crawler, $this->getCommitteeRepository()->findOneByUuid(LoadAdherentData::COMMITTEE_1_UUID));
+        $this->seeSocialLinks($crawler, $this->getCommitteeRepository()->findOneByUuid(LoadAdherentData::COMMITTEE_1_UUID));
```
tgalopin: The assert prefix seems useful to understand it's an assertion, why remove it?

hhamon: Because the other new "assert" custom methods don't have this prefix. Maybe it's better to have `assert` prefix everywhere then.

tgalopin: As you prefer.

PR #268, opened 2017-02-08
Implemented committee event registration.

hhamon:
Closes EM-64 and EM-65.

### DIFF
--- a/src/AppBundle/Controller/CommitteeEventController.php
+++ b/src/AppBundle/Controller/CommitteeEventController.php
@@ -2,11 +2,14 @@
 
 namespace AppBundle\Controller;
 
+use AppBundle\Committee\Event\CommitteeEventRegistrationCommand;
 use AppBundle\Entity\CommitteeEvent;
+use AppBundle\Form\CommitteeEventRegistrationType;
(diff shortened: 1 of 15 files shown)


src/AppBundle/Controller/CommitteeEventController.php
```diff
@@ -34,21 +37,52 @@ public function showAction(CommitteeEvent $event): Response
...
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $this->get('app.committee.committee_event_registration_handler')->handle($command);
+            $this->addFlash('notice', $this->get('translator')->trans('committee.event.registration.success'));
+
+            return $this->redirectToRoute('app_committee_attend_event_confirmation', [
+                'committee_uuid' => (string) $committee->getUuid(),
+                'slug' => $event->getSlug(),
+                'uuid' => (string) $command->getRegistrationUuid(),
+            ]);
```
romain-pierre: Why not using something like `$this->redirect($this->get('app.twig.committee_extension')->getUrl('app_committee_attend_event_confirmation', $committee))`?

hhamon: because it's weird to call a twig extension from a controller...

PR #304, opened 2017-02-11
add ui to event detail page

marineb:


hhamon:
Can you also please display the event participants count somewhere on this page. We forgot to add it.

tgalopin:
Indeed, that's a very good point.

marineb:
good catch! will add that in!

PR #389, opened 2017-02-24
Forbid creation of events and messages in an inactive committee

tgalopin:


src/AppBundle/Controller/CommitteeController.php
```diff
@@ -218,7 +218,7 @@ public function contactMembersAction(Request $request, Committee $committee): Re
     /**
      * @Route("/rejoindre", name="app_committee_follow", condition="request.request.has('token')")
      * @Method("POST")
-     * @Security("is_granted('FOLLOW_COMMITTEE', committee)")
+     * @Security("is_granted('FOLLOW_COMMITTEE', committee) and committee.isApproved")
```
hhamon: Checking that the committee is approved should be done inside the voter IMO.

src/AppBundle/Controller/CommitteeController.php
```diff
@@ -240,7 +240,7 @@ public function followAction(Request $request, Committee $committee): Response
     /**
      * @Route("/quitter", name="app_committee_unfollow", condition="request.request.has('token')")
      * @Method("POST")
-     * @Security("is_granted('UNFOLLOW_COMMITTEE', committee)")
+     * @Security("is_granted('UNFOLLOW_COMMITTEE', committee) and committee.isApproved")
```
hhamon: Checking that the committee is approved should be done inside the voter IMO.

src/AppBundle/Controller/CommitteeController.php
```diff
@@ -36,7 +36,7 @@ class CommitteeController extends Controller
     public function showAction(Request $request, Committee $committee): Response
     {
         $form = null;
-        if ($this->isGranted(CommitteePermissions::HOST, $committee)) {
+        if ($this->isGranted(CommitteePermissions::HOST, $committee) && $committee->isApproved()) {
```
hhamon: This security policy rule deserves a dedicated voter maybe:

```php
if ($this->isGranted(CommitteePermissions::POST_MESSAGE, $committee)) {
```

PR #470, opened 2017-02-28
Block donation banner on Je Marche

marineb:


tgalopin:
Can you use `{% block banner '' %}` instead?

```

marineb:
Oh cool. I didn't know this functionality. Took the opportunity to adjust other pages as well!

PR #471, opened 2017-02-28
Add new program page

marineb:


tgalopin:
It's good for me but it would be better if you reenabled the proposals tests in tests/AppBundle/Controller/PageControllerTest.php (line 42 and 43).

PR #581, opened 2017-03-15
Add CSS to render UL/OL in committee messages events (fix #521)

### LINKED ISSUE #521: Corriger le CSS des listes à puces (<ul> et <li>)
Le CSS des listes à puces (éléments HTML UL et LI) est remis à zéro.

Par conséquent, les listes à puces sont "invisibles".

Exemple sur cet événement : https://en-marche.fr/evenements/d1c94ea8-dbed-4f24-b9c9-b71c9832da88/2017-03-09-environnement-et-democratie-le-programme-d-emmanuel-macron

La description "A propos" contient une liste à puces qui apparaît comme de simples retours à la ligne.

ngrilly:


tgalopin:
@marineb what do you think? Shouldn't we use specific lists icons?

marineb:
@tgalopin In this context, I think it's better to use standard bullets.

front/style/_committee.scss
```diff
@@ -330,6 +330,17 @@
                 }
             }
         }
+
+        &-description {
```
marineb: I think the lists look a bit tight within the event description paragraph. Do you mind adding breathing room above and below the UL/OLs?

ngrilly: I've answered on the Slack channel, with a screenshot and some questions :-)

tgalopin: @ngrilly did you really? in which channel?

ngrilly: @tgalopin You can't see because we used Slack direct messages.

ngrilly: @marineb In our private discussion on Slack, you told me you'd like to rename one the CSS classes. Can you you give some details?

marineb: Yes, the comment is at: https://github.com/EnMarche/en-marche.fr/pull/581#discussion_r106326343.

app/Resources/views/committee/timeline/_message.html.twig
```diff
@@ -1,5 +1,5 @@
 {% if is_granted('ROLE_ADHERENT') %}
-    <div class="text--body">{{ item.content|raw }}</div>
+    <div class="committee-event-description text--body">{{ item.content|raw }}</div>
```
ngrilly: I agree that the current naming is confusing.

The styling is used in three different places:

- In the timeline for messages (app/Resources/views/committee/timeline/_message.html.twig)
- In the timeline for event descriptions (app/Resources/views/committee/add_event.html.twig)
- On the page of each event for the event description (app/Resources/views/events/show.html.twig)

The simple solution is to rename `committee-event-description` to something more general like `committee-feed-item-description` or 'feed-item-description`.

Another solution, which you suggested, would be to avoid using a class in the CSS selector, but then I'd need to move the styling in a mixin because it would be used in 2 or 3 different places.

What do you want me to do? ;-)

tgalopin: I don't know about the words themselves (I'll let @marineb decide on that) but you should be careful to use BEM (even if we didn't everywhere, that's possible): http://getbem.com/introduction/

marineb: @ngrilly I think neither of those would work for the event page. Currently on master, there's only one `committee-event-description`, and it's on the event page. Which makes sense. But having `feed` on that page wouldn't work. 

What I would suggest is adding a new utility class. Something like `text--has-list` or `text--with-list`, or something like that. The same way we have `text--breakword` or `icon--with-text`, in the `text.scss` file. You could also make it a mixin if and import it where you need it, and make a class of it where you don't have a class you can put it under. 

Yes, and @tgalopin is right about BEM. I think somehow the timeline and a few other admin pages don't follow this convention but where we're using BEM, we should use BEM. 

Let me know if that makes sense! And sorry, I hadn't seen those responses until now. Feel free to ping me on Slack in the future if I'm slow!

tgalopin:
@ngrilly is there something to do here still? Can we finalize this :) ?

marineb:
@ngrilly Totally get it if you don't have the time to wrap it up. I'm happy to finalize it if you can't. Let me know!

ngrilly:
Sorry for not having followed-up, and thanks for finalizing this! I regularly miss GitHub notifications on issues/PR I'm involved in, because they are messed with other notifications...

marineb:
@ngrilly no worries. 😄 Yes, Github need to be better at notification management. I also miss mines sometimes.

ngrilly:
@marineb FYI, I found this info about GitHub notifications:

https://help.github.com/articles/about-notification-emails/#the-cc-address

And created this filter in Gmail:

> Matches: (cc:(mention@noreply.github.com OR author@noreply.github.com OR assign@noreply.github.com))
> Do this: Categorize as Personal	

😄

HeahDude:
@ngrilly I've done that for a while ;), don't forget the new review_requested@noreply.github.com and you may use comment@noreply.github.com to know when someone reply to a comment you made.

PR #601, opened 2017-03-19
Minor perf improvement

xavismeh:
When the number of articles is a multiple of the pagination size, we don't have to perform a query on article to throw a 404


tgalopin:
I don't understand how this would work.

Imagine I have 50 articles and 5 articles per page: `$articlesCount = 50` and `self::PER_PAGE = 5`. Here, `$articlesCount % self::PER_PAGE` would be `0`, therefore throwing a 404 on each page. I don't think that's what you expected :) . Am I wrong?

tgalopin:
Thanks @xavismeh

PR #616, opened 2017-03-20
Fix attend event

mykiwi:


app/Resources/views/committee/timeline/_event.html.twig
```diff
@@ -27,10 +27,14 @@
             <li>
                 {% if item.event.finishAt < date() %}
                     Cet événement est terminé
+                {% elseif event.participantsCount < event.capacity %}
```
hhamon: Better to have a dedicated method like `$event.->isFull()` so that the Twig template can be simplified to `{% elseif not event.full %}`.

app/Resources/views/committee/timeline/_event.html.twig
```diff
@@ -27,10 +27,14 @@
             <li>
                 {% if item.event.finishAt < date() %}
```
hhamon: Better to have something like `event.finished`

app/Resources/views/events/show.html.twig
```diff
@@ -102,8 +102,15 @@
                         <div class="text--body text--white">
                             Cet événement est terminé
                         </div>
-                    {% else %}
+                    {% elseif event.participantsCount < event.capacity %}
```
hhamon: Same as above.

app/Resources/views/search/_events.html.twig
```diff
@@ -29,13 +29,17 @@
 
                 <div class="search__results__cta">
                     {% if event.finishAt < date() %}
-                        <div class="text--body text--white">
+                        <div class="text--body">
                             Cet événement est terminé
                         </div>
-                    {% else %}
+                    {% elseif event.participantsCount < event.capacity %}
```
hhamon: Same as above

tests/AppBundle/Entity/EventTest.php
```diff
@@ -0,0 +1,85 @@
...
+use AppBundle\Entity\Event;
+use AppBundle\Entity\PostAddress;
+use Ramsey\Uuid\UuidInterface;
+
+class EventTest extends \PHPUnit_Framework_TestCase
+{
+    public function testIsFinished()
+    {
+        $event = new Event(
+            $this->prophesize(UuidInterface::class)->reveal(),
```
hhamon: We don't use prophecy for mocks.

hhamon:
👍

PR #620, opened 2017-03-21
enable account by default

mykiwi:


hhamon:
Don't forget to squash and rebase.

tgalopin:
If I understand correctly, you update `activedAt` only on mail confirmation but put the enabled status by default? This seems like a good idea to avoid unecessary migration :) .

mykiwi:
Yes

PR #621, opened 2017-03-21
Remove limit of 5 displayed committees in the adherent navigation bar.

hhamon:


src/AppBundle/Collection/CommitteeMembershipCollection.php
```diff
@@ -40,4 +48,13 @@ public function getCommitteeFollowerMemberships(): self
             return $membership->isFollower();
         });
     }
+
+    private function merge(self $other)
```
mykiwi: `: void`

hhamon: thanks! fixed.

PR #627, opened 2017-03-22
Add copy updates for the procuration page

marineb:


src/AppBundle/Form/ProcurationElectionsType.php
```diff
@@ -19,15 +19,19 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder
             ->add('electionPresidentialFirstRound', CheckboxType::class, [
                 'required' => false,
+                'data' => false,
```
tgalopin: This shouln't be necessary. Did you have an issue?

marineb: @tgalopin there was a change request to set the checkboxes to `unchecked` by default. per the documentation, it seemed to be how to handle it. let me know your thoughts!

tgalopin: It's not here it should be modified then, but in the entity itself: https://github.com/EnMarche/en-marche.fr/blob/master/src/AppBundle/Entity/ProcurationRequest.php#L219

marineb: Oh I see. Thank you!

PR #628, opened 2017-03-22
Update CTAs on events search page

marineb:


tgalopin:
What is the reason behind this change?

marineb:
Usability I imagine, clarifying the call to actions. The change came from TR.

PR #641, opened 2017-03-22
add organizer name, email and phone on admin event list

mykiwi:


app/Resources/views/admin/event_organizer.html.twig
```diff
@@ -0,0 +1,14 @@
+{% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
+
+{% block field%}
```
tgalopin: Space missing

PR #642, opened 2017-03-22
Include supervisors as hosts in adhernt entity

tgalopin:


src/AppBundle/Entity/Adherent.php
```diff
@@ -583,7 +583,7 @@ public function getMemberships()
     public function isHost(): bool
     {
         foreach ($this->memberships as $membership) {
-            if ($membership->isHostMember()) {
+            if ($membership->isSupervisor() || $membership->isHostMember()) {
```
hhamon: Better to use the `canHostCommittee()` method that does already this logic.

src/AppBundle/Entity/Adherent.php
```diff
@@ -597,7 +597,8 @@ public function isHost(): bool
     public function isHostOf(Committee $committee): bool
     {
         foreach ($this->memberships as $membership) {
-            if ($membership->isHostMember() && $membership->getCommitteeUuid() === $committee->getUuid()->toString()) {
+            if (($membership->isSupervisor() || $membership->isHostMember())
```
hhamon: Same as above.

src/AppBundle/Entity/Adherent.php
tgalopin: This method is used in some places for this behavior, perhaps can we remove it but in the meantime we should support it here too.

PR #643, opened 2017-03-22
Switch to the md5 encoder in test env to speed up tests

rpg600:
The test duration is divided by 3 now.


hhamon:
Can you also please override the number of hash iterations to 1.

tgalopin:
@hhamon are you ok with this? can we merge?

PR #653, opened 2017-03-24
add admin event list filters

mykiwi:


src/AppBundle/Admin/EventAdmin.php
```diff
@@ -125,6 +125,75 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
             ->add('name', null, [
                 'label' => 'Nom',
             ])
+            ->add('createdAt', 'doctrine_orm_date_range', [
```
hhamon: Can't you use a FCQN instead here?

src/AppBundle/Admin/EventAdmin.php
```diff
@@ -125,6 +125,75 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                'label' => 'Date de création',
+                'field_type' => 'sonata_type_date_range_picker',
+            ])
+            ->add('hostFirstName', 'doctrine_orm_callback', [
```
hhamon: Same here

src/AppBundle/Admin/EventAdmin.php
tgalopin: Not a problem (I don't think it's supported by Sonata yet).

PR #667, opened 2017-03-27
fix event filter

mykiwi:


src/AppBundle/Admin/EventAdmin.php
```diff
@@ -128,8 +128,8 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
             ->add('name', null, [
                 'label' => 'Nom',
             ])
-            ->add('createdAt', DateRangeFilter::class, [
```
tgalopin: Can you keep both and put `beginAt` as `'show_filter' => true` ?

PR #671, opened 2017-03-27
Send mail on match and create summary page for members

tgalopin:


app/Resources/views/procuration/my_request.html.twig
```diff
@@ -0,0 +1,212 @@
...
+            <div class="procuration-manager__associate__col">
+                <h4 class="b__nudge--bottom">
+                    Mandant (a besoin d'une procuration)
+                </h4>
+
+                <div class="profile-label">
+                    Auteur
+                </div>
+                <div class="profile-value">
+                    {{ request.gender == 'male' ? 'Monsieur' : 'Madame' }}
```
hhamon: I think we need some sort of filter for handling gender related words:

```twig
{{ 'mister_or_madame'|gender_choice(request.gender) }}
```

tgalopin: This is quite urgent, I need to push it in prod tonight :)

PR #672, opened 2017-03-28
Add homepage action cards

marineb:


app/Resources/views/home/index.html.twig
```diff
@@ -42,6 +42,19 @@
             </a>
         </div>
 
+        <div class="content">
+            <article class="l__row l__row--v-stretch l__tablet--col">
+                <div class="action action__home yellow">
+                    <p class="pink">J’agis pour  la campagne !</p>
```
tgalopin: Strange character

app/Resources/views/home/index.html.twig
```diff
@@ -42,6 +42,19 @@
             </a>
         </div>
 
+        <div class="content">
```
tgalopin: Can you put this block around `{% if enable_canary %}...{% endif %}` so we can disable it in prod while it's not ready but still merge this PR?

app/Resources/views/home/index.html.twig
```diff
@@ -8,7 +8,7 @@
 {% block content %}
 <main>
     <section class="l__wrapper">
-        <div class="l__row l__tablet--col l__row--v-stretch b__nudge--bottom">
+        <div class="l__row l__tablet--col l__row--v-stretch">
```
tgalopin: Same here, can you put `{% if not enable_canary %} b__nudge--bottom{% endif %}` so we can disable it in prod while it's not ready but still merge this PR?

app/Resources/views/home/index.html.twig
```diff
@@ -42,6 +43,21 @@
             </a>
         </div>
 
+        {% if enable_canary %}
+        <div class="content">
+            <article class="l__row l__row--v-stretch l__tablet--col">
+                <div class="action action__home yellow">
+                    <p class="pink">J'agis pour  la campagne !</p>
```
tgalopin: Still a weird character here :/

PR #676, opened 2017-03-28
Added event's status in the event admin form

rpg600:


hhamon:
It would be better to create an `EventStatusType` class instead.

PR #677, opened 2017-03-28
Added committee subscription date for each member in the committee members list.

hhamon:


src/AppBundle/Entity/Adherent.php
```diff
@@ -446,43 +446,46 @@ public function updateMembership(MembershipRequest $membership, PostAddress $pos
      * Joins a committee as a SUPERVISOR priviledged person.
      *
      * @param Committee $committee
+     * @param string    $subscriptionDate
```
mykiwi: I would remove these annotations (only keep the description) since everything is type hinted.

src/AppBundle/Repository/CommitteeMembershipRepository.php
```diff
@@ -302,20 +302,41 @@ private function findPriviledgedMembers(string $committeeUuid, array $privileges
...
+        $query = $this
+            ->createCommitteeMembershipsQueryBuilder($committeeUuid)
+            ->addSelect('a')
+            ->getQuery()
+        ;
+
+        return new CommitteeMembershipCollection($query->getResult());
+    }
+
+    private function createCommitteeMembershipsQueryBuilder(string $committeeUuid, string $alias = 'cm'): QueryBuilder
```
mykiwi: IMHO phpdoc is useful only when there is no type hinting.

tgalopin: I agree with @mykiwi : using docblock is not useful with typehint, espacially for private methods.

hhamon: So shall we remove docblocks then?

src/AppBundle/Entity/Adherent.php
hhamon: You're right.

src/AppBundle/Repository/CommitteeMembershipRepository.php
tgalopin: In PR only, yes. We don't care about the current state. Let's remove them when possible in changes (only when they have no value ofc).

src/AppBundle/Entity/Adherent.php
```diff
@@ -344,11 +343,6 @@ public function disableCommitteesNotifications()
     /**
      * Activates the Adherent account with the provided activation token.
      *
-     * @param AdherentActivationToken $token
-     * @param string                  $timestamp
-     *
-     * @throws AdherentException
-     * @throws AdherentTokenException
      */
     public function activate(AdherentActivationToken $token, string $timestamp = 'now')
```
mykiwi: `: void`

I would keep `@throws`

hhamon:
Removed obvious phpdoc blocks.
Improved typehinting.

src/AppBundle/Entity/Adherent.php
```diff
@@ -362,7 +351,7 @@ public function activate(AdherentActivationToken $token, string $timestamp = 'no
         $this->activatedAt = new \DateTime($timestamp);
     }
 
-    public function resetPassword(AdherentResetPasswordToken $token)
+    public function resetPassword(AdherentResetPasswordToken $token): void
```
mykiwi: You could add the `@throws`

PR #678, opened 2017-03-28
Fix hosts managed by referent list

tgalopin:


src/AppBundle/Repository/AdherentRepository.php
```diff
@@ -177,7 +177,7 @@ public function findHostsManagedBy(Adherent $referent): array
     {
         return array_filter($this->findAllManagedBy($referent), function (Adherent $adherent) {
             foreach ($adherent->getMemberships() as $membership) {
-                if ($membership->isHostMember()) {
+                if ($membership->isSupervisor() || $membership->isHostMember()) {
```
hhamon: better to use the `$membership->canHostCommittee()` method no?

tgalopin: Indeed, thx!

PR #679, opened 2017-03-28
JeMarche reports exports now embed all email addresses.

hhamon:


src/AppBundle/Admin/JeMarcheReportAdmin.php
```diff
@@ -83,6 +85,21 @@ protected function configureListFields(ListMapper $listMapper)
...
                     'delete' => [],
                 ],
-            ]);
+            ])
+        ;
+    }
+
+    public function getExportFields()
+    {
+        return array(
```
tgalopin: Please use `[...]`

src/AppBundle/Entity/JeMarcheReport.php
```diff
@@ -230,6 +240,11 @@ public function getNotConvinced(): ?int
         return $this->notConvinced;
     }
 
+    public function getNotConvincedList(string $glue = '|'): string
```
tgalopin: notConvinced is a number

PR #681, opened 2017-03-28
Referent user is now included in the managed adherents list.

hhamon:


src/AppBundle/Referent/ReferentDatabaseDumper.php
```diff
@@ -0,0 +1,118 @@
...
+            return $this->factory->createManagedNonFollowersCollectionFor($referent);
+        }
+
+        if ('followers' === $type) {
+            return $this->factory->createManagedFollowersCollectionFor($referent);
+        }
+
+        if ('hosts' === $type) {
+            return $this->factory->createManagedHostsCollectionFor($referent);
+        }
```
tgalopin: `return [];` to avoid errors.

src/AppBundle/Command/ReferentDumpDatabaseCommand.php
```diff
@@ -0,0 +1,57 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $referent = $input->getArgument('referent');
+        $type = $input->getArgument('type');
+
+        if (!$this->io->isQuiet()) {
+            $this->io->writeln(sprintf('Starting database export for referent <comment>%s</comment> (type: <comment>%s</comment>).', $referent, $type));
+        }
+
+        $this->dumper->dump($referent, $type);
```
tgalopin: If the type is not specified, it would be more useful if all the types were dumped (not only the `all` type)

src/AppBundle/Command/ReferentDumpDatabaseCommand.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @var ReferentDatabaseDumper
+     */
+    private $dumper;
+
+    protected function configure()
+    {
+        $this
+            ->setName('app:referent:dump-db')
```
tgalopin: `app:referent:dump`

src/AppBundle/Command/ReferentDumpDatabaseCommand.php
hhamon: Done.

tgalopin:
I know I'll be the devil here, but could you use the occasion to test the ReferentDatabaseDumper class? A few simple unit tests should be enough but would greatly improve the coverage :) .

hhamon:
@tgalopin tests were added

app/config/config_test.yml
```diff
@@ -44,6 +43,7 @@ parameters:
         sante: Santé
         securite: Sécurité
         territoire: Territoire
+    referent_database_root_dir:  '%kernel.cache_dir%/data'
```
tgalopin: It's not only referents data, it's global system data.

tests/AppBundle/Referent/ReferentDatabaseDumperTest.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $path = $this->getDatabaseDumpedFilePath($type);
+
+        $this->assertFileNotExists($path);
+
+        $this->dumper->dump(LoadAdherentData::ADHERENT_8_UUID, $type);
+
+        $this->assertFileExists($path);
+        $this->assertGreaterThan(0, filesize($path));
+    }
```
tgalopin: Could you also test it's valid JSON/serialized data? It could avoid having errors in production.

hhamon:
Requested changes were applied.

src/AppBundle/Referent/ReferentDatabaseDumper.php
```diff
@@ -0,0 +1,117 @@
...
+use AppBundle\Exception\ReferentNotFoundException;
+use AppBundle\Repository\AdherentRepository;
+use League\Flysystem\FilesystemInterface;
+
+class ReferentDatabaseDumper
+{
+    const STORAGE_SPACE = 'dumped_referents_users';
+
+    const EXPORT_TYPES = [
+        'all',
```
mykiwi: what about create a constant for each types?

hhamon: I thought about it but it doesn't really make sense so far because we don't really hardcode the names outside of the class itself. Only in the unit tests of course. This tool is called from a console command where the type argument must be given to the command.

PR #693, opened 2017-03-29
Rename components directory to icons

marineb:
Something I've been meaning to do.


tgalopin:
Not all the files in `components` are icons (see footer--slim for instance). I would like to have an `include` directory for these

marineb:
@tgalopin totally oversaw / forgot about the footer component. Updated. Let me know what you think!

marineb:
👋

PR #699, opened 2017-03-29
Fix new home blocks

tgalopin:


app/Resources/views/home/index.html.twig
```diff
@@ -51,7 +50,7 @@
                     <a href="{{ path('app_je_marche') }}" class="pink-yellow">Voir les actions</a>
                 </div>
                 <div class="action action__home blue">
-                    <p>2000 événements cette semaine !</p>
```
hhamon: I like this dynamic live counter ^^

tgalopin: Such dynamism :D

PR #704, opened 2017-03-29
Optimize number of SQL queries whenever possible.

hhamon:
This PR greatly improves the number of SQL queries performed on the frontend side when an adherent is logged in to his\her account. From 21 SQL queries to only 2!

Before
enmarche-dev-after

After
enmarche-dev-before

The duplicated SQL queries came from two different locations:

1. the adherent's navigation bar in the top right end corner,
2. security voters.

All voters now receive the `CommitteeManager` class that performs SQL queries with Doctrine repositories. This means optimizing can now be done inside the `CommitteeManager` class directly as everything is centralized. The `CommitteeManager` object is now able to check an `Adherent` entity's memberships directly before falling back to repositories to perform SQL queries. This way we can prevent some extra unneeded SQL queries to the database.

Thus, this design also enables us to add a caching layer on top of the committee manager by wrapping the `CommitteeManager` object in a `CachedCommitteeManager` object if necessary. The main downside of using a decorator here for this use case will be to reimplement all the methods from `CommitteeManager` in the decorator object and forward all method calls.

### DIFF
--- a/src/AppBundle/Collection/CommitteeCollection.php
+++ b/src/AppBundle/Collection/CommitteeCollection.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace AppBundle\Collection;
+
+use AppBundle\Entity\Committee;
+use Doctrine\Common\Collections\ArrayCollection;
+
+final class CommitteeCollection extends ArrayCollection
+{
+    public function getOrderedCommittees(CommitteeMembershipCollection $memberships): self
+    {
+        $uuids = $memberships->getCommitteeSupervisorMemberships()->getCommitteeUuids();
+        $supervisedCommittees = $this->filter(function (Committee $committee) use ($uuids) {
+            return in_array((string) $committee->getUuid(), $uuids);
+        });
+
+        $uuids = $memberships->getCommitteeHostMemberships(CommitteeMembershipCollection::EXCLUDE_SUPERVISORS)->getCommitteeUuids();
+        $hostedCommittees = $this->filter(function (Committee $committee) use ($uuids) {
+            return in_array((string) $committee->getUuid(), $uuids);
+        });
+
+        $uuids = $memberships->getCommitteeFollowerMemberships()->getCommitteeUuids();
+        $followedCommittees = $this->filter(function (Committee $committee) use ($uuids) {
+            return in_array((string) $committee->getUuid(), $uuids);
+        });
+
+        return new static(array_merge(
+            $supervisedCommittees->toArray(),
+            $hostedCommittees->toArray(),
+            $followedCommittees->toArray()
+        ));
+    }
+}
--- a/src/AppBundle/Committee/CommitteeManager.php
+++ b/src/AppBundle/Committee/CommitteeManager.php
@@ -50,50 +50,62 @@ class CommitteeManager
 
     public function isCommitteeHost(Adherent $adherent): bool
     {
-        $memberships = $this->getMembershipRepository()->findMemberships($adherent);
-
-        if (!$memberships->count()) {
-            return false;
+        // Optimization to prevent a SQL query if the current adherent already
+        // has a loaded list of related committee memberships entities.
+        if ($adherent->isHost()) {
+            return true;
         }
 
-        $hostedCommittees = $this->getCommitteeRepository()->findCommittees(
-            $memberships->getCommitteeHostMemberships()->getCommitteeUuids(),
-            CommitteeRepository::ONLY_APPROVED
-        );
-
-        return count($hostedCommittees) >= 1;
+        return $this->getMembershipRepository()->hostCommittee($adherent);
     }
 
-    public function getAdherentCommittees(Adherent $adherent, int $limit = 1000): array
+    public function hostCommittee(Adherent $adherent, Committee $committee): bool
     {
-        $memberships = $this->getMembershipRepository()->findMemberships($adherent);
+        // Optimization to prevent a SQL query if the current adherent already
+        // has a loaded list of related committee memberships entities.
+        if ($adherent->isHostOf($committee)) {
+            return true;
+        }
 
-        if (!$memberships->count()) {
+        return $this->getMembershipRepository()->hostCommittee($adherent, $committee->getUuid());
+    }
+
+    public function superviseCommittee(Adherent $adherent, Committee $committee): bool
+    {
+        // Optimization to prevent a SQL query if the current adherent already
+        // has a loaded list of related committee memberships entities.
+        if ($adherent->isSupervisorOf($committee)) {
+            return true;
+        }
+
+        return $this->getMembershipRepository()->superviseCommittee($adherent, $committee->getUuid());
+    }
+
+    public function getAdherentCommittees(Adherent $adherent): array
+    {
+        // Prevent SQL query if the adherent doesn't follow any committees yet.
+        if (!count($memberships = $adherent->getMemberships())) {
             return [];
         }
 
-        // We want the hosted committees first.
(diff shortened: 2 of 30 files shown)


tgalopin:
@hhamon isn't this finished then?

hhamon:
@tgalopin it's finished and ready for reviews.

tgalopin:
I would like to have a second review as it's a big PR. @mykiwi @rpg600 ?

src/AppBundle/Collection/CommitteeCollection.php
```diff
@@ -0,0 +1,39 @@
...
+use AppBundle\Entity\Committee;
+use Doctrine\Common\Collections\ArrayCollection;
+
+final class CommitteeCollection extends ArrayCollection
+{
+    public function getOrderedCommittees(CommitteeMembershipCollection $memberships): self
+    {
+        $tmp = [];
+        foreach ($this as $committee) {
+            $uuid = (string) $committee->getUuid();
```
HeahDude: This var is needless, it can be inlined just below.

src/AppBundle/Collection/CommitteeCollection.php
```diff
@@ -0,0 +1,39 @@
...
+
+use AppBundle\Entity\Committee;
+use Doctrine\Common\Collections\ArrayCollection;
+
+final class CommitteeCollection extends ArrayCollection
+{
+    public function getOrderedCommittees(CommitteeMembershipCollection $memberships): self
+    {
+        $tmp = [];
+        foreach ($this as $committee) {
```
HeahDude: `$this->elements` would be more performant (micro optim).

src/AppBundle/Collection/CommitteeCollection.php
```diff
@@ -0,0 +1,39 @@
...
+            $uuid = (string) $committee->getUuid();
+            $tmp[$uuid] = $committee;
+        }
+
+        $uuids = $memberships->getCommitteeSupervisorMemberships()->getCommitteeUuids();
```
HeahDude: This code looks ugly and has so many calls... what about:
```php
$supervised = [];
foreach ($memberships->getCommitteeSupervisorMemberships()->getCommitteeUuids() as $uuid) {
    if (isset($tmp[$uuid])) {
        $supervised[] = $tmp[$uuid];
        unset($tmp[$uuid]);
    }
}

// Same for the two others
// ...
return new static(array_merge($supervised, $hosted, $followed));
```

src/AppBundle/Collection/CommitteeCollection.php
```diff
@@ -0,0 +1,39 @@
...
+namespace AppBundle\Collection;
+
+use AppBundle\Entity\Committee;
+use Doctrine\Common\Collections\ArrayCollection;
+
+final class CommitteeCollection extends ArrayCollection
+{
+    public function getOrderedCommittees(CommitteeMembershipCollection $memberships): self
+    {
+        $tmp = [];
```
HeahDude: This var seems unused.

src/AppBundle/Committee/CommitteeManager.php
```diff
@@ -223,6 +251,15 @@ public function refuseCommittee(Committee $committee, bool $flush = true): void
         }
     }
 
+    public function isFollowingCommittee(Adherent $adherent, Committee $committee): bool
+    {
+        if (!$membership = $this->getCommitteeMembership($adherent, $committee)) {
```
HeahDude: Simply `return $this->getCommitteeMembership($adherent, $committee)`?

src/AppBundle/Entity/CommitteeMembership.php
```diff
@@ -236,4 +236,13 @@ public function getSubscriptionDate(): \DateTimeImmutable
...
         return new \DateTimeImmutable($this->joinedAt->format(DATE_RFC822), $this->joinedAt->getTimezone());
     }
+
+    public function matches(Adherent $adherent, Committee $committee): bool
+    {
+        if (!$adherent->equals($this->getAdherent())) {
+            return false;
+        }
+
+        return $this->committeeUuid->equals($committee->getUuid());
```
rpg600: `return ... && ...`  ?

src/AppBundle/Committee/Voter/PostMessageCommitteeVoter.php
```diff
@@ -23,6 +23,10 @@ protected function supports($attribute, $committee)
 
     protected function doVoteOnAttribute(string $attribute, Adherent $adherent, Committee $committee): bool
     {
-        return $this->repository->hostCommittee($adherent, (string) $committee->getUuid()) && $committee->isApproved();
+        if (!$committee->isApproved()) {
```
HeahDude: Could be:
```php
return $committee->isApproved() && $this->manager->hostCommittee($adherent, $committee);
```

src/AppBundle/Entity/Adherent.php
```diff
@@ -588,34 +589,50 @@ public function setProcurationManagedAreaCodesAsString(string $codes = null): vo
         $this->procurationManagedArea->setCodesAsString((string) $codes);
     }
 
-    /**
-     * @return CommitteeMembership[]|Collection
-     */
-    public function getMemberships()
+    final public function getMemberships(): CommitteeMembershipCollection
     {
+        if (null === $this->memberships) {
```
HeahDude: This code looks broken as it would decorate the collection on each call, try instead:
```php
if ($this->memberships instanceof Collection) {
    if (!$this->memberships instanceof CommitteeMembershipCollection) {
        $this->memberships = new CommitteeMembershipCollection($this->memberships->toArray());
    }
} else {
    $this->memberships = new CommitteeMembershipCollection((array) $this->memberships);
}

return $this->memberships;
```

src/AppBundle/Entity/Adherent.php
```diff
@@ -588,34 +589,50 @@ public function setProcurationManagedAreaCodesAsString(string $codes = null): vo
...
         return $this->memberships;
     }
 
-    public function isHost(): bool
+    public function getMembershipFor(Committee $committee): ?CommitteeMembership
     {
-        foreach ($this->memberships as $membership) {
-            if ($membership->canHostCommittee()) {
-                return true;
+        foreach ($this->getMemberships() as $membership) {
```
HeahDude: This function call looks useless, why do you need the special collection for?

src/AppBundle/Entity/CommitteeMembership.php
hhamon: Why not!

src/AppBundle/Collection/CommitteeCollection.php
hhamon: True! thanks.

src/AppBundle/Committee/CommitteeManager.php
hhamon: The method must return a `bool` value not a `CommitteeMembership` instance.

HeahDude: `return (bool) $this->getCommitteeMembership($adherent, $committee)`?

src/AppBundle/Entity/Adherent.php
hhamon: True, it's not needed.

hhamon:
PR has been updated according to your comments.

src/AppBundle/Collection/CommitteeCollection.php
hhamon: it's a private var.

PR #707, opened 2017-03-30
Update je marche page

marineb:


tgalopin:
I find it difficult to read white on green, and moreover there is also black text on the same block: what do you think of putting the main title/subtitle in black too?

marineb:
Yeah. You're totally right. I need to think to deactivate .flux when I work at night. I also feel the same way about the committee map button on the committee search page. I've been meaning to update it so the text is black too.

PR #738, opened 2017-04-03
Add navigation for clarification page

marineb:


app/Resources/views/page/emmanuel-macron/nav.html.twig
```diff
@@ -0,0 +1,12 @@
+{% set currentRoute = app.request.attributes.get('_route') %}
+{% for route, title in {
+    'page_emmanuel_macron': 'Ce que je suis',
+    'page_emmanuel_macron_revolution': 'Révolution',
+    'page_emmanuel_macron_desintox_list': 'Désintox'
```
tgalopin: Could you keep the `{% if enable_canary %}` part so we can put this in production without issue?

PR #742, opened 2017-04-03
Committee hosts and supervisors are not allowed to leave their committees anymore.

hhamon:


app/Resources/views/committee/show.html.twig
```diff
@@ -131,6 +133,16 @@
...
+        {% endif %}
+
+        {% if is_host %}
+            <button role="button"
+                    class="btn btn--no-border btn--text-blue committee-unfollow btn--no-wrap"
+                    id="committee-toggle-membership"
+                    disabled="disabled"
+                    title="En tant qu'animateur, vous ne pouvez pas cesser de suivre votre comité">
+                Quitter ce comité
+            </button>
```
tgalopin: Can you also add the same kind of text than on line https://github.com/EnMarche/en-marche.fr/pull/742/files#diff-5f04b136c3d27995986e6b2fbc7da527R114 ?

PR #743, opened 2017-04-03
Send a confirmation email to the user who registered to an event.

hhamon:


src/AppBundle/Mailjet/Message/EventRegistrationConfirmationMessage.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace AppBundle\Mailjet\Message;
+
+use AppBundle\Entity\EventRegistration;
+use Ramsey\Uuid\Uuid;
+
+final class EventRegistrationConfirmationMessage extends MailjetMessage
```
rpg600: There are unit tests for this kind of class (in Mailjet/Message). Can you add one ?

hhamon: Unit tests have been added.

PR #747, opened 2017-04-03
Added committee link or organiser in event info on the map

HeahDude:


front/pages/events_map.js
```diff
@@ -22,8 +22,16 @@ export default (mapFactory, api) => {
                     infowindow.close();
                 }
 
+                let content = `<a href="${event.url}" target="_blank">${event.name}</a><br>Organisé par`;
+
+                if ('committee_url' in event) {
+                    content += `<a href="${event.committee_url}" target="_blank">${event.committee_name}</a>`;
+                } else {
+                    content += '&nbsp;' + event.organizer;
```
mykiwi: what about move `&nbsp;` in line 25 since line 28 it is missing (I guess) ?

HeahDude: Hmm, we've discussed it with Titouan, the space is only needed here.

tgalopin: I guess it comes from the templating syntax. Let's keep it like that, it's fine.

src/AppBundle/Api/EventProvider.php
```diff
@@ -38,6 +38,20 @@ public function getUpcomingEvents(): array
                     'lng' => (float) $event->getLongitude(),
                 ],
             ];
+
+            if ($committee = $event->getCommittee()) {
+                $el += [
+                    'committee_name' => $committee->getName(),
+                    'committee_url' => $this->urlGenerator->generate('app_committee_show', [
```
hhamon: You could instead inject the `CommitteeUrlGenerator` which has a dedicated method for generating a path by passing a committee object.

src/AppBundle/Repository/EventRepository.php
```diff
@@ -130,11 +130,9 @@ private function createUpcomingEventsQueryBuilder(): QueryBuilder
         return $qb
             ->leftJoin('e.committee', 'c')
             ->andWhere($qb->expr()->in('e.status', Event::ACTIVE_STATUSES))
-            ->andWhere('c.status = :committee_status')
```
hhamon: Why do you remove this line?

src/AppBundle/Repository/EventRepository.php
```diff
@@ -130,11 +130,9 @@ private function createUpcomingEventsQueryBuilder(): QueryBuilder
...
             ->andWhere('e.beginAt >= :today')
             ->orderBy('e.beginAt', 'ASC')
             ->setParameter('today', date('Y-m-d'))
-            ->setParameter('committee_status', Committee::APPROVED)
```
hhamon: Why do you remove this line?

src/AppBundle/Repository/EventRepository.php
HeahDude: This is a bug fix with its dedicated commit.

hhamon: I don't understand why because we want to fetch events that are related to active committees.

src/AppBundle/Api/EventProvider.php
HeahDude: I don't think we need to inject this decorator we already have the router to generate the other url.

hhamon: Oh yes true!

src/AppBundle/Repository/EventRepository.php
hhamon: oh I didn't know we can create events that are not linked to a committee.

PR #752, opened 2017-04-04
Add event types filters on the map.

hhamon:


src/AppBundle/Api/EventProvider.php
```diff
@@ -16,11 +18,14 @@ public function __construct(EventRepository $repository, UrlGeneratorInterface $
         $this->urlGenerator = $urlGenerator;
     }
 
-    public function getUpcomingEvents(): array
+    public function getUpcomingEvents(string $type = null): array
```
HeahDude: Please add a `@throws` phpdoc.

src/AppBundle/Api/EventProvider.php
```diff
@@ -16,11 +18,14 @@ public function __construct(EventRepository $repository, UrlGeneratorInterface $
...
     {
-        $data = [];
+        if ($type && !EventCategories::getCategoryName($type)) {
+            throw new EventException(sprintf('Given event category "%s" is invalid.', $type));
+        }
 
-        foreach ($this->repository->findUpcomingEvents() as $event) {
+        $data = [];
```
HeahDude: In PHP 7 a micro optim is to do:
```php
return $data ?? [];
```
and remove this memory call which might be needless.

src/AppBundle/Controller/Api/CommitteesController.php
```diff
@@ -25,8 +27,12 @@ public function getApprovedCommitteesAction()
      * @Route("/events", name="api_committees_events")
      * @Method("GET")
      */
-    public function getUpcomingCommitteesEventsAction()
+    public function getUpcomingCommitteesEventsAction(Request $request)
     {
-        return new JsonResponse($this->get('app.api.event_provider')->getUpcomingEvents());
+        try {
+            return new JsonResponse($this->get('app.api.event_provider')->getUpcomingEvents($request->query->get('type')));
+        } catch (\Exception $e) {
```
HeahDude: Please catch `EventException` only, we don't want bad request code for anything else.

src/AppBundle/Api/EventProvider.php
hhamon: Fixed. Thanks.

src/AppBundle/Api/EventProvider.php
hhamon: Fixed. Thanks.

src/AppBundle/Controller/Api/CommitteesController.php
hhamon: Fixed. Thanks.

src/AppBundle/Api/EventProvider.php
```diff
@@ -16,11 +18,16 @@ public function __construct(EventRepository $repository, UrlGeneratorInterface $
...
     }
 
-    public function getUpcomingEvents(): array
+    /**
+     * @throws EventException
+     */
+    public function getUpcomingEvents(string $type = null): array
     {
-        $data = [];
+        if ($type && !EventCategories::getCategoryName($type)) {
```
HeahDude: Why testing it? Maybe this should not be nullable in the first place?

hhamon: because `null` means I want to all events no matter the category is.

PR #766, opened 2017-04-04
[Admin] improve list view for events admin.

hhamon:


app/Resources/views/admin/event_actions.html.twig
```diff
@@ -0,0 +1,22 @@
+{% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
+
+{% block field%}
```
HeahDude: A space is missing in all templates.

hhamon: Fixed all missing spaces.

PR #774, opened 2017-04-05
Fixed newsletter invitation form validation

HeahDude:


app/Resources/views/newsletter/invitation.html.twig
```diff
@@ -29,7 +29,7 @@
                 </p>
 
                 <div class="newsletter-page__form l__wrapper-500">
-                    {{ form_start(invitation_form) }}
+                    {{ form_start(invitation_form, {attr: {novalidate: 'novalidate'}}) }}
```
tgalopin: I would much prefer to change the FormType to remove the "required" attribute. Is there a reason not to do so?

HeahDude: I think there are indeed.

First, validation is not consistent across browsers and is done server side; second, from a FormType perspective the tree is resized dynamically depending on the submitted data and according to this point of view every field is required and must be validated, changing the configuration of children would expose potential side effects on validation and empty data handling, I would not advise it.

tgalopin:
It's fine for me, let's merge this

PR #785, opened 2017-04-06
Added TonMacron Doctrine entity models.

hhamon:


src/AppBundle/Entity/TonMacronFriendInvitation.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public function getMailBody(): ?string
+    {
+        return $this->mailBody;
+    }
+
+    public function getCreatedAt(): \DateTimeImmutable
+    {
+        if ($this->createdAt instanceof \DateTimeImmutable) {
```
HeahDude: This will never be true after Doctrine loaded the entity, instead you can do:
```php
if (!$this->createdAt instanceof \DateTimeImmutable) {
    $this->createdAt = \DateTimeImmutable::createFromMutable($this->createdAt);
}

return $this->createdAt;
```

hhamon: Fixed.

PR #794, opened 2017-04-06
Add Ton Macron administration

rpg600:


src/AppBundle/Admin/TonMacronFriendInvitationAdmin.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace AppBundle\Admin;
+
+use Sonata\AdminBundle\Admin\AbstractAdmin;
+use Sonata\AdminBundle\Datagrid\ListMapper;
+use Sonata\AdminBundle\Show\ShowMapper;
+
+class TonMacronFriendInvitationAdmin extends AbstractAdmin
+{
```
tgalopin: Can you put ordered by createdAt by default?

src/AppBundle/Admin/TonMacronChoiceAdmin.php
```diff
@@ -0,0 +1,57 @@
...
+
+use AppBundle\Entity\TonMacronChoice;
+use Sonata\AdminBundle\Admin\AbstractAdmin;
+use Sonata\AdminBundle\Datagrid\ListMapper;
+use Sonata\AdminBundle\Form\FormMapper;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+
+class TonMacronChoiceAdmin extends AbstractAdmin
+{
+    protected function configureFormFields(FormMapper $formMapper)
```
tgalopin: Can you put ordered by createdAt by default?

rpg600: We don't have createdAt field, do i need to add one ?

tgalopin: No it's fine

src/AppBundle/Admin/TonMacronFriendInvitationAdmin.php
hhamon: I think it's better to order by step number first and sequence (content key) number then. Thus we can disable pagination for such a list. This way we could override the list template to have an output similar to the following one:

````
Étape 0 : choix hors étapes du tunnel Ton Macron
---------------------------------------------------------

Label | Contenu
------------ | -------------
Introduction | Contenu de l'introduction
Conclusion | Contenu de la conclusion

Étape 1 : situation personnelle de l'ami
---------------------------------------------

Label | Contenu
------------ | -------------
A | Contenu de A
B | Contenu de B
C | Contenu de C

Étape 2 : projet de l'ami
---------------------------

Label | Contenu
------------ | -------------
D | Contenu de D
E | Contenu de E
F | Contenu de F
```

tgalopin: That's a very good idea, let's do this!

tgalopin: Actually, it's not the right entity. You're right and we should do this on TonMacronChoiceAdmin.php, but this entity should be ordered by createdAt DESC.

PR #797, opened 2017-04-06
Added services to build and send a TonMacron invitation message to a friend

hhamon:


app/Resources/views/campaign/ton_macron.html.twig
```diff
@@ -0,0 +1,15 @@
...
+<ul>
+{% for argument in invitation.arguments %}
+    {{- argument|raw }}
+{% endfor %}
+</ul>
+
+{{ conclusion.content|raw }}
+
+<p>
+{{ invitation.authorFirstName|e('html') }}
```
HeahDude: Is the escape filter really needed here?

hhamon: Yes because author first name is filled by the user in the form. It's an input data.

PR #811, opened 2017-04-07
Send a procuration reminder

mykiwi:


src/AppBundle/Command/SendMailToProcurationProxiesCommand.php
```diff
@@ -0,0 +1,51 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->em = $this->getContainer()->get('doctrine.orm.entity_manager');
+        $this->mailjet = $this->getContainer()->get('app.mailjet.client.transactional');
+        $this->messageFactory = $this->getContainer()->get('app.procuration.proxy_message_factory');
+
+        $this->sendMails();
+    }
+
+    private function sendMails($offset = 0, $limit = 100)
```
HeahDude: Shouldn't this be a service with the proper injections? Does it worth having unused parameters on this private method?

src/AppBundle/Procuration/ProcurationProxyMessageFactory.php
```diff
@@ -40,4 +41,17 @@ public function createProxyFoundMessage(Adherent $procurationManager, Procuratio
...
     }
+
+    public function createProxyReminderMessage(ProcurationRequest $request): ProcurationProxyReminderMessage
+    {
+        $url = $this->urlGenerator->generate('app_procuration_my_request', [
+            'id' => $request->getId(),
+            'token' => $request->generatePrivateToken(),
+        ]);
+
+        $message = ProcurationProxyReminderMessage::create($request, $request->getFoundProxy(), $url, $request->getFoundBy());
```
HeahDude: There is no need to pass the extra arguments which depend on the request object, just pass the request and use it to get its values in the static method instead, or is there other usages I'm missing?

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -320,7 +335,7 @@ private static function createPhoneNumber(int $countryCode = 33, string $number
      *
      * @param ExecutionContextInterface $context
      */
-    public function validateElectionsChosen(ExecutionContextInterface $context)
+    public function validateElectionsChosen(ExecutionContextInterface $context): void
```
hhamon: `validateChosenElections`

src/AppBundle/Command/SendMailToProcurationProxiesCommand.php
mykiwi: It could, be it will (and should) never be used anywhere else but here.  
I used this parameters.

src/AppBundle/Procuration/ProcurationProxyMessageFactory.php
mykiwi: I know, it's just that `ProcurationProxyFoundMessage` and `ProcurrationProxyCancelledMessage` are done like this, so I did it also.

hhamon: Agreeed with @HeahDude. Just pass the arguments the method requires.

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -273,6 +281,13 @@ class ProcurationRequest
     private $processedAt;
 
     /**
+     * @var bool
+     *
+     * @ORM\Column(type="boolean")
+     */
+    private $reminded = false;
```
hhamon: Instead of a boolean value, I would better like to see a counter. This way we can force to resend a reminder several times.

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -47,6 +47,14 @@ class ProcurationRequest
     private $foundProxy;
 
     /**
+     * The user who associated the found proxy.
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Adherent")
+     * @ORM\JoinColumn(name="procuration_request_found_by_id", referencedColumnName="id", nullable=true)
```
HeahDude: Nullable is not need here, it is true by default for a join column.

src/AppBundle/Entity/ProcurationRequest.php
HeahDude: This could be a good idea but do we really want to support it, we could skip setting it to false or reset it to true to use it again when needed. The real use case is to initially program many.

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -320,7 +335,7 @@ private static function createPhoneNumber(int $countryCode = 33, string $number
      *
      * @param ExecutionContextInterface $context
      */
-    public function validateElectionsChosen(ExecutionContextInterface $context)
+    public function validateChosenElections(ExecutionContextInterface $context): void
```
HeahDude: @hhamon this required change was not related to this PR, or should we also change the constants name to `..._REASON` :trollface:?

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -341,7 +356,7 @@ public function validateElectionsChosen(ExecutionContextInterface $context)
         $context->addViolation('Vous devez choisir au moins une élection');
     }
 
-    public function importAdherentData(Adherent $adherent)
+    public function importAdherentData(Adherent $adherent): void
```
HeahDude: Please try to make other commits for cs fixes to ease the review with a thinner diff.

src/AppBundle/Mailjet/Message/ProcurationProxyReminderMessage.php
```diff
@@ -0,0 +1,34 @@
...
+namespace AppBundle\Mailjet\Message;
+
+use AppBundle\Entity\ProcurationRequest;
+use Ramsey\Uuid\Uuid;
+
+final class ProcurationProxyReminderMessage extends MailjetMessage
+{
+    public static function create(ProcurationRequest $request, string $infosUrl): self
+    {
+        $message = new static(
```
HeahDude: This should be new self since the class is final.

src/AppBundle/Entity/ProcurationRequest.php
hhamon: Just because it's better English ;)

src/AppBundle/DataFixtures/ORM/LoadProcurationData.php
```diff
@@ -206,5 +206,17 @@ public function load(ObjectManager $manager)
...
+        $manager->refresh($proxy1);
+        $request1->process($proxy1, $manager->getRepository(Adherent::class)->findByUuid(LoadAdherentData::ADHERENT_4_UUID));
+
+        $reflectionClass = new \ReflectionClass(ProcurationRequest::class);
+        $reflectionProperty = $reflectionClass->getProperty('processedAt');
+        $reflectionProperty->setAccessible(true);
+        $reflectionProperty->setValue($request1, new \DateTime('-48 hours'));
+
+        $manager->flush();
+        dump($request1, $proxy1);
```
hhamon: debug statement...

mykiwi: Because there is a bug...

src/AppBundle/Entity/ProcurationRequest.php
mykiwi: No, you can do it in a other PR ;)

src/AppBundle/Entity/ProcurationRequest.php
mykiwi: done

src/AppBundle/Command/SendMailToProcurationProxiesCommand.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $this->limit = is_numeric($input->getOption(self::MAILS_PER_LOOP))
+            ? intval($input->getOption(self::MAILS_PER_LOOP))
+            : self::MAILS_PER_LOOP_DEFAULT_VALUE;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        for ($i = 0, $total = 0; ; ++$i) {
+            $requests = $this->manager->getRepository(ProcurationRequest::class)
```
hhamon: Can you get the repository before the `for()` loop to prevent many method calls to the manager.

src/AppBundle/Mailjet/Message/ProcurationProxyReminderMessage.php
```diff
@@ -0,0 +1,34 @@
...
+            [
+                'target_firstname' => self::escape($request->getFirstNames()),
+                'voter_first_name' => $request->getFoundProxy()->getFirstNames(),
+                'voter_last_name' => $request->getFoundProxy()->getLastName(),
+                'info_link' => $infosUrl,
+            ]
+        );
+
+        $message->setSenderName('Procuration Macron');
+        if ($request->getFoundBy()) {
```
hhamon:
```php
if ($foundBy = $request->getFoundBy()) {
    $message->addCC($foundBy->getEmailAddress());
}
```

src/AppBundle/Mailjet/Message/ProcurationProxyReminderMessage.php
```diff
@@ -0,0 +1,34 @@
...
+            $message->addCC($request->getFoundBy()->getEmailAddress());
+        }
+        $message->addCC($request->getFoundProxy()->getEmailAddress());
```
hhamon: You should store the `foundProxy` object in a temp variable as you're making multiple `->getFoundProxy()` method calls. It'll also help the readability.

hhamon:
Good to go after squash.

tgalopin:
@mykiwi can you squash and merge?

app/migrations/Version20170411110645.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20170411110645 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE procuration_requests ADD procuration_request_found_by_id INT UNSIGNED DEFAULT NULL, ADD reminded INT NOT NULL');
```
HeahDude: Adding a not null column without default may break when migrating.

src/AppBundle/Command/SendMailToProcurationProxiesCommand.php
```diff
@@ -0,0 +1,71 @@
...
+     * @var \AppBundle\Procuration\ProcurationReminderHandler
+     */
+    private $reminder;
+
+    /**
+     * @var int
+     */
+    private $limit;
+
+    public const COMMAND_NAME = 'app:mail:send-reminder-to-procuration-proxy';
```
HeahDude: Public is not needed, it's the default.

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -684,8 +702,38 @@ public function getFoundProxy(): ?ProcurationProxy
...
+    {
+        return $this->foundBy;
+    }
+
+    public function setFoundBy(?Adherent $foundBy): void
+    {
+        $this->foundBy = $foundBy;
+    }
+
+    public function getReminded(): int
```
HeahDude: Is this getter used? I would have liked "addReminder`/`remind` instead as increase/decrease methods and kept the haser.

mykiwi: No

src/AppBundle/Command/SendMailToProcurationProxiesCommand.php
mykiwi: I don't mind remove it, but it's cleaner with it IMHO.

app/migrations/Version20170411110645.php
mykiwi: Good catch // ping @tgalopin

PR #838, opened 2017-04-09
Updates to procuration flow

marineb:


app/Resources/views/procuration/proposal_thanks.html.twig
```diff
@@ -16,13 +16,45 @@
...
             Votre formulaire a bien été envoyé.<br />
             En cas de demande de procuration ayant lieu dans votre commune, votre responsable local vous contactera.
         </p>
 
+
+        <div class="space--30-0">
+            <div class="text--medium text--bold b__nudge--bottom">
+                Proposez à vos amis de devenir mandataires !
+            </div>
+            {#
```
tgalopin: You can remove this instead of putting it in comments

marineb: It's already hard to get rid of stuff :simple_smile:

PR #839, opened 2017-04-09
Fix positioning of bercy banner

marineb:


app/Resources/views/banner/bercy.html.twig
```diff
@@ -1,7 +1,8 @@
-<div class="bercy__banner l__hide--on-tablet progressive-background" id="header-banner"
-     style="display: none;"
-     data-sd="{{ static_asset('bercy-banner.jpg', {'w': 300, 'q': 20}) }}"
-     data-hd="{{ static_asset('bercy-banner.jpg', {'w': 2000, 'q': 70}) }}">
+<div class="bercy__banner l__hide--on-tablet" id="header-banner"
+     style="display: none;">
+    <img src="{{ static_asset('bercy-banner.jpg', {'w': 638, 'q': 70}) }}"
+          title="S'inscrire au rendez-vous du 17 Avril à Bercy !"
+          alt="S'inscrire au rendez-vous du 17 Avril à Bercy !" />
```
tgalopin: Why this instead of a background?

marineb: @tgalopin because I'm really dumb. PR updated. 😉

PR #845, opened 2017-04-09
Add procuration landing

marineb:


src/AppBundle/Controller/ProcurationController.php
```diff
@@ -23,7 +23,16 @@
 class ProcurationController extends Controller
 {
     /**
-     * @Route("", name="app_procuration_index")
+     * @Route("", name="app_procuration_landing")
```
HeahDude: You can just use `@Route(name="app_procuration_landing")`.

PR #857, opened 2017-04-10
Fixes wording for procuration details.

hhamon:


app/Resources/views/procuration/my_request.html.twig
```diff
@@ -11,6 +11,28 @@
         </h2>
 
         <div class="text--center b__nudge--bottom-large">
+            Cette procuration concerne le(s) tour(s) de scrutin suivant(s) :
```
HeahDude: You could translate depending on the count:
```twig
{% set rounds_count = request.electionPresidentialFirstRound + request.electionPresidentialSecondRound + request.electionLegislativeFirstRound + request.electionLegislativeSecondRound %}

Cette procuration concerne {{ 'le tour de scrutin suivant|les tours de scrutins suivants'|transchoice(rounds_count) }} :
```

hhamon: Good catch, I'll do it! Thanks.

PR #871, opened 2017-04-11
Passed events for less than 36 hours are kept on the events page.

hhamon:


src/AppBundle/Repository/CommitteeRepository.php
```diff
@@ -227,4 +228,33 @@ private function createSitemapQb(): QueryBuilder
...
+            $qb = $this
+                ->createNearbyQueryBuilder($coordinates)
+                ->andWhere($this->getNearbyExpression().' < :distance_max')
+                ->setParameter('distance_max', $search->getRadius());
+        } else {
+            $qb = $this->createQueryBuilder('c');
+        }
+
+        if (!empty($query = $search->getQuery())) {
+            $qb->andWhere('n.name like :query');
```
mykiwi: Is it possible to get inside the previous else and inside this if ? If yes, you may have an issue since the alias is `c` and not `n`.

hhamon: True that's a bug. I just did refactor the code.

hhamon: Fixed. Thanks.

mykiwi: I prefer create the builder builder inside the method and pass it through parameter if needed, like this you are sure of what you have.

hhamon: Yes me too! Again I just moved this code to the repository because it was outside of it. But I agree it would be better to create the builder and pass it as an argument of repository private methods.

PR #872, opened 2017-04-11
Fixes warning message display when it's not needed.

hhamon:


src/AppBundle/Form/AbstractProcurationType.php
```diff
@@ -98,7 +98,7 @@ public function onPreSubmit(FormEvent $event): void
         // not there are already existing procuration requests with the same
         // email address. If yes, the user will have to confirm his choice.
         if ('new_request' === $request['state']) {
-            if ($mustConfirm = count($this->matchEmailAddress($emailAddress)) > 0) {
+            if ($mustConfirm = $this->matchEmailAddress($emailAddress)) {
```
mykiwi: if `$mustConfirm` is never used, you can remove it.

hhamon: true! it used to be used! Thanks.

PR #880, opened 2017-04-12
Add videos page

marineb:
@tgalopin: 
* You'll see I commented around the code I think should be managed in the CMS (would be great if we kept control over the outer part, where I've set a few settings). 
* After having looked closer, I think the text within the `p` tags is only present for SEO/accessibility. We should still make sure it's filled, but makes it easier I think. That means we can have a field for the description, one for the fb link (which appears in 2 places), and that's it [see my note in the second video]. 
* When I use our app ID, the videos weren't loading wasn't working. I wonder if it must be using a different type of app id (so right now, I'm using an app ID that belongs to me) -- anyway it's easy to figure out which app ID it should be: someone can view [this page](https://developers.facebook.com/docs/plugins/embedded-video-player/?prefill_href=https%3A%2F%2Fwww.facebook.com%2FEnMarche%2Fvideos%2F769283806583157%2F#) and will be able to see the app id as an option (someone who has access to the En Marche account)


tgalopin:
Let's wait for @rpg600 to create the backend before merging this

rpg600:
My changes have been added.

PR #881, opened 2017-04-12
add procuration request filter

mykiwi:


app/Resources/views/procuration_manager/_filters.html.twig
```diff
@@ -1,4 +1,4 @@
-<form method="get" action="{{ path('app_procuration_manager_index') }}">
+<form method="get" action="{{ path(app.request.get('_route', '_route_params')) }}">
```
tgalopin: I don't understand this line: why not simply no action attribute?

mykiwi: Yes, I can remove the action attribute.

tgalopin:
Can we merge this?

src/AppBundle/Search/ProcurationParametersFilter.php
```diff
@@ -0,0 +1,86 @@
...
+
+    public function getCountries(): array
+    {
+        return UnitedNationsBundle::getCountries();
+    }
+
+    public function hasData(): bool
+    {
+        return $this->country || $this->city || $this->type;
+    }
```
HeahDude: The should have a method `apply(QueryBuilder $builder)` to avoid ugly private methods in repository.

PR #885, opened 2017-04-12
add elections dates in producration mails

mykiwi:
Can't use it in [`app/Resources/views/procuration_manager/_requests_list.html.twig`](https://github.com/EnMarche/en-marche.fr/blob/master/app/Resources/views/procuration_manager/_requests_list.html.twig#L22-L36) since it is an array.

### DIFF
--- a/src/AppBundle/Mailjet/Message/ProcurationProxyFoundMessage.php
+++ b/src/AppBundle/Mailjet/Message/ProcurationProxyFoundMessage.php
@@ -26,6 +26,7 @@ final class ProcurationProxyFoundMessage extends MailjetMessage
                 'voter_first_name' => $proxy->getFirstNames(),
                 'voter_last_name' => $proxy->getLastName(),
                 'info_link' => $infosUrl,
+                'elections' => implode(', ', $request->getElections()),
             ]
         );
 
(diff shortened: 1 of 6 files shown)


src/AppBundle/Mailjet/Message/ProcurationProxyFoundMessage.php
```diff
@@ -26,6 +26,7 @@ public static function create(
                 'voter_first_name' => $proxy->getFirstNames(),
                 'voter_last_name' => $proxy->getLastName(),
                 'info_link' => $infosUrl,
+                'elections' => implode(' ', $request->getElections()),
```
tgalopin: Could you put a comma instead in the implode?

src/AppBundle/Mailjet/Message/ProcurationProxyReminderMessage.php
```diff
@@ -21,6 +21,7 @@ public static function create(ProcurationRequest $request, string $infosUrl): se
                 'voter_first_name' => $proxy->getFirstNames(),
                 'voter_last_name' => $proxy->getLastName(),
                 'info_link' => $infosUrl,
+                'elections' => implode(' ', $request->getElections()),
```
tgalopin: Could you put a comma instead in the implode?

PR #891, opened 2017-04-12
Implement a custom TonMacron export

tgalopin:


src/AppBundle/TonMacron/TonMacronSerializer.php
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @return string
+     */
+    public function serializeChoices(array $choices): string
+    {
+        if (!is_iterable($choices)) {
+            throw new \InvalidArgumentException();
+        }
+
+        $handle = fopen('php://memory', 'rb+');
```
HeahDude: There is a [CsvEncoder](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Encoder/CsvEncoder.php) since symfony 3.2 ;).

tgalopin: I won't do this right now as I have a lot to do, but feel free to improve this tomorrow :) .

PR #897, opened 2017-04-12
Always return default image in Algolia thumbnails

tgalopin:


HeahDude:
Can we delete the branch here?

PR #901, opened 2017-04-13
Update/procurations

marineb:


src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -272,6 +272,13 @@ class ProcurationRequest
     /**
      * @var bool
      *
+     * @Assert\IsTrue(message="procuration.authorization.required", groups={"front"})
+     */
+    public $authorization;
```
HeahDude: We do not need such public property on the entity, we can use the `mapped` option set to `false` on the form fields instead.

marineb:
@HeahDude Made an update. Let me know if that's what you meant / work for you!

tgalopin:
I think that's not what he meant. Can you either explain more or even add a commit in the PR @HeahDude ?

HeahDude:
Done, I've just pushed https://github.com/EnMarche/en-marche.fr/pull/901/commits/2972501105ce46527fcb481fd7ecf8d04c0fc974.

HeahDude:
Ok, it's green now, I've fixed a bug in validation groups while I was at it ;).

PR #906, opened 2017-04-13
clean procuration filter

mykiwi:
Based on https://github.com/EnMarche/en-marche.fr/pull/881#discussion_r111247876

### DIFF
--- a/src/AppBundle/Repository/ProcurationProxyRepository.php
+++ b/src/AppBundle/Repository/ProcurationProxyRepository.php
@@ -49,7 +49,7 @@ class ProcurationProxyRepository extends EntityRepository
         $this->addAndWhereManagedBy($qb, $procurationManager);
 
         if ($filters) {
-            $this->applyFilter($qb, $filters);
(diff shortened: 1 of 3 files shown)


src/AppBundle/Repository/ProcurationProxyRepository.php
```diff
@@ -49,7 +49,7 @@ public function findManagedBy(Adherent $procurationManager, int $page, int $perP
         $this->addAndWhereManagedBy($qb, $procurationManager);
 
         if ($filters) {
```
HeahDude: I would use a `$filter->hasQuery()` method so the filter can say if it has values to give to the builder.

mykiwi: ?  
The parameter `$filters` is optional, so I don't get it.

HeahDude: Ok, forget it, this is indeed needless here.

PR #909, opened 2017-04-13
Make upcoming events only in the events list

hhamon:


hhamon:
This is still WIP for now.

src/AppBundle/Repository/EventRepository.php
```diff
@@ -182,7 +182,7 @@ public function searchEvents(SearchParametersFilter $search): array
                 ->andWhere($this->getNearbyExpression().' < :distance_max')
                 ->andWhere('n.beginAt > :now')
                 ->setParameter('distance_max', $search->getRadius())
-                ->setParameter('now', new \DateTime('-36 hours'))
```
mykiwi: Why remove this?

tgalopin: It's the issue, that was a misexplaination in the US

hhamon: To fix the problem!

src/AppBundle/Controller/EventController.php
```diff
@@ -61,6 +61,10 @@ public function exportIcalAction(Event $event): Response
      */
     public function attendAction(Request $request, Event $event): Response
     {
+        if ($event->isFinished()) {
+            throw $this->createNotFoundException(sprintf('Event "%s" is finished and does not accept registrations anymore', $event->getUuid()));
```
rpg600: Duno if it's relevant to do a functional test for this case.

tgalopin: Let's do this in a new PR.

PR #915, opened 2017-04-13
Take events located in foreign countries and different timezones into account in the lists of my past and upcoming events.

hhamon:


hhamon:
I'm adding some additional tests to this PR before merging it.

src/AppBundle/Entity/EventRegistration.php
```diff
@@ -104,6 +105,15 @@ public function isNewsletterSubscriber(): bool
         return $this->newsletterSubscriber;
     }
 
+    public function isEventFinished(): bool
+    {
+        if (!$this->event instanceof Event) {
```
HeahDude: Useless the `Event` object is required by the constructor, am I missing something?

src/AppBundle/Repository/EventRegistrationRepository.php
```diff
@@ -66,26 +67,39 @@ public function findGuestRegistration(string $eventUuid, string $emailAddress):
...
     {
-        return $this
+        $registrations = $this
             ->createAdherentEventRegistrationQueryBuilder($adherentUuid)
-            ->andWhere('e.beginAt >= :now')
-            ->setParameter('now', date('Y-m-d H:i:s'))
+            ->andWhere('e.beginAt >= :begin')
+            // The extra 24 hours enable to include events in foreign
+            // countries that are on different timezones.
+            ->setParameter('begin', new \DateTime('-24 hours'))
```
HeahDude: 13 hours should be enough, otherwise the date won't the same.

src/AppBundle/Repository/EventRegistrationRepository.php
```diff
@@ -66,26 +67,39 @@ public function findGuestRegistration(string $eventUuid, string $emailAddress):
...
             ->orderBy('e.beginAt', 'ASC')
             ->getQuery()
             ->getResult()
         ;
+
+        return iterator_to_array($this->createEventRegistrationCollection($registrations)->getUpcomingRegistrations());
```
HeahDude: Why not using `toArray()`?

src/AppBundle/Entity/EventRegistration.php
hhamon: Look at the JoinColumn constraint. In case of the event's deletion, the registration record is kept and the foreign key is set to null.

src/AppBundle/Repository/EventRegistrationRepository.php
hhamon: I want to be large as we don't fetch too many data.

src/AppBundle/Repository/EventRegistrationRepository.php
hhamon: I can! Just didn't think about it.

src/AppBundle/Entity/EventRegistration.php
tgalopin: I gues he meant `if (!$this->event)`

hhamon: Yep It's the same.

PR #930, opened 2017-04-14
Add access to the legislative candidates platform

tgalopin:


app/Resources/views/admin/adherent_status.html.twig
```diff
@@ -5,11 +5,15 @@
         {% if object.isReferent %}
             <span class="label label-primary" style="padding: 2px;">Référent</span><br/>
         {% endif %}
+
+        {% if object.isLegislativeCandidate %}
```
mykiwi: You could remove all prefix
`object.legislativeCandidate` == `object.isLegislativeCandidate`

app/Resources/views/legislatives/platform.html.twig
```diff
@@ -0,0 +1,67 @@
...
+
+-->
+<html>
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="msapplication-tap-highlight" content="no" />
+    <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width" />
+
+    <link rel="apple-touch-icon" sizes="180x180" href="{{ asset('apple-touch-icon.png') }}?v2" />
```
mykiwi: instead of hardcoded `?v2` you could use the [assets version configuration](http://symfony.com/doc/current/reference/configuration/framework.html#version).

PR #962, opened 2017-04-19
Handled event invitations

HeahDude:


src/AppBundle/Controller/EventController.php
```diff
@@ -280,4 +283,47 @@ public function contactMembersAction(Request $request, Event $event): Response
...
+
+        return $this->render('events/invitation.html.twig', [
+            'committee_event' => $event,
+            'invitation_form' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/invitation/merci", name="app_event_invitation_sent")
+     * @Method("GET|POST")
```
rpg600: GET ?

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+
+    public function getEvent(): Event
+    {
+        return $this->event;
+    }
+
+    public function getCreatedAt(): \DateTimeImmutable
+    {
+        if ($this->createdAt instanceof \DateTime) {
+            $this->createdAt =  \DateTimeImmutable::createFromMutable($this->createdAt);
```
rpg600: double space

src/AppBundle/Controller/EventController.php
HeahDude: Of course! Fixed, thanks.

src/AppBundle/Entity/EventInvite.php
HeahDude: Yup, the ci was complaining too, thanks.

app/Resources/views/events/invitation.html.twig
```diff
@@ -0,0 +1,30 @@
...
+            {% if not is_granted('ROLE_ADHERENT') %}
+                <p>
+                    Avez-vous déjà un compte En Marche ?
+                    <a href="{{ path('app_adherent_login') }}">Connectez-vous</a>.
+                </p>
+            {% endif %}
+
+            {{ form(invitation_form, {attr: {novalidate: 'novalidate'}}) }}
+        </div>
+
```
rpg600: CS

app/Resources/views/events/invitation.html.twig
```diff
@@ -0,0 +1,30 @@
+{% extends 'base.html.twig' %}
+
+{% block page_title 'Inviter des amis à un événement' %}
+
+{% block content %}
+
```
rpg600: CS

app/Resources/views/events/invitation_sent.html.twig
```diff
@@ -0,0 +1,24 @@
+{% extends 'base.html.twig' %}
+
+{% block page_title 'Inviter des amis à un événement' %}
+
+{% block content %}
+
```
rpg600: CS

app/Resources/views/events/invitation_sent.html.twig
```diff
@@ -0,0 +1,24 @@
...
+        {{ include('events/_event_header.html.twig') }}
+        <hr/>
+    </section>
+    <section class="l__wrapper-500 space--30-0 b__nudge--bottom-huge">
+        <div class="event_invitation-result">
+            <h2 class="b__nudge--bottom text--normal">Inviter des amis à l'événement</h2>
+
+            <p>{{ 'event.invitation.form.invite_sent'|transchoice(invitations_count) }}</p>
+        </div>
+
```
rpg600: CS

src/AppBundle/Event/EventInvitationHandler.php
```diff
@@ -0,0 +1,40 @@
...
+            $url = $this->urlGenerator->generate('app_committee_show_event', [
+                'slug' => $event->getSlug(),
+                'uuid' => $event->getUuid()->toString(),
+            ]);
+
+            $this->manager->persist($invite);
+            $this->mailjet->sendMessage(EventInvitationMessage::createFromInvite($invite, $event->getName(), $url));
+        }
+
+        $this->manager->flush();
```
hhamon: It's a bit dangerous to flush at the very end because if something fails in between, nothing will be saved in the database. At least, I would flush after each successful sent invite.

app/Resources/translations/messages.fr.yml
```diff
@@ -115,3 +115,14 @@ ton_macron.invitation_form.self_last_name: Votre nom
...
+#
+# Event
+#
+event.invitation.form.email: Votre e-mail
+event.invitation.form.first_name: Votre prénom
+event.invitation.form.last_name: Votre nom
+event.invitation.form.message: Votre message (200 caractères max)
+event.invitation.form.guest: E-mail d'un ami
+event.invitation.form.invite: Envoyer
+event.invitation.form.invite_sent: Merci ! Votre invitation a bien été envoyée ! | Merci ! Vos %count% invitations ont bien été envoyées !
```
rpg600: These is a `forms.fr.yml` for form specific messages.

src/AppBundle/Event/EventInvitationHandler.php
HeahDude: I'm not sure it is worth having many transactions here, what's so dangerous about those calls?

app/Resources/translations/messages.fr.yml
HeahDude: Yeah, but it's not really used as it needs to enforce translation domain option everywhere.

src/AppBundle/Event/EventInvitationHandler.php
hhamon: Let's say you have 3 invites and the following scenario:

The first invite is correctly persisted in memory and its email is sent.
The second invite is correctly persisted in memory but its email is not send and the Mailjet service throws an exception.

Then nothing will be persisted in the database in the end.

app/Resources/views/events/invitation_sent.html.twig
HeahDude: The style is copy pasted from other templates, so unless we should fix all of them, these are not related to this PR.

src/AppBundle/Event/EventInvitationHandler.php
HeahDude: Emails are not sent at this point, they're just queued.

HeahDude: Even if it breaks, there is nothing that handle the error in the controller in the first place so it ends with a 500, I'm not sure everything should be persisted in such cases.

hhamon: The mailjet service still saves data in the database. This is why I think it's safer to flush after each successful sent invite. Maybe you should have a try/catch statement to catch any error when processing an invite. If one fails, you stack it and let the program process the next one. At the end you'll know if everything was successful or if some invite were abandonned.

app/Resources/views/events/invitation_sent.html.twig
mykiwi: I prefer this:
> This opportunistic refactoring is referred to by Uncle Bob as following the boy-scout rule - always leave the code behind in a better state than you found it.
https://martinfowler.com/bliki/OpportunisticRefactoring.html

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+class EventInvite
+{
+    use EntityIdentityTrait;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(length=50)
+     */
+    private $firstName = '';
```
mykiwi: What about use `EntityPersonNameTrait` ?

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+    public function __construct(Event $event)
+    {
+        $this->uuid = Uuid::uuid4();
+        $this->event = $event;
+        $this->createdAt = new \DateTimeImmutable();
+    }
+
+    public function __toString()
+    {
+        return 'Invitation à l\'évenement '.$this->event.' de '.$this->getSenderFullName().' à '.$this->email;
```
mykiwi: sprintf ?

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+
+        return $invite;
+    }
+
+    public function getSenderFullName(): string
+    {
+        return $this->firstName.' '.$this->lastName;
+    }
+
+    public function getFirstName(): string
```
mykiwi: `?string`

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        return $this->firstName;
+    }
+
+    public function getLastName(): string
```
mykiwi: `?string`

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        return $this->lastName;
+    }
+
+    public function getEmail(): string
```
mykiwi: `?string`

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,136 @@
...
+     *
+     * @ORM\Column(type="datetime")
+     */
+    private $createdAt;
+
+    public function __construct(Event $event)
+    {
+        $this->uuid = Uuid::uuid4();
+        $this->event = $event;
+        $this->createdAt = new \DateTimeImmutable();
```
mykiwi: Good idea 👍
Maybe we could do the same in `EntityTimestampableTrait`

src/AppBundle/Event/EventInvitationHandler.php
```diff
@@ -0,0 +1,40 @@
...
+        $this->manager = $manager;
+        $this->mailjet = $mailjet;
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function handle(EventInvitation $invitation, ?string $ip, Event $event)
+    {
+        foreach ($invitation->guests as $guest) {
+            $invite = EventInvite::create($invitation->firstName, $invitation->lastName, $guest, $ip, $event);
+            $url = $this->urlGenerator->generate('app_committee_show_event', [
```
mykiwi: Absolute URL ?

app/Resources/views/events/invitation_sent.html.twig
mykiwi: I prefer this:
> This opportunistic refactoring is referred to by Uncle Bob as following the boy-scout rule - always leave the code behind in a better state than you found it.
https://martinfowler.com/bliki/OpportunisticRefactoring.html

src/AppBundle/Event/EventInvitationHandler.php
HeahDude: I don't get it, do we wrap other email sending in `try` blocks? It does not look so. Maybe this safe guard should be global then?

>At the end you'll know if everything was successful or if some invite were abandonned.

If something went wrong we should roll back everything IMHO, I don't see what benefice it brings, otherwise what to do with it?

app/Resources/views/events/invitation_sent.html.twig
HeahDude: Fair enough, `event` folder checked in https://github.com/EnMarche/en-marche.fr/pull/962/commits/755c6533bc2e6c7d18f47e660ab71884233717ce.

src/AppBundle/Event/EventInvitationHandler.php
```diff
@@ -0,0 +1,40 @@
...
+        $this->manager = $manager;
+        $this->mailjet = $mailjet;
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function handle(EventInvitation $invitation, ?string $ip, Event $event)
+    {
+        foreach ($invitation->guests as $guest) {
+            $invite = EventInvite::create($invitation->firstName, $invitation->lastName, $guest, $ip, $event);
+            $url = str_replace('localhost', 'en-marche.fr', $this->urlGenerator->generate('app_committee_show_event', [
```
mykiwi: Why str_replace? If you want the real domain, you should [do like this](http://symfony.com/doc/current/console/request_context.html#configuring-the-request-context-per-command).

tgalopin: I did this too, we should do this.

HeahDude: We should not change the context as other urls should not be changed, this one is meant to be sent by email, not printed in any page.

app/migrations/Version20170419153243.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20170419153243 extends AbstractMigration
```
hhamon: The version number will have to be incremented to the current date to avoid problems with more recent migration version numbers.

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,113 @@
...
+
+    public function __toString()
+    {
+        return sprintf('Invitation à l\'évenement %s de %s à %s', $this->event, $this->getSenderFullName(), $this->email);
+    }
+
+    public static function create(string $firstName, string $lastName, string $email, string $clientIp, Event $event): self
+    {
+        $invite = new static($event);
+
```
hhamon: Blank line can be removed.

src/AppBundle/Entity/EventInvite.php
```diff
@@ -0,0 +1,113 @@
...
+        $invite->firstName = EmojisRemover::remove($firstName);
```
hhamon: Emojis removal can now be done automatically by the form thanks to the global `filter_emojis` option you can set to `true` on any text based form field type.

src/AppBundle/Event/EventInvitationHandler.php
hhamon: IMO this kind of change should be done inside our own implementation of URL generator.

src/AppBundle/Entity/EventInvite.php
HeahDude: I always put blank lines between assignments and calls, better readability.

app/migrations/Version20170419153243.php
HeahDude: Done, but note that this should not cause any problem since versions are sorted in a db table.

src/AppBundle/Entity/EventInvite.php
HeahDude: Noted, done.

src/AppBundle/Event/EventInvitationHandler.php
HeahDude: Done.

src/AppBundle/Routing/RemoteUrlGenerator.php
```diff
@@ -0,0 +1,49 @@
...
+            $this->generator->generate($name, $parameters, self::ABSOLUTE_URL)
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setContext(RequestContext $context)
+    {
+        $this->setContext($context);
```
hhamon: Infinite recursive call. I guess it's

```php
$this->generator->setContext($context);
```

HeahDude: Of course it is.

PR #963, opened 2017-04-19
Added entities for legislatives website.

hhamon:


src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->addKeyword($name);
+    }
+
+    public function addKeyword(string $keyword): void
+    {
+        if (null === $this->keywords) {
+            $this->keywords = '';
+        }
+
+        $this->keywords.= "\n".$keyword;
```
rpg600: missing space

src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -0,0 +1,111 @@
...
+        if ($keyword !== $lowercase = mb_strtolower($keyword, 'UTF-8')) {
+            $this->keywords.= "\n".$lowercase;
```
rpg600: missing space

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
      * @ORM\GeneratedValue
      */
     private $id;
 
     /**
-     * @var string
-     *
+     * @ORM\Column(length=30)
+     */
+    private $firstName;
```
mykiwi: What about use `EntityPersonNameTrait` ?

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
+
+    /**
+     * @ORM\Column(length=30)
+     */
+    private $lastName;
+
+    /**
+     * @ORM\Column(length=100, nullable=true)
+     */
+    private $profilePicture;
```
mykiwi: What about use `EntityMediaTrait` ?

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
 class LegislativeCandidate
```
mykiwi: We may need to implement `GeocodableInterface, GeoPointInterface` later

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
      */
-    private $candidate;
+    private $description;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\LegislativeDistrictZone")
+     */
+    private $districtZone;
 
     public function getId(): int
```
mykiwi: `?int` since the constructor does not initialize the value

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
     {
         return $this->id;
     }
 
+    public function getFirstName(): string
```
mykiwi: `?string`

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
+    {
+        return $this->firstName;
+    }
+
+    public function setFirstName(string $firstName): void
+    {
+        $this->firstName = $firstName;
+    }
+
+    public function getLastName(): string
```
mykiwi: `?string`

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
+    {
+        return $this->districtName;
+    }
+
+    public function setDistrictName(string $districtName): void
+    {
+        $this->districtName = $districtName;
+    }
+
+    public function getDistrictNumber(): string
```
mykiwi: `?string`

src/AppBundle/Entity/LegislativeCandidate.php
```diff
@@ -14,34 +14,214 @@
...
+    {
+        return $this->websiteUrl;
+    }
+
+    public function setWebsiteUrl(?string $websiteUrl): void
+    {
+        $this->websiteUrl = $websiteUrl;
+    }
+
+    public function getDistrictName(): string
```
mykiwi: `?string`

src/AppBundle/Entity/LegislativeCandidate.php
hhamon: Agree.

src/AppBundle/Entity/LegislativeCandidate.php
hhamon: Yes good point.

src/AppBundle/Entity/LegislativeCandidate.php
hhamon: Well no in fact. Because we don't have full addresses. Admin will manually fill the coordinates from the backend area.

src/AppBundle/Entity/LegislativeCandidate.php
hhamon: Done. thanks.

src/AppBundle/Entity/LegislativeCandidate.php
hhamon: Good catch.

mykiwi:
You may need to adapt https://github.com/EnMarche/en-marche.fr/blob/master/src/AppBundle/Admin/AdherentAdmin.php#L81-L88 & https://github.com/EnMarche/en-marche.fr/blob/master/src/AppBundle/Admin/AdherentAdmin.php#L180-L186

hhamon:
@mykiwi fixed. thanks.

mykiwi:
One last thing, you should adapt the doctrine migration to select all adherent who are a legislative candidate, and set the boolean to true.

rpg600:
You should add the firstName and lastName attribute since the relationship with the adherent entity was dropped. And maybe generate the slug based on these two fields.

hhamon:
@rpg600 first name and last name are already there thanks to the `EntityPersonNameTrait` trait. For now slugs are manually generated. I don't think we need to automate slug generation as all candidates' profiles will be manually inserted from the BO.

tgalopin:
Automated them on insertion to avoid having to enter them by hand could be useful though

hhamon:
@rpg600 @tgalopin @mykiwi I applied your requested changes. I think PR is good to merge now.

src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -0,0 +1,115 @@
...
+    public function __construct(string $type, string $name, array $keywords = [])
+    {
+        $this->setAreaType($type);
+        $this->setKeywords($keywords);
+        $this->setName($name);
+    }
+
+    public function __toString(): string
+    {
+        return $this->name;
```
HeahDude: This is broken, the "name" property is not initialized as an empty string, returning null with a string type hint how a runtime exception.

src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public function getId(): int
```
HeahDude: `?int`

src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return $this->id;
+    }
+
+    public function setAreaType(string $type): void
+    {
+        if (!in_array($type = strtolower($type), $types = [self::TYPE_DEPARTMENT, self::TYPE_REGION])) {
```
HeahDude: Why making the constants private if we need to pass them as argument?

hhamon: Because the constants are not used outside of the class at the moment. I will make them public later only if I have a real need for accessing them outside of the class.

PR #979, opened 2017-04-21
add command for load circonscriptions from csv

mykiwi:


src/AppBundle/Command/LoadLegislativeDistrictZonesCommand.php
```diff
@@ -0,0 +1,91 @@
...
+{
+    /**
+     * @var \Doctrine\ORM\EntityManager
+     */
+    private $manager;
+
+    protected function configure()
+    {
+        $this
+            ->setName('app:load:legislative-district-zones')
```
hhamon: Shall we name it `app:legislatives:load-district-zones` instead? It sounds more logical to me.

src/AppBundle/DataFixtures/ORM/LoadLegislativesData.php
```diff
@@ -1,384 +0,0 @@
-<?php
-
-namespace AppBundle\DataFixtures\ORM;
```
hhamon: This file must not be removed as it's used for functional tests.

PR #1009, opened 2017-04-24
Fixes Sentry issue #226663790

rpg600:
https://sentry.io/en-marche-i7/en-marchefr/issues/226663790/

### DIFF
--- a/src/AppBundle/Entity/Committee.php
+++ b/src/AppBundle/Entity/Committee.php
@@ -352,7 +352,7 @@ class Committee implements GeoPointInterface
 
     public function isCreatedBy(UuidInterface $uuid): bool
     {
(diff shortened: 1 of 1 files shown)


src/AppBundle/Entity/Committee.php
```diff
@@ -352,6 +352,10 @@ public function getCreatedBy(): ?string
 
     public function isCreatedBy(UuidInterface $uuid): bool
     {
+        if (!$this->createdBy) {
+            return false;
+        }
+
         return $this->createdBy->equals($uuid);
```
mykiwi: inline ? `return $this->createdBy && ...`

rpg600: Yes!

PR #1011, opened 2017-04-24
Fixes Sentry issue #258906440.

hhamon:
This PR is not really a fix for Sentry issue [258906440](
https://sentry.io/en-marche-i7/en-marchefr/issues/258906440/). This issue is a both a bug and an invalid value. The user tried to store the integer `605,525,601` in a `SIGNED SMALLINT` field. To prevent this situation from happening again in the future, the PR provides the following fixes:

* Make the `not_convinced` SQL field `unsigned` to increase its value from up to `65535` instead of `32767`,
* Add server-side data validation to prevent the user submitting a value greater than `65000` for this field.

### DIFF
--- a/src/AppBundle/Entity/JeMarcheReport.php
+++ b/src/AppBundle/Entity/JeMarcheReport.php
@@ -94,9 +94,10 @@ class JeMarcheReport
     /**
      * @var int|null
      *
-     * @ORM\Column(type="smallint", nullable=true)
+     * @ORM\Column(type="smallint", nullable=true, options={"unsigned": true})
      *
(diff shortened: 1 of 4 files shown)


src/AppBundle/Entity/JeMarcheReport.php
```diff
@@ -15,7 +15,7 @@
  *
  * @Algolia\Index(autoIndex=false)
  */
-class JeMarcheReport
+final class JeMarcheReport
```
tgalopin: Why?

hhamon: No valid reasons, I'm reverting it.

PR #1015, opened 2017-04-25
Add special message for procuration bc of residency

marineb:


front/listeners/procuration-reason.js
```diff
@@ -0,0 +1,15 @@
+/*
+ * Display special message if the reason for procuration is 'residency'
+ */
+export default () => {
+    const dropdown = dom('#app_procuration_elections_reason');
```
tgalopin: You shouldn't put this in a listener as it would be loaded on every page. Instead, you could put this code in the procurations pages files `front/pages/procuration*.js`

tgalopin:
Closed by https://github.com/EnMarche/en-marche.fr/pull/1020 (https://github.com/EnMarche/en-marche.fr/pull/1020/commits/e71d397d453fa88b400dc19c5e74e004f3ab3a0b)

PR #1016, opened 2017-04-25
Fixes Sentry issue #259291665.

hhamon:
https://sentry.io/en-marche-i7/stagingen-marchefr/issues/259291665/

### DIFF
--- a/src/AppBundle/Entity/LegislativeDistrictZone.php
+++ b/src/AppBundle/Entity/LegislativeDistrictZone.php
@@ -74,24 +74,23 @@ class LegislativeDistrictZone
      */
     private $keywords;
 
-    public static function createDepartmentZone(string $areaCode, string $name, array $keywords = [], int $rank = 1): self
+    public static function createDepartmentZone(string $areaCode, string $name, array $keywords = []): self
     {
-        return self::create($areaCode, self::TYPE_DEPARTMENT, $name, $keywords, $rank);
+        return self::create($areaCode, self::TYPE_DEPARTMENT, $name, $keywords);
     }
 
-    public static function createRegionZone(string $areaCode, string $name, array $keywords = [], int $rank = 1): self
+    public static function createRegionZone(string $areaCode, string $name, array $keywords = []): self
     {
-        return self::create($areaCode, self::TYPE_REGION, $name, $keywords, $rank);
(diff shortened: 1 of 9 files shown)


src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -139,11 +144,40 @@ public function getAreaCode(): ?string
...
     {
-        $this->areaCode = $code;
+        $this->areaCode = sprintf('%04s', $code);
+
+        if (null === $this->rank) {
+            $this->rank = self::guessRankFromAreaCode($code);
+        }
+    }
+
+    private static function guessRankFromAreaCode(string $areaCode): int
```
mykiwi: No need to be static I guess.
Maybe you should explain why you add rank+1 after corse area in a phpdoc.

src/AppBundle/Entity/LegislativeDistrictZone.php
```diff
@@ -139,11 +144,40 @@ public function getAreaCode(): ?string
...
+    {
+        if (self::isCorsica($areaCode)) {
+            return 'A' === substr($areaCode, -1) ? 20 : 21;
+        }
+
+        $rank = (int) ltrim($areaCode, '0');
```
mykiwi: What about `intval` ?
Or directly play with `$areaCode`, it should works.

hhamon: I always use `(int)` instead of `intval()` function. It's an explicit casting. But I don't know if one is better than the other.

src/AppBundle/Entity/LegislativeDistrictZone.php
hhamon: Okay for removing static and adding phpdoc.

src/AppBundle/Entity/LegislativeDistrictZone.php
tgalopin: A cast is better than a function call.

mykiwi: Ok. But it will also works without the cast (but I don't know if it's better).

hhamon: According to J.Pauli, explicit casting like I do is the best.

PR #1022, opened 2017-04-26
set user's city by default

mykiwi:


tgalopin:
Did you check both the first results and the following one use the right city? I would like a check by @hhamon as he worked on this recently.

mykiwi:
Yes.

hhamon:
That's a tricky fix and it might not work for all situations. When the same city name exist in multiple countries (or event in the same country), which one will it take for geocoding? I guess Google will take the most popular of all or something like this. I think we need to add more information like the country name at least to lower the risk of strange results: `Paris, France`.

tgalopin:
I agree, we should put the country name

PR #1031, opened 2017-04-27
fix search for admins

mykiwi:


src/AppBundle/Controller/SearchController.php
```diff
@@ -21,7 +22,8 @@ public function searchEventsAction(Request $request)
         $request->query->set(SearchParametersFilter::PARAMETER_TYPE, SearchParametersFilter::TYPE_EVENTS);
 
         $search = $this->getSearch($request);
-        if ($user = $this->getUser()) {
+        $user = $this->getUser();
+        if ($user && in_array(EntityPostAddressTrait::class, class_uses($user))) {
```
HeahDude: What about an interface for methods implemented by the trait?

PR #1036, opened 2017-04-27
fix search & events with algolia

mykiwi:
https://sentry.io/en-marche-i7/en-marchefr/issues/260843273/


src/AppBundle/Entity/EntityCrudTrait.php
```diff
@@ -59,6 +59,11 @@ public function __set($property, $value)
     {
         $class = get_class($this);
         $ucProperty = ucfirst($property);
+
+        if (false !== strpos($property, '.')) { // Algolia fix
```
hhamon: Can you please explain more why this fix?

PR #1045, opened 2017-04-28
Automatically open bot on load

marineb:


app/Resources/views/base.html.twig
```diff
@@ -459,6 +459,7 @@
         <script type="text/javascript">
             on(window, 'load', function() {
                 Kernel.boot(
+                    motionAI_Open(),
```
tgalopin: If this is working, that's not normal ^^

marineb:
@tgalopin does it look better now?

tgalopin:
Better, but actually you should put this in https://github.com/EnMarche/en-marche.fr/blob/master/app/Resources/views/bot/_javascript.html.twig (to avoid JS issues on pages not having the bot).

marineb:
@tgalopin sounds good! updated.

PR #1069, opened 2017-05-02
Fixed emoji removal

HeahDude:


src/AppBundle/Admin/PageAdmin.php
```diff
@@ -22,16 +22,19 @@ protected function configureFormFields(FormMapper $formMapper)
         if ($this->getSubject()->getId() === null) {
             $formMapper->add('title', null, [
```
tgalopin: TextType?

src/AppBundle/Entity/SocialShareCategory.php
```diff
@@ -45,28 +44,28 @@ class SocialShareCategory
...
+        return $this->name;
     }
 
     public function getId(): int
     {
         return $this->id;
     }
 
-    public function getName(): ?string
+    public function getName(): string
```
tgalopin: I think this won't always work (when the form will put null data in the name property).

src/AppBundle/Admin/PageAdmin.php
HeahDude: Good catch!

src/AppBundle/Entity/SocialShareCategory.php
HeahDude: Yes that's why there is the cast below ;).

PR #1073, opened 2017-05-03
Fixed event invitation message empty field value

HeahDude:


tgalopin:
Is it possible to test our custom TextType?

HeahDude:
It is, but it does not bring actual value in such cases IMO, we only add two listeners and some options, so it would be like testing Symfony itself, whereas the current test suit fails if we remove the extension.

PR #1076, opened 2017-05-03
Add user informations to event invitation if authentificated

tony-tran:


app/Resources/views/events/invitation.html.twig
```diff
@@ -24,19 +24,19 @@
                 <div class="form__row">
                     {{ form_label(invitation_form.email, 'Votre adresse e-mail') }}
                     {{ form_errors(invitation_form.email) }}
-                    {{ form_widget(invitation_form.email, { 'attr': { 'class': 'form--full' } }) }}
+                    {{ form_widget(invitation_form.email, { 'value': app.user.emailAddress is defined ? app.user.emailAddress, 'attr': { 'class': 'form--full' } }) }}
```
HeahDude: Maybe you should create the object in the controller instead, using a static method like `createFromAdherent($this->getUser())` to pre set the data, doing it in the view prevent the user to change it as it would be overridden in case the template is reloaded with errors.

src/AppBundle/Event/EventInvitation.php
```diff
@@ -41,8 +42,69 @@ class EventInvitation
      */
     public $guests = [];
 
+    public static function createFromAdherent($adherent = null): self
+    {
+        $dto = new self();
+
+        if ($adherent instanceof Adherent) {
```
HeahDude: Please remove this check and type hint the parameter, also considering the name of the method it should not be nullable.

HeahDude: Ok got why the if is needed, still the parameter should be type hinted with a nullable class:
```php
public static function createFromAdherent(?Adherent $adherent)
{
    // ...
}
```

src/AppBundle/Event/EventInvitation.php
```diff
@@ -41,8 +42,69 @@ class EventInvitation
      */
     public $guests = [];
 
+    public static function createFromAdherent(?Adherent $adherent): self
+    {
+        $dto = new self();
+
+        if ($adherent instanceof Adherent) {
```
HeahDude: Thanks to the type hint, `if ($adherent) {` is enough.

HeahDude:
Thank you @tony-tran.

src/AppBundle/Event/EventInvitation.php
```diff
@@ -41,8 +42,69 @@ class EventInvitation
      */
     public $guests = [];
 
+    public static function createFromAdherent(?Adherent $adherent): self
+    {
+        $dto = new self();
+
+        if ($adherent) {
+            $dto->setLastName($adherent->getLastName());
```
HeahDude: I don't get why you added getters and setters, I've removed them in https://github.com/EnMarche/en-marche.fr/commit/4b8f8712c66d85c240d86d6c195d1f8038e850dc.

PR #1088, opened 2017-05-06
Enforced uuid validation

HeahDude:


web/app.php
```diff
@@ -21,7 +21,6 @@
 
 /** @var \Composer\Autoload\ClassLoader $loader */
 $loader = require __DIR__.'/../app/autoload.php';
-include_once __DIR__.'/../var/bootstrap.php.cache';
```
tgalopin: Why?

composer.json
```diff
@@ -73,11 +73,9 @@
     "scripts": {
         "symfony-scripts": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
```
HeahDude: https://github.com/symfony/symfony-standard/blob/master/web/app.php#L8, opcache does the job, it adds overhead for no gain.

web/app.php
HeahDude: Same as above.

composer.json
tgalopin: Nice! However, I would prefer to have a different PR, it's not related to the current one.

HeahDude: ok, not a problem.

HeahDude: Done.

src/AppBundle/Controller/DonationController.php
```diff
@@ -91,11 +92,15 @@ public function callbackAction(Request $request)
     {
         $id = explode('_', $request->query->get('id'))[0];
```
hhamon: This line and the following condition should be encapsulated somewhere in a static method. Something similar to this.

```
try {
    Donation::validateId($request->query->get('id'));
} catch (InvalidDonationIdException $e) {
    return $this->redirectToRoute('donation_index');
}

File: tests/AppBundle/Controller/AssetsControllerTest.php [hunk 4]
```diff
@@ -91,7 +91,7 @@ public function testStaticMaps()
 
     public function testStaticMapsWithWrongQuery()
     {
-        $client = static::createClient();
+        $client = $this->client;
```
hhamon: Why making a copy?

tests/AppBundle/Controller/AssetsControllerTest.php
```diff
@@ -45,7 +45,7 @@ public function testInvalidAssetWithSignatureIsNotFound()
 
     public function testStaticMaps()
     {
-        $client = static::createClient();
+        $client = $this->client;
```
hhamon: Why making a copy?

hhamon:
Can't we manage the UuidException in a custom `kernel.exception` listener? We could then simplify the controllers and avoid all these try/catch blocks.

src/AppBundle/Controller/DonationController.php
HeahDude: The donation controller needs more love but it's for another PR ;).

tests/AppBundle/Controller/AssetsControllerTest.php
HeahDude: IT was a quick fix. Addressed.

tests/AppBundle/Controller/AssetsControllerTest.php
HeahDude: Fixed.

HeahDude:
>Can't we manage the UuidException in a custom kernel.exception listener? We could then simplify the controllers and avoid all these try/catch blocks.

I'm not sure about the exception listener, there is already one in charge for the response because we throw an HttpExceptionInterface, this could not be done in a listener.

Repository have to throw an InvalidArgumentException, letting the controller transform it to an HTTP one make it explicit also when we read the code, and allow to adjust handling things per actions.

We already have other try catch/blocks in controllers so it looks consistent to me, especially the Facebook one.

@tgalopin what do you think?

tests/AppBundle/Controller/AdherentControllerTest.php
```diff
@@ -18,7 +18,7 @@
 use Tests\AppBundle\MysqlWebTestCase;
 
 /**
- * @group functional
+ * @group functionnal
```
tony-tran: I create a PR to fix typo. Can you please change to 

> functional

PR #1089, opened 2017-05-06
Removed needless composer scripts

HeahDude:


web/app.php
```diff
@@ -21,7 +21,6 @@
 
 /** @var \Composer\Autoload\ClassLoader $loader */
 $loader = require __DIR__.'/../app/autoload.php';
-include_once __DIR__.'/../var/bootstrap.php.cache';
```
tgalopin: Shouldn't you remove `$kernel->loadClassCache();` as well?

HeahDude: Indeed, fixed.

PR #1099, opened 2017-05-09
add firebase cloud messaging for web notifications

mykiwi:


src/AppBundle/Controller/Api/ArticlesController.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * @Route("/articles/{date}T{time}.{ms}Z{timezone}", requirements={
+     *     "date"="\d{4}\-\d{2}\-\d{2}",
+     *     "time"="\d{2}:\d{2}:\d{2}",
+     *     "ms"="\d+",
+     *     "timezone"="(\+|-)\d{3}"
+     * }, name="api_articles_since")
+     * @Method("GET")
+     */
+    public function indexAction(string $date, string $time, string $timezone): JsonResponse
```
rpg600: Maybe you can add a simple functional test for this action.

app/Resources/views/firebase/firebase-messaging-ws.js.twig
```diff
@@ -0,0 +1,18 @@
+importScripts('https://www.gstatic.com/firebasejs/3.9.0/firebase-app.js');
+importScripts('https://www.gstatic.com/firebasejs/3.9.0/firebase-messaging.js');
+
+firebase.initializeApp({
+    messagingSenderId: "{{ firebase_messaging_sender_id }}"
+});
+
+const messaging = firebase.messaging();
```
tgalopin: You have to compile this to use ES6 features.

front/listeners/firebase-cloud-messaging-notification.js
```diff
@@ -0,0 +1,40 @@
...
+                api.addFirebaseToken(refreshedToken);
+            })
+            .catch((err) => {
+                // console.log('Unable to retrieve refreshed token ', err);
+            });
+    });
+
+    // if the website is on focus, do it inside the page ?
+    messaging.onMessage((payload) => {
+        // console.log(payload);
```
tgalopin: I guess this PR is not fully ready then? Or do you want to implement this in a following PR?

app/Resources/views/firebase/firebase-messaging-ws.js.twig
mykiwi: I don't get it, what should I change?

front/listeners/firebase-cloud-messaging-notification.js
mykiwi: Notifications are done, but we can also add an interaction inside the site if it is open. This part is not done since I don't know which UI we want.

app/Resources/views/firebase/firebase-messaging-ws.js.twig
tgalopin: You are including this file directly in your Twig templates aren't you? You should use Webpack instead, for minification and as you use ES6 features.

src/AppBundle/Notification/SendOnArticlePublished.php
```diff
@@ -0,0 +1,94 @@
...
+use sngrl\PhpFirebaseCloudMessaging\ClientInterface;
+use sngrl\PhpFirebaseCloudMessaging\Message;
+use sngrl\PhpFirebaseCloudMessaging\Notification;
+use sngrl\PhpFirebaseCloudMessaging\Recipient\Topic;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+class SendOnArticlePublished implements EventSubscriber
+{
+    const TOPIC = 'news';
```
javiereguiluz: You may already know this, but since the project requires PHP 7.1, you can use `private const TOPIC = 'news';` to make the constant private in case it's only used in this class.

mykiwi: Yes, but I use it in [FirebaseController.php](https://github.com/EnMarche/en-marche.fr/pull/1099/files#diff-2586e748739d54a1be4395bd14ba04b8R34).

PR #1105, opened 2017-05-10
Don't depend on container

mleko:


tgalopin:
There is a circular reference as you are loading redis, which is loading the logger, which is booting the app, which is loading redis. That's the reason why I used the container directly. If you find something to fix this, I'll be ok with merging this PR.

HeahDude:
Let's wait for the upcoming Symfony 3.3 service locators then.

mleko:
Simple solutions are the best ones, use lazy service

HeahDude:
Hello @mleko! Thank you for this contribution, with the recent release of Symfony 3.3, I propose to fix this in #1179 so I'm closing here.

I hope we will see other PRs from you, cheers!

PR #1118, opened 2017-05-16
fix donate

mykiwi:
https://sentry.io/en-marche-i7/en-marchefr/issues/268887245/

### DIFF
--- a/app/Resources/translations/validators.fr.yml
+++ b/app/Resources/translations/validators.fr.yml
@@ -81,7 +81,7 @@ committee.message.min_length: Le message doit contenir au moins {{ limit }} cara
 # Donation
 #
 donation.amount.not_blank: Cette valeur ne doit pas être vide.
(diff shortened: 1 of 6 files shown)


app/Resources/translations/validators.fr.yml
```diff
@@ -81,7 +81,7 @@ committee.message.min_length: Le message doit contenir au moins {{ limit }} cara
 # Donation
 #
 donation.amount.not_blank: Cette valeur ne doit pas être vide.
-donation.amount.greater_than_0: La valeur minimale d'un don est de 0 €.
+donation.amount.greater_than_0: La valeur minimale d'un don est de 0.01 €.
```
hhamon: Is there any real minimum like 5 or 10€?

tgalopin: No there isn't.

PR #1120, opened 2017-05-17
add role legislative_candidate for users with a @en-marche.fr email

mykiwi:


src/AppBundle/Entity/Adherent.php
```diff
@@ -194,7 +194,7 @@ public function getRoles(): array
             $roles[] = 'ROLE_PROCURATION_MANAGER';
         }
 
-        if ($this->legislativeCandidate) {
+        if ($this->legislativeCandidate || false !== strpos($this->emailAddress, '@en-marche.fr')) {
```
hhamon: `stripos()` maybe instead?

PR #1126, opened 2017-05-18
Simplify adherent files access

hhamon:


src/AppBundle/Documents/DocumentManager.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * @return Document[]
+     */
+    public function listDirectory(string $type, string $path): array
+    {
+        return $this->repository->listDirectory($type, $path);
+    }
+
+    /**
+     * @return array
```
mykiwi: Not really useful

PR #1142, opened 2017-05-22
Mandatory procuration vote office

tony-tran:


tony-tran:
> required = true / false is just for HTML5 validation, you should update entity constraints too.

Done.

src/AppBundle/Entity/ProcurationProxy.php
```diff
@@ -612,7 +613,7 @@ public function getVoteOffice(): ?string
         return $this->voteOffice;
     }
 
-    public function setVoteOffice(?string $voteOffice)
+    public function setVoteOffice(string $voteOffice): void
```
tgalopin: Are you sure this will work? The Form component may inject `null` if not configured and no string is provided. Can you check?

src/AppBundle/Entity/ProcurationRequest.php
```diff
@@ -619,7 +620,7 @@ public function getVoteOffice(): ?string
         return $this->voteOffice;
     }
 
-    public function setVoteOffice(?string $voteOffice): void
+    public function setVoteOffice(string $voteOffice): void
```
tgalopin: Same here.

src/AppBundle/Entity/ProcurationProxy.php
tony-tran: Sorry, i forgot to push force ^^

HeahDude: Let me remind you I introduced a form type extension for TextType and its children (Email, Url, Textarea, ...) that defines `empty_data` as string by default ;).

HeahDude:
Thank you @tony-tran.

PR #1224, opened 2017-06-05
Fixed retrying donation

HeahDude:


src/Donation/DonationRequest.php
```diff
@@ -303,11 +303,14 @@ public function retryPayload(array $payload): self
...
         if (isset($payload['ad'])) {
-            $retry->address = urldecode((string) $payload['ad']);
+            $retry->address = (string) $payload['ad'];
         }
 
         if (isset($payload['phc']) && isset($payload['phn'])) {
             $retry->phone = self::createPhoneNumber((string) $payload['phc'], (string) $payload['phn']);
+        } else {
+            // remove the invalid state
+            $retry->phone = null;
```
tgalopin: What about putting this over the if? I don't like elses.

src/Donation/DonationRequestUtils.php
```diff
@@ -115,15 +115,15 @@ public function createCallbackStatus(Donation $donation): array
 
     private function hydrateFromRetryPayload(DonationRequest $request, string $payload): DonationRequest
     {
-        $data = array_filter(json_decode($payload, true));
+        $data = array_filter(json_decode(urldecode($payload), true));
```
tgalopin: The payload is not checked to be valid JSON, this could easily fail. We should handle the case IMO.

src/Donation/DonationRequest.php
HeahDude: Done.

src/Donation/DonationRequestUtils.php
HeahDude: Nice catch, done.

PR #1230, opened 2017-06-06
Disconnect admin after 30 min of inactivity

illusionOfParadise:


src/Security/InactiveAdminListener.php
```diff
@@ -0,0 +1,53 @@
...
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        if (HttpKernelInterface::MASTER_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        if ($token = $this->securityToken->getToken()) {
+            $user = $token->getUser();
+
+            if (!$user instanceof Administrator) {
```
tgalopin: Can you also check if the user is impersonating an adherent?

src/Security/InactiveAdminListener.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    private $session;
+    private $securityToken;
+    private $maxIdleTime;
+    private $logoutUrlGenerator;
+
+    public function __construct(SessionInterface $session, TokenStorageInterface $securityToken, LogoutUrlGenerator $logoutUrlGenerator, int $maxIdleTime = 0)
+    {
+        $this->session = $session;
+        $this->securityToken = $securityToken;
```
HeahDude: Please name it `$tokenStorage`.

src/Security/InactiveAdminListener.php
```diff
@@ -0,0 +1,53 @@
...
+
+        if ($token = $this->securityToken->getToken()) {
+            $user = $token->getUser();
+
+            if (!$user instanceof Administrator) {
+                return;
+            }
+
+            if ($this->maxIdleTime > 0) {
+                $this->session->start();
```
HeahDude: This line is useless, if there is a stored token, there is also a session.

app/config/config_test.yml
```diff
@@ -21,6 +21,7 @@ services:
         public: false
 
 parameters:
+    env(SESSION_MAX_IDLE_TIME): 1800
```
HeahDude: This line seems useless since it's defined in `parameters.yml.dist`.

app/config/services/system.xml
```diff
@@ -65,6 +65,13 @@
             <tag name="kernel.event_listener" event="security.interactive_login" method="onSecurityInteractiveLogin" />
         </service>
 
+        <service id="AppBundle\Security\InactiveAdminListener">
+            <argument type="service" id="service_container"/>
```
HeahDude: Does it work with `container` only instead of `service_container`? I think you may inject the whole container whereas we need a service locator (a subset of it).

src/Security/InactiveAdminListener.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace AppBundle\Security;
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\Administrator;
+use Symfony\Component\DependencyInjection\ContainerInterface;
```
HeahDude: Please use the PSR-11.

PR #1255, opened 2017-06-12
Fixed Paybox error status

HeahDude:


src/Donation/DonationRequestUtils.php
```diff
@@ -103,7 +103,7 @@ public function createRetryPayload(Donation $donation, Request $request): array
 
     public function createCallbackStatus(Donation $donation): array
     {
-        $code = self::PAYBOX_STATUSES[$donation->getPayboxResultCode()] ?? self::PAYBOX_UNKNOWN;
+        $code = self::PAYBOX_STATUSES[$donation->getPayboxResultCode()] ?? 'error';
```
illusionOfParadise: What about `self::PAYBOX_STATUSES[self::PAYBOX_UNKNOWN]` or creating another constant ?

HeahDude: I thought about that, was not sure of the benefit of a new constant, I guess we should go for `self::PAYBOX_STATUSES[self::PAYBOX_UNKNOWN]` then.

tgalopin: `self::PAYBOX_STATUSES[self::PAYBOX_UNKNOWN]` seems nice indeed

HeahDude: Done.

PR #1259, opened 2017-06-12
Make checkboxes checkable on 'My profile' page and save values

illusionOfParadise:


app/Resources/views/adherent/_membership_form.html.twig
```diff
@@ -105,17 +105,23 @@
         </div>
     {% endif %}
 
+    {% if form.vars.data.firstName %}
```
HeahDude: `{% set prefix = form.vars.data.firstName ? 'update_' %}` is enough.

PR #1262, opened 2017-06-13
Handled publication of events

HeahDude:


src/Repository/CommitteeFeedItemRepository.php
```diff
@@ -33,9 +33,8 @@ private function findMostRecentFeedItem(string $type, string $committeeUuid = nu
     {
         $qb = $this
             ->createQueryBuilder('i')
-            ->select('i, a, e')
+            ->select('i, a')
             ->leftJoin('i.author', 'a')
-            ->leftJoin('i.event', 'e')
```
tgalopin: Why do you remove the left join? You can keep it and put a WITH. Here, it would try to load the event lazily.

HeahDude: It's not removed, it's moved below :).

PR #1294, opened 2017-06-19
Show no error or image if Google Map connection failure

illusionOfParadise:


src/Controller/AssetsController.php
```diff
@@ -160,4 +160,22 @@ private function getTypeRepository(string $type)
 
         return $manager->getRepository(Article::class);
     }
+
+    /**
+     * Create a transparent image PNG with sizes 1px x 1px.
```
HeahDude: Create**s**.

src/Controller/AssetsController.php
```diff
@@ -160,4 +160,22 @@ private function getTypeRepository(string $type)
...
+     *
+     * @return string Image path
+     */
+    private function createWhiteImage()
```
HeahDude: Please add a return type hint.

PR #1295, opened 2017-06-19
Added Summary entities

HeahDude:


src/Entity/Adherent.php
```diff
@@ -157,6 +158,13 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
      */
     private $memberships;
 
+    /**
+     * @var Summary|null
+     *
+     * @ORM\OneToOne(targetEntity="Summary", mappedBy="member", cascade={"all"})
```
tgalopin: Shouldn't we put this as lazy to avoid loading it on each request?

HeahDude: This is not the owner side, will test it later if it's worth it but it should not be needed.

PR #1304, opened 2017-06-21
Add mission type admin and fixtures

illusionOfParadise:


app/config/services/admin.xml
```diff
@@ -366,5 +366,19 @@
             <argument />
         </service>
 
+        <!--
+        ######################################################
+        # CV
+        ######################################################
+         -->
+
+        <service id="app.admin.summary_mission_type" class="AppBundle\Admin\MissionTypeAdmin">
```
HeahDude: Do we really need a specific id here? it would be best to use the class only when possible.

src/Entity/MemberSummary/MissionType.php
```diff
@@ -23,6 +31,8 @@ class MissionType
     /**
      * @var string
      *
+     * @ORM\Column(length=255, unique=true)
```
HeahDude: You can remove the length attribut since it's the default.

app/config/services/admin.xml
illusionOfParadise: Id can't be removed, it's used by Sonata.

PR #1310, opened 2017-06-21
Made Twig extensions lazy by using runtime

HeahDude:


src/Intl/UnitedNationsBundle.php
```diff
@@ -7,15 +7,17 @@
...
-     *
-     * @param string $locale
-     *
-     * @return array
      */
-    public static function getCountries($locale = 'fr'): array
+    public static function getCountries(?string $locale = 'fr'): array
     {
+        if (self::$countries) {
+            return self::$countries;
```
HeahDude: Is it really necessary? The locale should not change during the request.

tgalopin: This is not logical in the sense of this class. It does not cost much to use an array.

HeahDude: Ok done.

PR #1312, opened 2017-06-21
Improve administration panel globally

tgalopin:


app/Resources/views/admin/committee/list_actions.html.twig
```diff
@@ -29,11 +29,13 @@
         </a>
     </div>
 
+    {% if object.isApproved %}
     <div style="width: 160px; margin-top: 5px;">
```
HeahDude: You should indent these lines.

app/Resources/views/admin/event/list_status.html.twig
```diff
@@ -1,7 +1,9 @@
 {% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
 
 {% block field %}
-    {% if object.isCancelled %}
+    {% if not object.isPublished %}
```
HeahDude: No need for `is` prefix.

app/Resources/views/admin/event/list_status.html.twig
```diff
@@ -1,7 +1,9 @@
...
+        <span class="label label-warning">Dépublié</span>
+    {% elseif object.isCancelled %}
```
HeahDude: Same here.

src/Admin/AdministratorAdmin.php
```diff
@@ -73,6 +73,7 @@ protected function configureFormFields(FormMapper $formMapper)
                     'ROLE_ADMIN_TON_MACRON' => 'ROLE_ADMIN_TON_MACRON',
                     'ROLE_ADMIN_LEGISLATIVES' => 'ROLE_ADMIN_LEGISLATIVES',
                     'ROLE_ADMIN_ADHERENTS' => 'ROLE_ADMIN_ADHERENTS',
+                    'ROLE_ADMIN_SUMMARY' => 'ROLE_ADMIN_SUMMARY',
```
HeahDude: This array is ugly, to simplify this kind of stuff just remove the keys and add:
```php
`choice_label` => function ($choice) { return $choice; },
```

PR #1319, opened 2017-06-22
Added a SummaryManagerController

HeahDude:


app/config/services/community.xml
```diff
@@ -145,5 +145,15 @@
             <argument>%kernel.cache_dir%</argument>
         </service>
 
+        <!-- Summary -->
+        <service id="AppBundle\Summary\SummaryFactory">
+            <argument type="service" id="cocur_slugify" />
```
tgalopin: Where does cocur_slugify come from?

src/Summary/SummaryManager.php
```diff
@@ -0,0 +1,28 @@
...
+        $this->repository = $repository;
+    }
+
+    public function getForAdherent(Adherent $adherent): Summary
+    {
+        if ($summary = $this->repository->findOneForAdherent($adherent)) {
+            return $summary;
+        }
+
+        return $this->factory->createFromAdherent($adherent);
```
tgalopin: You will need to be careful to persist this properly as it will not be managed by Doctrine in this case but will be in the other case.

HeahDude: No this controller is not supposed to change the state. Forms will be handled in other controllers.

app/config/services/community.xml
HeahDude: https://github.com/EnMarche/en-marche.fr/blob/master/app/AppKernel.php#L29 in https://github.com/EnMarche/en-marche.fr/commit/e9b88ab550f943ba7a09eab70a3f68524767cbcd

tgalopin:
OK after the fix StyleCI

PR #1323, opened 2017-06-23
Create CV public page

illusionOfParadise:


templates/summary/_block/trainings.html.twig
```diff
@@ -0,0 +1,20 @@
...
+            <h2>Formations</h2>
+            <p>
+                {% for training in summary.trainings %}
+                    <h3>{{ training.diploma }} - {{ training.studyField }}</h3>
+                    <h4>{{ training.organization|upper }}</h4>
+                    <span>{{ training.startedAt|localizeddate('none', 'none', 'fr_FR', null, 'MMMM Y')|capitalize }} - {{ training.endedAt|localizeddate('none', 'none', 'fr_FR', null, 'MMMM Y')|capitalize }} ( todo )</span>
+                    <br />
+                    <p>{{ training.description }}</p>
+                    <br />
+                    <i>{{ training.extracurricular }}</i>
```
HeahDude: Please use an upper "C".

templates/summary/index.html.twig
```diff
@@ -0,0 +1,35 @@
+{% extends 'base.html.twig' %}
+
+{% block meta_noindex %}
```
HeahDude: the block should be named `meta_index` IMO, what do you think Titouan?

templates/summary/_block/header.html.twig
```diff
@@ -0,0 +1,21 @@
...
+    {% endblock %}
+    {% block summary_right_part %}{% endblock %}
+    {% block summary_bottom_part %}
+        <div>
+            <div>
+                {{ summary.member.activatedAt|localizeddate('none', 'none', 'fr_FR', null, 'MMMM Y') }}<br>
+                Date d'adhésion
+            </div>
+            <div>
+                23 <br> participations à des events
```
alOneh: `23` should be a variable right ?

tgalopin: Indeed, but this is the topic of another ticket as there are multiple places in the page where we have to find past data about the user.

PR #1327, opened 2017-06-25
Reorganize SCSS to fix AMP and for better readability

tgalopin:


front/style/layout/_layout.scss
```diff
@@ -193,7 +193,7 @@
 
         &--on-tablet {
             @media (max-width: $tablet) {
-                display: none !important;
+                display: none;
```
tgalopin: This is not compatible with AMP. Would this break things @marineb ?

PR #1330, opened 2017-06-26
Handled Summary trainings

HeahDude:


src/Entity/EntitySpanableTrait.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace AppBundle\Entity;
+
+use Symfony\Component\Validator\Constraints as Assert;
+
+trait EntitySpanableTrait
```
tgalopin: Spanable does not exist in English. I would simply use EntitySpanTrait

src/Form/SummaryItemType.php
```diff
@@ -0,0 +1,43 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class SummaryItemType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $item = $builder->getData();
+
+        if ($options['summary'] instanceof Summary) {
```
tgalopin: SummaryItemPositionableInterface ?

src/Summary/SummaryManager.php
```diff
@@ -52,6 +54,26 @@ public function removeExperience(Adherent $adherent, JobExperience $experience):
...
+        $this->updateSummary($summary);
+    }
+
+    public function removeTraining(Adherent $adherent, Training $training): bool
+    {
+        if (!$summary = $this->repository->findOneForAdherent($adherent)) {
+            return false;
+        }
+
+        SummaryItemDisplayOrderer::removeItem($summary->getTrainings(), $training);
```
tgalopin: This could probably be encapsulated instead of using static (perhaps for another PR after the CV).

src/Entity/EntitySpanableTrait.php
HeahDude: Adding `able` to English words is an English thing :D. But ok I renamed it.

src/Summary/SummaryManager.php
HeahDude: Ok as discussed we'll see that later.

PR #1332, opened 2017-06-27
Handled Summary steps

HeahDude:


src/Form/SummaryType.php
```diff
@@ -4,22 +4,24 @@
...
 
 class SummaryType extends AbstractType
 {
-    const STEP_SYNTHESIS = 'synthèse';
+    const STEP_SYNTHESIS = 'synthesis';
     const STEP_MISSION_WISHES = 'missions';
     const STEP_MOTIVATION = 'motivation';
-    const STEP_SKILLS = 'compétences';
-    const STEP_INTERESTS = 'centre d\'intérêts';
+    const STEP_SKILLS = 'competences';
```
tgalopin: skills

src/Form/SummaryType.php
```diff
@@ -97,20 +99,30 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ;
                 break;
         }
+
+        $builder->add('submit', SubmitType::class, ['label' => 'Enregistrer']);
```
tgalopin: not sure it's useful to add the submit here

src/Form/SummaryType.php
illusionOfParadise: Will be changed in my PR to have less conflicts

src/Form/SummaryType.php
HeahDude: It's easiest this way, and this is where it belongs IMO.

PR #1335, opened 2017-06-30
Add upload profile photo

illusionOfParadise:


src/Controller/EnMarche/SummaryManagerController.php
```diff
@@ -33,9 +34,17 @@ public function indexAction()
     {
         $this->disableInProduction();
 
+        $summary = $this->get(SummaryManager::class)->getForAdherent($this->getUser());
+        $pathImage = 'images/'.$summary->getMemberUuid().'.jpg';
```
HeahDude: What if the summary does not exist yet? Maybe we should use the summary slug instead of the adherent uuid, or both?

src/Form/SummaryType.php
```diff
@@ -47,7 +68,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                $builder
+                    ->addEventListener(FormEvents::POST_SUBMIT, function (FormEvent $event) {
+                        if (!$event->getForm()->isValid()) {
+                            return;
+                        }
+
+                        /** @var Summary $summary */
+                        $summary = $event->getData();
+                        // Save profile picture to cloud storage
+                        if ($summary->getProfilePicture()) {
```
HeahDude: We should enter that if only when this is an instance of `UploadedFile`. Please add a check for it.

src/Form/SummaryType.php
```diff
@@ -47,7 +68,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                            $pathImage = 'images/'.$summary->getMemberUuid().'.jpg';
```
HeahDude: If we were about to use the slug the logic of the listener should then move in the `SummaryManager` so it is persisted first.

src/Controller/EnMarche/SummaryManagerController.php
```diff
@@ -33,9 +34,17 @@ public function indexAction()
...
+        $signature = SignatureFactory::create($this->getParameter('kernel.secret'))->generateSignature($pathImage, []);
+
         return $this->render('summary_manager/index.html.twig', [
-            'summary' => $this->get(SummaryManager::class)->getForAdherent($this->getUser()),
+            'summary' => $summary,
             'recent_activities' => [], // TODO $this->get(MembershipTracker::class)->getRecentActivitiesForAdherent($this->getUser()),
+            'urlPhoto' => $this->generateUrl('asset_url', [
```
HeahDude: Please use snake_case for template vars name.

templates/summary_manager/handle_synthesis.html.twig
```diff
@@ -0,0 +1,18 @@
+{% extends 'summary_manager/layout.twig' %}
+
+{% block summary_content %}
+    <section class="l__wrapper--medium">
+
+        <h2>TEST synthèse</h2>
```
HeahDude: This file and title should be renamed.

src/Form/SummaryType.php
```diff
@@ -33,9 +41,58 @@ class SummaryType extends AbstractType
...
+                $builder
+                    ->addEventListener(FormEvents::POST_SUBMIT, function (FormEvent $event) {
+                        if (!$event->getForm()->isValid()) {
+                            return;
+                        }
+
+                        /** @var Summary $summary */
+                        $summary = $event->getData();
+                        // Save profile picture to cloud storage
+                        if ($summary->getProfilePicture() && $summary->getProfilePicture() instanceof UploadedFile) {
```
HeahDude: `if ($summary->getProfilePicture() instanceof UploadedFile) {` is enough.

tests/Controller/EnMarche/SummaryManagerControllerTest.php
```diff
@@ -39,6 +42,7 @@ class SummaryManagerControllerTest extends SqliteWebTestCase
 
     private const SECTIONS = [
         self::SECTION_HEADER,
+        self::SECTION_PHOTO,
```
HeahDude: If this is not a `<section>` html tag (what I guess), you should revert this.

src/Controller/EnMarche/SummaryManagerController.php
```diff
@@ -37,14 +39,39 @@ public function indexAction()
         $this->disableInProduction();
 
         $member = $this->getUser();
+        $summary = $this->get(SummaryManager::class)->getForAdherent($this->getUser());
+        $pathImage = 'images/'.$summary->getMemberUuid().'.jpg';
```
HeahDude: Maybe we could add a boolean flag on the entity to know that the CV has a picture, it would allow to count it for completion (it's currently ignored at 100%), and would avoid this logic if not necessary.

ping @tgalopin @marineb

tgalopin: good idea, but not necessary now: let's focus on releasing

HeahDude: Ok at least I would move this logic in the factory two avoid the duplication between controllers by using a getter/setter for the picture path.

tests/Controller/EnMarche/SummaryManagerControllerTest.php
```diff
@@ -1054,12 +1056,12 @@ public function testStepContactWithoutSummary()
         $this->assertSame('Vos modifications ont bien été enregistrées.', $crawler->filter('.flash__inner')->text());
         $this->assertSummaryCompletion(16, $crawler);
 
-        $missions = $this->getSummarySection($crawler, self::SECTION_HEADER);
+        $contact = $this->getSummarySection($crawler, self::SECTION_CONTACT);
```
HeahDude: there is not longer contact section, please revert that change.

PR #1346, opened 2017-07-06
Fix problem of activities number if member is not activated

illusionOfParadise:


src/Membership/MemberActivityCollection.php
```diff
@@ -18,10 +18,13 @@ class MemberActivityCollection
      */
     public function __construct(Adherent $adherent, array $eventRegistrations)
     {
-        $this->activities[self::formatDate($adherent->getActivatedAt())][] = [
-            'type' => self::TYPE_JOINED_EN_MARCHE,
-            'log' => sprintf('A rejoint le mouvement En Marche (%s)', self::formatDate($adherent->getActivatedAt())),
-        ];
+        $this->activities = [];
+        if($adherent->getRegisteredAt()) {
```
tgalopin: `if (`

src/Membership/MemberActivityCollection.php
```diff
@@ -18,10 +18,13 @@ class MemberActivityCollection
      */
     public function __construct(Adherent $adherent, array $eventRegistrations)
     {
-        $this->activities[self::formatDate($adherent->getActivatedAt())][] = [
-            'type' => self::TYPE_JOINED_EN_MARCHE,
-            'log' => sprintf('A rejoint le mouvement En Marche (%s)', self::formatDate($adherent->getActivatedAt())),
-        ];
+        $this->activities = [];
+        if ($adherent->getRegisteredAt()) {
```
HeahDude: The if is needless, registeredAt it set in the constructor.

tgalopin: We imported data from the past, we may have users without registeredAt. Let's assume this can be the case.

PR #1347, opened 2017-07-06
Publish not completed CV

illusionOfParadise:


src/Entity/Summary.php
```diff
@@ -548,11 +548,7 @@ public function isPublic(): bool
 
     public function publish(): bool
```
HeahDude: I guess it can return `void` now

PR #1377, opened 2017-07-10
Add captcha to contact form

illusionOfParadise:


src/Controller/EnMarche/AdherentController.php
```diff
@@ -192,30 +193,34 @@ public function contactAction(Request $request, Adherent $adherent): Response
             throw new BadUuidRequestException($e);
         }
 
-        $message = new ContactMessage($this->getUser(), $adherent);
+        $message = ContactMessage::createWithCaptcha((string) $request->request->get('g-recaptcha-response'), $this->getUser(), $adherent);
 
         $form = $this->createForm(ContactMessageType::class, $message);
         $form->handleRequest($request);
```
tgalopin: Won't the recaptcha check be done in handleRequest? If so, the ConnectException will be thrown outside of the try/catch block.

PR #1379, opened 2017-07-11
Add administration for banners

illusionOfParadise:


src/Admin/HomeBlockAdmin.php
```diff
@@ -89,6 +90,24 @@ protected function configureFormFields(FormMapper $formMapper)
...
             ;
         }
+
+        if (HomeBlock::TYPE_BANNER === $this->getSubject()->getType()) {
+            $formMapper
+                ->add('titleCta', null, [
+                    'label' => 'Texte du CTA',
+                    'required' => false,
+                ])
+                ->add('colorCta', ChoiceType::class, [
```
tgalopin: Could you use an input text to put the hexadecimal code instead?

src/Admin/Color.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace AppBundle\Admin;
+
+class Color
+{
+    const COLOR_BLACK = 'black';
```
HeahDude: It would be better to have `Color::BLACK` than `Color::COLOR_BLACK`.

src/Admin/HomeBlockAdmin.php
```diff
@@ -89,6 +90,24 @@ protected function configureFormFields(FormMapper $formMapper)
                 ])
             ;
         }
+
+        if (HomeBlock::TYPE_BANNER === $this->getSubject()->getType()) {
+            $formMapper
+                ->add('titleCta', null, [
```
HeahDude: We prefer snake case for field names by convention.

src/Entity/HomeBlock.php
```diff
@@ -131,15 +154,15 @@ public function getId()
     }
 
     /**
-     * @return null|string
+     * @return int
```
HeahDude: Type hints should be used instead of phpdoc when possible.

src/Entity/HomeBlock.php
```diff
@@ -282,6 +305,36 @@ public function setDisplayTitles(bool $displayTitles)
         $this->displayTitles = $displayTitles;
     }
 
+    public function isDisplayBlock(): bool
+    {
+        return $this->displayBlock;
+    }
+
+    public function setDisplayBlock(bool $displayBlock)
```
HeahDude: `: void`

src/Entity/HomeBlock.php
```diff
@@ -282,6 +305,36 @@ public function setDisplayTitles(bool $displayTitles)
...
+    {
+        $this->displayBlock = $displayBlock;
+    }
+
+    public function getTitleCta()
```
HeahDude: `: ?string`

src/Entity/HomeBlock.php
```diff
@@ -282,6 +305,36 @@ public function setDisplayTitles(bool $displayTitles)
...
+    {
+        return $this->titleCta;
+    }
+
+    public function setTitleCta($titleCta)
```
HeahDude: `public function setTitleCta(?string $titleCta): void`

and same below.

src/Admin/HomeBlockAdmin.php
illusionOfParadise: @tgalopin I have a list of colors (specified in the ticket), and more it matches the class names. Do you really prefer to use text input ?

tgalopin: Oh OK, let's go that way then.

PR #1380, opened 2017-07-12
Add administration for header banner

illusionOfParadise:


src/Controller/BannerController.php
```diff
@@ -0,0 +1,22 @@
...
+class BannerController extends Controller
+{
+    /**
+     * @Route("/header-banner")
+     * @Method("GET")
+     */
+    public function showHeaderBannerAction(): Response
+    {
+        return $this->render('banner/call.html.twig', [
+            'banner' => $this->getDoctrine()->getRepository('AppBundle:HomeBlock')->findOneBy(['position' => 11]),
```
HeahDude: Please use FQCN `HomeBlock::class` instead of doctrine notation.

src/Entity/HomeBlock.php
```diff
@@ -158,83 +158,69 @@ public function getPosition(): ?int
         return $this->position;
     }
 
-    public function setPosition(int $position): self
+    public function setPosition(int $position): void
     {
         $this->position = $position;
-
-        return $this;
```
HeahDude: Are you sure this is not breaking anything?

illusionOfParadise: Apparently, yes. I checked that there is no another method call of entity after.
Other suggestions why can it break anything ?

hhamon: Method calls chaining in Doctrine just sucks! There are no reasons for returning `$this` in setter methods just to provide a fluent interface. So better to remove them as done in this PR and update locations where methods are chained.

PR #1381, opened 2017-07-12
Add filter by city on the referent page

illusionOfParadise:


src/DataFixtures/ORM/LoadReferentManagedUserData.php
```diff
@@ -25,7 +25,7 @@ public function load(ObjectManager $manager)
             'original_id' => $this->getReference('adherent-1')->getId(),
             'email' => $this->getReference('adherent-1')->getEmailAddress(),
             'postal_code' => $this->getReference('adherent-1')->getPostalCode(),
-            'city' => $this->getReference('adherent-1')->getCity(),
+            'city' => $this->getReference('adherent-1')->getCityName(),
```
HeahDude: I don't think this should be changed.

illusionOfParadise: `getCity()` matches `address_city_insee`, but script SQL which fills the `projection_referent_managed_users` table takes values of `address_city_name`.

HeahDude: So this is a bug fix?

illusionOfParadise: I can say yes, just this method fills the tests datas and the data of this field was not really used in tests before.

tgalopin: It is indeed the city name which is currently in the database: https://github.com/EnMarche/worker/blob/master/job_referent_managed_users.go#L32

PR #1384, opened 2017-07-13
Fix error if extension is NULL

illusionOfParadise:


src/Documents/DocumentRepository.php
```diff
@@ -111,7 +111,7 @@ private function map(string $pathPrefix, array $files): array
             $directory[] = new Document(
                 $file['type'],
                 $file['filename'],
-                ('file' === $file['type']) ? $file['extension'] : '',
+                ('file' === $file['type'] && array_key_exists('extension', $file) && $file['extension']) ? $file['extension'] : '',
```
HeahDude: No need for parenthesis. What about just `$file['extension']) ?? ''`?

PR #1387, opened 2017-07-18
Fix problem caused by rendering banner controller

illusionOfParadise:


templates/base.html.twig
```diff
@@ -95,9 +95,7 @@
             </div>
         {% endfor %}
 
-        {% block banner %}
-            {{ render(controller('AppBundle:Banner:showHeaderBanner')) }}
-        {% endblock %}
+        {% block banner include('banner/call.html.twig', { banner: get_home_block(11) }) %}
```
HeahDude: 11 Should better be a constant in the `HomeBlock` class.

PR #1393, opened 2017-07-20
Change class name to manage banner background color

illusionOfParadise:


templates/banner/call.html.twig
```diff
@@ -1,5 +1,5 @@
 {% if banner.isDisplayBlock %}
-    <div class="newsletter__banner l__hide--on-tablet {{ banner.bgColor }}" id="header-banner" style="display: none;">
+    <div class="newsletter__banner l__hide--on-tablet background--{{ banner.bgColor }}" id="header-banner" style="display: none;">
```
hhamon: I would recommend to have a default value just in case the bgColor for a banner is not set.

```
background--{{ banner.bgColor|default('black') }}
```

PR #1395, opened 2017-07-20
Add constant for header banner

illusionOfParadise:


src/Twig/HomeBlockExtension.php
```diff
@@ -25,6 +25,6 @@ public function getFunctions(): array
 
     public function getHomeBlock(int $position): HomeBlock
     {
-        return $this->manager->getRepository(HomeBlock::class)->findOneBy(['position' => $position]);
+        return $this->manager->getRepository(HomeBlock::class)->findOneBy(['position' => $position]) ?? new HomeBlock();
```
HeahDude: Wouldn't it better to return a nullable block, and properly handle `null` in the template.

PR #1403, opened 2017-07-25
Update resume's interest style

marineb:


illusionOfParadise:
@marineb Fix tests and it's ok

marineb:
@illusionOfParadise oh woops. need to resume getting the circle emails. totally missed that. thank you!

PR #1416, opened 2017-07-28
Add video controls management

illusionOfParadise:


src/Admin/HomeBlockAdmin.php
```diff
@@ -114,6 +114,19 @@ protected function configureFormFields(FormMapper $formMapper)
                 ])
             ;
         }
+
+        if ($this->getSubject()->getMedia() && $this->getSubject()->getMedia()->isVideo()) {
```
HeahDude: So we cannot set those fields until the entity is saved as a video, using a listener here would be better.

PR #1470, opened 2017-08-30
[FIX] Remove FeedItem when Event is removed [branch ch678]

PaskR:


src/Repository/CommitteeFeedItemRepository.php
```diff
@@ -82,4 +83,16 @@ private function createCommitteeTimelineQueryBuilder(string $committeeUuid): Que
 
         return $qb;
     }
+
+    public function removeEventItems(BaseEvent $event)
+    {
+        $qb = $this->createQueryBuilder('i');
+        $qb->delete()
```
salahm: do you really need `expr` here ?

```php
->where('i.itemType = :type')
->andWhere('i.event = :event')
```

isn't enough ?

PaskR: Soz, it's an old habit ;)

salahm:
Also, can't you use doctrine cascade deleting instead of doing it manually ? 
something like "onDelete="CASCADE"  here https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/CommitteeFeedItem.php#L44 ?

PaskR:
Oh i totally missed this thing ! Thanks

PaskR:
PR redone here : https://github.com/EnMarche/en-marche.fr/pull/1471

PR #1479, opened 2017-09-01
A few improvements to ordonnances project

marineb:


templates/explainer/article.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html.twig' %}
+ {% extends 'base.html.twig' %}
```
illusionOfParadise: a small extra space

marineb: Woops. Thanks, fixing!

PR #1483, opened 2017-09-04
Highlight IC events

Remg:
Closes [#667](https://app.clubhouse.io/enmarche/story/667/ic-page-d-affichage-des-%C3%A9v%C3%A9nements-donner-une-couleur-diff%C3%A9rente-aux-ic-pour-les-distinguer-des-%C3%A9v%C3%A9nements-d-al)

### DIFF
--- a/templates/adherent/events.html.twig
+++ b/templates/adherent/events.html.twig
@@ -38,7 +38,7 @@
                         </div>
                     {% endif %}
 
-                    <div class="event-registration search__results__row l__row l__row--h-stretch l__row--v-center {{ event.organizer and event.organizer.isReferent ? 'search__results__row--referent-event' : '' }}">
+                    <div class="event-registration search__results__row l__row l__row--h-stretch l__row--v-center {{ event.type and event.type is same as('initiative') ? 'search__results__row--citizen_initiative' : '' }} {{ event.organizer and event.organizer.isReferent ? 'search__results__row--referent-event' : '' }}">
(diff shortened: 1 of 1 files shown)


templates/adherent/events.html.twig
```diff
@@ -54,7 +54,11 @@
                                     {{ event.beginAt|localizeddate('none', 'none', 'fr_FR', null, 'd MMMM Y') }} &mdash; à {{ event.cityName }}<br />
                                 </div>
                                 {% if event.type and event.isCitizenInitiative %}
-                                    <div class="text--body"><span class="text--bold">Initiative citoyenne</span> organisée par l'adhérent <span class="text--bold">{{ event.organizer.fullName|default('Anonyme') }}</span></div>
+                                    {% if event.organizer %}
```
illusionOfParadise: No need of `if` and even of `|default('Anonyme')`

Remg: Shouldn't we keep consistency with the event list of the search template ?
https://github.com/EnMarche/en-marche.fr/blob/master/templates/search/_events.html.twig#L27

or should we refactor both ?

Remg: got it. this check is useless on this page

illusionOfParadise: Sorry for this erroneous comment. 
This page lists events you are subscribed.
Your code is good !

PR #1485, opened 2017-09-04
Add skills admin page [ch514]

illusionOfParadise:


app/migrations/Version20170904153003.php
```diff
@@ -0,0 +1,31 @@
...
+
+class Version20170904153003 extends AbstractMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    public function up(Schema $schema)
+    {
+        foreach (LoadSkillData::SKILL_CI as $skill) {
+            $skillSlug = $this->container->get('sonata.core.slugify.cocur')->slugify($skill);
+            $existingSkill = $this->connection->executeQuery('SELECT id FROM skills WHERE slug LIKE :skillSlug', ['skillSlug' => $skillSlug]);
```
Remg: If we use a LIKE statement here then we could have collisions.
e.g: The skill named 'Veille' will match the skill named 'Veille juridique'

Why don't we use a strict comparison ?

illusionOfParadise: In this case only 'Veille' will be found. `LIKE` without wildcard is the same as strict comparison, but I'm agree that the last is better.

PR #1504, opened 2017-09-08
Tweaks to the account deletion flow

marineb:


templates/adherent/overview.html.twig
```diff
@@ -0,0 +1,28 @@
+{% extends 'adherent/layout.twig' %}
+
+{% block adherent_profile_form %}
+
+    <h2>Mon compte</h2>
+
+    <p>
```
Remg: `p` tags should not contain block level tags.

see https://www.w3.org/TR/html401/struct/text.html#h-9.3.1

> The P element represents a paragraph. It cannot contain block-level elements (including P itself).

templates/adherent/overview.html.twig
```diff
@@ -0,0 +1,28 @@
...
+        <div>
+            <span class="settings__username">{{ app.user.fullname }}</span>
+            <span class="text--gray">({{ app.user.emailaddress }})</span>.
+        </div>
+        <div class="settings__membership">Adhérent{{ app.user.gender == "female" ? "e" }} depuis {{ app.user.activatedAt|localizeddate('none', 'none', 'fr_FR', null, 'MMMM Y') }}.</div>
+    </p>
+    <p>
+    </p>
```
Remg: `p` tags should not be empty.

same link: https://www.w3.org/TR/html401/struct/text.html#h-9.3.1
> We discourage authors from using empty P elements. User agents should ignore empty P elements.

templates/adherent/overview.html.twig
marineb: 😳 I wasn't thinking. Will send a PR back shortly.

PR #1514, opened 2017-09-09
src/adress and src/admin: use keyword refactorisation & return type add & code review on possible Fatal error

ThomasDupont:
WIP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@
 /supervisord.log
 /supervisord.pid
 /yarn-error.log
+.idea
+.DS_Store
+/*.DS_Store
--- a/src/Address/GeocodableAddress.php
+++ b/src/Address/GeocodableAddress.php
@@ -6,17 +6,24 @@ use AppBundle\Intl\FranceCitiesBundle;
 
 final class GeocodableAddress
 {
-    private $address;
-    private $postalCode;
-    private $cityName;
-    private $countryCode;
+    private $_address;
+    private $_postalCode;
+    private $_cityName;
+    private $_countryCode;
 
+    /**
+     * GeocodableAddress constructor.
+     * @param string $address
+     * @param string $postalCode
+     * @param string $cityName
+     * @param string $countryCode
+     */
     public function __construct(string $address, string $postalCode, string $cityName, string $countryCode)
     {
-        $this->address = $address;
-        $this->postalCode = $postalCode;
-        $this->cityName = $cityName;
-        $this->countryCode = $countryCode;
+        $this->_address      = $address;
+        $this->_postalCode   = $postalCode;
+        $this->_cityName     = $cityName;
+        $this->_countryCode  = $countryCode;
     }
 
     public static function createFromAddress(AddressInterface $address): self
@@ -32,17 +39,17 @@ final class GeocodableAddress
     public function __toString()
     {
         $address = [];
-        if ($this->address) {
-            $address[] = str_replace(',', '', $this->address);
+        if ($this->_address) {
+            $address[] = str_replace(',', '', $this->_address);
         }
 
-        if ($this->postalCode) {
-            $address[] = sprintf('%s %s', $this->postalCode, $this->cityName);
+        if ($this->_postalCode) {
+            $address[] = sprintf('%s %s', $this->_postalCode, $this->_cityName);
         }
 
-        $countryCode = $this->countryCode;
+        $countryCode = $this->_countryCode;
         if ('FR' === $countryCode) {
-            $countryCode = FranceCitiesBundle::getCountryISOCode($this->postalCode);
+            $countryCode = FranceCitiesBundle::getCountryISOCode($this->_postalCode);
         }
         $address[] = $countryCode;
 
--- a/src/Address/PostAddressFactory.php
+++ b/src/Address/PostAddressFactory.php
@@ -6,25 +6,33 @@ use AppBundle\Entity\PostAddress;
 
 class PostAddressFactory
 {
+    /**
+     * @param null|string $country
+     * @param null|string $postalCode
+     * @param null|string $cityName
+     * @param null|string $address
+     * @return PostAddress
+     */
     public function createFlexible(?string $country, ?string $postalCode, ?string $cityName, ?string $address): PostAddress
     {
         return PostAddress::createForeignAddress($country, $postalCode, $cityName, $address);
     }
 
+    /**
+     * @param Address $address
+     * @return PostAddress
+     */
     public function createFromAddress(Address $address): PostAddress
     {
-        if ($address->isFrenchAddress()) {
-            return PostAddress::createFrenchAddress(
+        return $address->isFrenchAddress() ?
+            PostAddress::createFrenchAddress(
                 $address->getAddress(),
                 $address->getCity()
+            ) : PostAddress::createForeignAddress(
+                $address->getCountry(),
+                $address->getPostalCode(),
+                $address->getCityName(),
+                $address->getAddress()
             );
-        }
-
-        return PostAddress::createForeignAddress(
-            $address->getCountry(),
-            $address->getPostalCode(),
-            $address->getCityName(),
-            $address->getAddress()
-        );
     }
 }
--- a/src/Admin/AdherentAdmin.php
+++ b/src/Admin/AdherentAdmin.php
@@ -3,22 +3,17 @@
 namespace AppBundle\Admin;
 
 use AppBundle\Entity\Adherent;
-use AppBundle\Form\ActivityPositionType;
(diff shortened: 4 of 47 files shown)


src/Admin/CitizenInitiativeAdmin.php
```diff
@@ -27,29 +24,29 @@ class CitizenInitiativeAdmin extends AbstractAdmin
...
-        if ('edit' === $name) {
-            return 'admin/citizen_initiative/edit.html.twig';
-        }
-
-        return parent::getTemplate($name);
+        return $name === 'show' ?
+            'admin/citizen_initiative/show.html.twig' :
+            $name === 'edit' ?
+                'admin/citizen_initiative/edit.html.twig':
+                parent::getTemplate($name);
```
Remg: Ternary operators should not be stacked.

http://php.net/manual/en/language.operators.comparison.php#example-104
> It is recommended that you avoid "stacking" ternary expressions. PHP's behaviour when using more than one ternary operator within a single statement is non-obvious:

src/Address/GeocodableAddress.php
```diff
@@ -6,17 +6,24 @@
...
 {
-    private $address;
-    private $postalCode;
-    private $cityName;
-    private $countryCode;
-
+    private $_address;
+    private $_postalCode;
+    private $_cityName;
+    private $_countryCode;
```
Remg: Property visibilities should not affect their names.

http://www.php-fig.org/psr/psr-2/#42-properties

> Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.

.gitignore
```diff
@@ -17,3 +17,6 @@
 /supervisord.log
 /supervisord.pid
 /yarn-error.log
+.idea
+.DS_Store
+/*.DS_Store
```
Remg: You should add these in your global gitignore.

atamano:
@ThomasDupont thanks for the contribution. We currently follow the guidelines defined in styleci for the coding style (https://styleci.io/analyses/zOd3l0) and changing this would give us too much work on the entire project. I am closing the PR for now, do not hesitate for any suggestion.

src/Admin/AdherentAdmin.php
```diff
@@ -3,22 +3,17 @@
...
 use AppBundle\Intl\UnitedNationsBundle;
 use Misd\PhoneNumberBundle\Form\Type\PhoneNumberType;
-use Sonata\AdminBundle\Admin\AbstractAdmin;
-use Sonata\AdminBundle\Datagrid\ListMapper;
-use Sonata\AdminBundle\Datagrid\DatagridMapper;
-use Sonata\AdminBundle\Form\FormMapper;
-use Sonata\AdminBundle\Show\ShowMapper;
+use Sonata\AdminBundle\{
+    Admin\AbstractAdmin, Datagrid\ListMapper, Datagrid\DatagridMapper, Form\FormMapper, Show\ShowMapper
+};
```
Remg: You should use one use statement per class.
If not we won't be able to use `grep` anymore, for example.

http://www.php-fig.org/psr/psr-2/#namespace-and-use-declarations

> There MUST be one use keyword per declaration.

src/Admin/AdministratorFactory.php
```diff
@@ -14,6 +14,10 @@ public function __construct(EncoderFactoryInterface $encoders)
         $this->encoders = $encoders;
     }
 
+    /**
+     * @param array $data
+     * @return Administrator
+     */
```
Remg: This docblock is not really needed since it does not bring any information.
Type hints explain it all.

src/Address/PostAddressFactory.php
```diff
@@ -6,25 +6,33 @@
...
 
+    /**
+     * @param Address $address
+     * @return PostAddress
+     */
     public function createFromAddress(Address $address): PostAddress
     {
-        if ($address->isFrenchAddress()) {
-            return PostAddress::createFrenchAddress(
+        return $address->isFrenchAddress() ?
```
hhamon: The ternary operator is not suitable here as it makes the code less readable. This is why there's an explicit `if`  statement.

src/Admin/AdherentAdmin.php
```diff
@@ -3,22 +3,17 @@
...
 use Sonata\CoreBundle\Form\Type\DateRangePickerType;
-use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
-use Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter;
-use Sonata\DoctrineORMAdminBundle\Filter\DateRangeFilter;
-use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
-use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
-use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Sonata\DoctrineORMAdminBundle\{
```
hhamon: One `use` statement per page.

src/Admin/AdherentAdmin.php
```diff
@@ -31,15 +26,11 @@ class AdherentAdmin extends AbstractAdmin
...
-        if ('show' === $name) {
-            return 'admin/adherent/show.html.twig';
-        }
-
-        if ('edit' === $name) {
-            return 'admin/adherent/edit.html.twig';
-        }
-
-        return parent::getTemplate($name);
+        return $name === 'show' ?
```
hhamon: This is less readable that explicit `if` statements.

src/Admin/AdministratorFactory.php
```diff
@@ -28,6 +32,10 @@ public function createFromArray(array $data): Administrator
         return $admin;
     }
 
+    /**
+     * @param string $password
```
hhamon: Not needed

src/Admin/CitizenInitiativeAskingHelpAdmin.php
```diff
@@ -25,13 +26,24 @@ class CitizenInitiativeAskingHelpAdmin extends AbstractAdmin
 
     private $skillRepository;
 
+    /**
+     * CitizenInitiativeAskingHelpAdmin constructor.
```
hhamon: Docblock is not needed, constructor arguments must be typehinting instead.

src/Admin/ClarificationAdmin.php
```diff
@@ -30,12 +26,18 @@ class ClarificationAdmin extends AbstractAdmin
      *
      * @return Metadata
      */
-    public function getObjectMetadata($object)
+    public function getObjectMetadata(Clarification $object)
+        : Metadata
     {
-        return new Metadata($object->getTitle(), $object->getDescription(), $object->getMedia()->getPath());
+        return new Metadata(
```
hhamon: Why?

web/index.php
```diff
@@ -1,4 +1,6 @@
 <?php
+//force the strict type check
+declare(strict_types=1);
```
hhamon: We don't want to enforce strict typehint check.

src/Admin/ClarificationAdmin.php
ThomasDupont: PSR2 : Lines SHOULD NOT be longer than 80 characters; lines longer than that SHOULD be split into multiple subsequent lines of no more than 80 characters each.

Here the line lenght is over 80

src/Admin/AdministratorFactory.php
ThomasDupont: It's just to ask what is array? what should it contains?

src/Admin/CitizenInitiativeAskingHelpAdmin.php
ThomasDupont: Here, the parameters are no typed and put without explaination

src/Admin/AdherentAdmin.php
ThomasDupont: We could have a debate :)

https://wiki.php.net/rfc/group_use_declarations

src/Address/GeocodableAddress.php
ThomasDupont: You're all right, Symfony are using psr-2

src/Admin/AdministratorFactory.php
Remg: Then you should add a description to the argument:

/**
 * @param array $data The collection indexed by name.
 */

Only then the docblock brings an extra information, but is it useful ?

src/Admin/AdherentAdmin.php
Remg: Unfortunately there is no debate for this since we are following psr-2 standards:

http://www.php-fig.org/psr/psr-2/#namespace-and-use-declarations
> There MUST be one use keyword per declaration.

PR #1522, opened 2017-09-13
Added filters to summary admin page [ch200]

Remg:


src/Admin/SummaryAdmin.php
```diff
@@ -106,6 +118,115 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                     return true;
                 },
             ])
+
```
illusionOfParadise: extra line

Remg: good catch

src/Entity/MemberSummary/JobExperience.php
```diff
@@ -128,6 +128,11 @@ class JobExperience implements SummaryItemPositionableInterface
      */
     private $summary;
 
+    public function __toString(): string
+    {
+        return $this->company;
```
illusionOfParadise: You use `experiences.company` in the filtre, you don't really need this, do you ?

Remg: totally.
it was useful before we decided to split experiences.company and experiences.name

PR #1523, opened 2017-09-13
[CI] Allow connected users to create CI [ch968]

illusionOfParadise:


src/Controller/EnMarche/CitizenInitiativeController.php
```diff
@@ -298,7 +296,6 @@ public function attendConfirmationAction(Request $request, CitizenInitiative $in
      *     condition="request.isXmlHttpRequest()"
      * )
      * @Method("GET")
-     * @Security("is_granted('CREATE_CITIZEN_INITIATIVE')")
```
Remg: `@Security("is_granted('IS_AUTHENTICATED_FULLY')")` ?

or unauthenticated users may use this route ?

PR #1530, opened 2017-09-15
Fix type check for citizen initiatives [ch667]

Remg:


templates/adherent/events.html.twig
```diff
@@ -38,7 +38,7 @@
                         </div>
                     {% endif %}
 
-                    <div class="event-registration search__results__row l__row l__row--h-stretch l__row--v-center {{ event.type and event.type is same as('initiative') ? 'search__results__row--citizen_initiative' : '' }} {{ event.organizer and event.organizer.isReferent ? 'search__results__row--referent-event' : '' }}">
+                    <div class="event-registration search__results__row l__row l__row--h-stretch l__row--v-center {{ event.type and event.type is same as('citizen_initiative') ? 'search__results__row--citizen_initiative' : '' }} {{ event.organizer and event.organizer.isReferent ? 'search__results__row--referent-event' : '' }}">
```
hhamon: Can't we have an explicit method in the event class to check for the organizer:

```php
$event->isOrganizedByReferent();
```

This would simplify the Twig code:

```twig
{# Before #}
{{ event.organizer and event.organizer.isReferent ? 'search__results__row--referent-event' : '' }}

{# After #}
{{ event.organizedByReferent ? 'search__results__row--referent-event' : '' }}
```

Remg: i merged the PR when you were writing your comment :)

yes you are right ! i will do another PR for that.
this could be the very same to check if the event is a Citizen Initiative too

PR #1539, opened 2017-09-18
Update analytics on AMP pages

marineb:
I'm really not sure if what I did was right, almost sent it back to you @illusionOfParadise and @Remg, but thought I'd try. 
Here is the [card for reference](https://app.clubhouse.io/enmarche/story/539/tag-gtm-sur-pages-amp).

### DIFF
--- a/templates/amp/_layout.html.twig
+++ b/templates/amp/_layout.html.twig
@@ -126,7 +126,7 @@
     {% block content '' %}
 
     {% if not app.debug %}
-    <amp-analytics type="googleanalytics">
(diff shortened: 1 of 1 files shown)


templates/amp/_layout.html.twig
```diff
@@ -130,7 +130,8 @@
         <script type="application/json">
             {
                 "vars": {
-                    "account": "UA-72646335-1"
+                    "id": "GTM-TTW2FMQ",
```
illusionOfParadise: It's not here to add these vars. 
You have to change simply `<amp-analytics type="googleanalytics">` by `<amp-analytics config="https://www.googletagmanager.com/amp.json?id=GTM-TTW2FMQ&gtm.url=SOURCE_URL" data-credentials="include">`

PR #1543, opened 2017-09-19
Send IC absolute urls instead of slugs

Remg:


src/Mailjet/Message/CitizenInitiativeActivitySubscriptionMessage.php
```diff
@@ -90,7 +92,7 @@ private static function getTemplateVars(
             'IC_date' => $citizenInitiativeDate,
             'IC_hour' => $citizenInitiativeHour,
             'IC_address' => self::escape($citizenInitiativeAddress),
-            'IC_slug' => $citizenInitiativeSlug,
+            'IC_slug' => $citizenInitiativeLink,
```
illusionOfParadise: Is it correct that variable of slug contains a link ?

Remg: i totally agree this is kind of weird.
but for now, template variable names are completely decoupled from our app in Mailjet dashboard.

If i want to change the variable name here, it would need an update in the MailJet dashboard as well.

Remg: let's change the variable name!

PR #1550, opened 2017-09-20
Add pointer cursor to videos [ch628]

illusionOfParadise:


front/style/layout/_banners.scss
```diff
@@ -189,3 +189,7 @@
         }
     }
 }
+
+.pointer {
```
marineb: I'm thinking maybe adding it to the base file. And making it `cursor--pointer` so that if we need to introduce more cursor styles, we can do that in the future. 

Btw sorry, after I sent this back I thought I should have probably done it. It wasn't much of a back-end task. :)

PR #1551, opened 2017-09-20
Transform adherent to coordinator

illusionOfParadise:


src/Admin/AdherentAdmin.php
```diff
@@ -118,6 +121,11 @@ protected function configureShowFields(ShowMapper $showMapper)
                     'label' => 'Accepte de recevoir des SMS de la part d\'En Marche !',
                 ])
             ->end()
+            ->with('Candidat législatives', ['class' => 'col-md-6'])
+            ->add('legislativeCandidate', 'boolean', [
```
Remg: you should set back the indentation here

src/Admin/AdherentAdmin.php
```diff
@@ -195,6 +195,22 @@ protected function configureFormFields(FormMapper $formMapper)
                         'Utiliser les codes de pays (FR, DE, ...) ou des préfixes de codes postaux.',
                 ])
             ->end()
+            ->with('Référent', ['class' => 'col-md-3'])
+            ->add('managedArea.codesAsString', TextType::class, [
```
Remg: same here, one level of indentation is missing under the ->with()

PR #1553, opened 2017-09-21
Uncouple Mailjet from the application

Remg:


src/Consumer/AbstractMailerConsumer.php
```diff
@@ -0,0 +1,66 @@
...
+namespace AppBundle\Consumer;
+
+use AppBundle\Mailer\ClientInterface;
+use AppBundle\Mailer\Model\EmailRepository;
+use GuzzleHttp\Exception\ConnectException;
+use Symfony\Component\Validator\Constraints as Assert;
+
+abstract class AbstractMailerConsumer extends AbstractConsumer
+{
+    const CLIENT_ID = 'to_override';
```
hhamon: Overriding constants looks weird to me. If the value are "variable", then they should be defined as attributes.

src/Mailer/EmailTemplateFactory.php
```diff
@@ -0,0 +1,24 @@
...
+    public function __construct(string $senderEmail, string $senderName, string $templateClass)
+    {
+        $this->senderEmail = $senderEmail;
+        $this->senderName = $senderName;
+        $this->templateClass = $templateClass;
+    }
+
+    public function createFromMessage(Message $message): EmailTemplate
+    {
+        return $this->templateClass::createWithMessage($message, $this->senderEmail, $this->senderName);
```
hhamon: This is very fragile...

hhamon: At least use `call_user_func_array()`

src/Mailer/Message/CitizenInitiativeActivitySubscriptionMessage.php
```diff
@@ -53,7 +54,7 @@ public static function create(
             Uuid::uuid4(),
             '196524',
             $recipient->getEmailAddress(),
-            self::fixMailjetParsing($recipient->getFullName()),
+            MailjetUtils::fixMailjetParsing($recipient->getFullName()),
```
hhamon: the static method still uses `mailjet` keyword.

src/Mailer/Message/CitizenInitiativeActivitySubscriptionMessage.php
```diff
@@ -67,7 +68,7 @@ public static function create(
         foreach ($recipients as $recipient) {
             $message->addRecipient(
                 $recipient->getEmailAddress(),
-                self::fixMailjetParsing($recipient->getFullName()),
+                MailjetUtils::fixMailjetParsing($recipient->getFullName()),
```
hhamon: same here.

src/Mailer/Message/CitizenInitiativeActivitySubscriptionMessage.php
Remg: yes i think i will make the ApiClient do this parsing, as they could be different for each api

src/Mailer/Message/CitizenInitiativeActivitySubscriptionMessage.php
Remg: yes, all subjects are fixed this way. same as above, will move it in the ApiClient logic

src/Consumer/AbstractMailerConsumer.php
Remg: i did not change the behavior. but i agree, let's change this

Remg: what about `abstract protected function getClientId(): string;` so that we ensure i'ts a non-nullable string? no property needed then?

src/Mailer/EventSubscriber/EmailPersisterEventSubscriber.php
```diff
@@ -23,25 +22,25 @@ public function __construct(ObjectManager $manager, MailjetEmailRepository $repo
...
     }
 
-    public function onMailjetDeliveryMessage(MailjetEvent $event): void
+    public function onMailerDeliveryMessage(MailerEvent $event): void
     {
         $email = $event->getEmail();
         $message = $event->getMessage();
 
-        $this->manager->persist(MailjetEmail::createFromMessage($message, $email->getHttpRequestPayload()));
+        $this->manager->persist($this->repository->getClassName()::createFromMessage($message, $email->getHttpRequestPayload()));
```
Remg: todo: use `call_user_func_array` here

src/Mailer/EmailTemplateFactory.php
```diff
@@ -0,0 +1,28 @@
...
+    public function __construct(string $senderEmail, string $senderName, string $templateClass)
+    {
+        $this->senderEmail = $senderEmail;
+        $this->senderName = $senderName;
+        $this->templateClass = $templateClass;
+    }
+
+    public function createFromMessage(Message $message): EmailTemplate
+    {
+        return call_user_func_array([$this->templateClass, 'createWithMessage'], [
```
hhamon: We should add a test before to ensure the method exists in the class or throw an exception instead.

src/Mailer/Message/AdherentAccountActivationMessage.php
```diff
@@ -1,19 +1,19 @@
...
-final class AdherentAccountActivationMessage extends MailjetMessage
+final class AdherentAccountActivationMessage extends Message
 {
     public static function createFromAdherent(Adherent $adherent, string $confirmationLink): self
     {
         return new self(
             Uuid::uuid4(),
             '54665',
             $adherent->getEmailAddress(),
-            self::fixMailjetParsing($adherent->getFullName()),
```
hhamon: Why removing this?

tests/Mailer/MailerServiceTest.php
```diff
@@ -0,0 +1,54 @@
...
+use Tests\AppBundle\Test\Mailer\Message\DummyMessage;
+use Tests\AppBundle\Test\Mailer\Transport\FailingTransport;
+use Tests\AppBundle\Test\Mailer\Transport\NullTransport;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+class MailerServiceTest extends TestCase
+{
+    public function testSendMessage()
+    {
+        $dispatcher = $this->getMockBuilder(EventDispatcherInterface::class)->getMock();
```
hhamon: `$dispatcher = $this->createMock(EventDispatcherInterface::class);`

src/Mailer/Message/AdherentAccountActivationMessage.php
Remg: i moved this logic inside the AppBundle\Mailjet\EmailTemplate class.
It's Mailjet specific now!

src/Mailjet/EmailTemplate.php
```diff
@@ -28,32 +17,23 @@ public function __construct(
...
+            $senderEmail,
+            $senderName,
+            $replyTo,
+            $cc
+        );
     }
 
     public function addRecipient(string $email, string $name = null, array $vars = []): void
     {
+        $name = $this->fixMailjetParsing($name);
```
Remg: Todo: move this parsing in the condition below

PR #1557, opened 2017-09-24
Add supervisor on workers

atamano:


docker/prod/supervisord-mailjet-campaign.conf
```diff
@@ -0,0 +1,11 @@
+[supervisord]
+nodaemon=true
+
+[program:mailjet-campaign]
+command=php /app/bin/console rabbitmq:consumer -e prod -w mailjet_campaign
```
B-Galati: Don't we use environment variable in prod so that we don't need to specify `-e prod` ?
Then this file would not be prod specific.

atamano: Every worker is currently pushed on separates docker images (https://github.com/EnMarche/en-marche.fr/blob/master/circle.yml#L70) and each image executes one specific command.
This could be refactored so the same image can execute different commands, I will investigate this.

PR #1579, opened 2017-10-03
Add ucfirst on name for Event

NyKo24:


NyKo24:
Need add functional test

PR #1581, opened 2017-10-04
Create adherent tags admin page

illusionOfParadise:


src/Entity/AdherentTag.php
```diff
@@ -0,0 +1,68 @@
...
+     * @Assert\NotBlank
+     * @Assert\Length(max="100")
+     *
+     * @Algolia\Attribute
+     */
+    private $name = '';
+
+    public function __construct(?string $name = null)
+    {
+        if ($name) {
```
Remg: Why would we allow a tag without a defined name ?

PR #1593, opened 2017-10-06
Fix referent filters

Remg:


tests/TestHelperTrait.php
```diff
@@ -192,9 +192,9 @@ public function getMailjetEmailRepository(): MailjetEmailRepository
         return $this->getRepository(MailjetEmail::class);
     }
 
-    public function getSummaryRepository(): SummaryRepository
```
illusionOfParadise: Are you sure that this method is not used anywhere ?
It could be better to create new method and not change existing one.

Remg: pure mistake!

PR #1599, opened 2017-10-09
Uncouple Mailjet from the application

Remg:


src/Contact/ContactMessageHandler.php
```diff
@@ -2,20 +2,20 @@
...
+    public function __construct(MailerService $mailer)
     {
-        $this->mailjet = $mailjet;
+        $this->mailer = $mailer;
     }
 
     public function handle(ContactMessage $contactMessage)
     {
-        $this->mailjet->sendMessage(AdherentContactMessage::createFromMdel($contactMessage));
+        $this->mailer->sendMessage(AdherentContactMessage::createFromMdel($contactMessage));
```
HeahDude: Not related to that PR but we've go a typo in the static method name here.

src/Mailer/EmailTemplateFactory.php
```diff
@@ -0,0 +1,38 @@
...
+        $this->senderName = $senderName;
+        $this->templateClass = $templateClass;
+    }
+
+    public function createFromMessage(Message $message): EmailTemplate
+    {
+        $callable = [$this->templateClass, 'createWithMessage'];
+
+        if (!is_callable($callable)) {
+            throw new MailerException(sprintf(
```
HeahDude: We should use a `\LogicException` here.

src/Mailer/Event/MailerEvents.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace AppBundle\Mailer\Event;
+
+final class MailerEvents
+{
+    const DELIVERY_MESSAGE = 'mailer.delivery.message';
```
HeahDude: Missing public scope.

src/Mailer/EventSubscriber/EmailPersisterEventSubscriber.php
```diff
@@ -0,0 +1,71 @@
...
+            MailerEvents::DELIVERY_SUCCESS => 'onMailerDeliverySuccess',
+        ];
+    }
+
+    public function onMailerDeliveryMessage(MailerEvent $event): void
+    {
+        $email = $event->getEmail();
+        $message = $event->getMessage();
+
+        $callable = [$this->repository->getClassName(), 'createFromMessage'];
```
HeahDude: Why not using the factory here instead of repeating the logic?

src/Mailjet/EmailClient.php
```diff
@@ -0,0 +1,24 @@
...
+class EmailClient extends AbstractEmailClient implements EmailClientInterface
+{
+    public function sendEmail(string $email): string
+    {
+        $response = $this->request('POST', 'send', [
+            'body' => $email,
+        ]);
+
+        if (Response::HTTP_OK !== $response->getStatusCode()) {
+            throw new MailjetException('Unable to send email to recipients.');
```
HeahDude: `MailjetException` 🤔

src/Mailjet/EmailTemplate.php
```diff
@@ -120,31 +70,8 @@ public function getBody(): array
...
-        return $this->httpRequestPayload;
-    }
-
-    public function getHttpResponsePayload(): ?string
-    {
-        return $this->httpResponsePayload;
-    }
-
-    public function jsonSerialize(): array
+    private function fixMailjetParsing(string $string = null): ?string
```
HeahDude: When does it return null?

src/Mailjet/EmailTemplate.php
```diff
@@ -120,31 +70,8 @@ public function getBody(): array
...
     {
-        return $this->getBody();
+        return str_replace(',', '', $string);
```
HeahDude: You don't need to call this function if `null` is passed.

src/Mailjet/EmailTemplate.php
HeahDude: Is the parameter really optional? Otherwise it should better be:
```php
private function fixMailjetParsing(?string $string): ?string

File: src/Mailjet/Exception/MailjetException.php [hunk 8]
```diff
@@ -2,6 +2,8 @@
 
 namespace AppBundle\Mailjet\Exception;
 
-class MailjetException extends \RuntimeException
+use AppBundle\Mailer\Exception\MailerException;
+
+class MailjetException extends MailerException
```
HeahDude: Is this class still needed?

app/config/config.yml
```diff
@@ -224,15 +224,15 @@ old_sound_rabbit_mq:
             heartbeat: 5
 
     producers:
-        mailjet_campaign:
+        mailer_campaign:
```
blazarecki: We need to create the new exchanges and queues before deploying this. @atamano

blazarecki: Same for the worker `-w mailjet_transactional` -> `-w mailer_transactional `

app/config/security.yml
```diff
@@ -134,10 +134,10 @@ security:
             - ROLE_APP_ADMIN_DONATION_LIST
             - ROLE_APP_ADMIN_DONATION_VIEW
 
-        ROLE_ADMIN_MAILJET:
-            - ROLE_APP_ADMIN_MAILJET_EMAIL_EXPORT
-            - ROLE_APP_ADMIN_MAILJET_EMAIL_LIST
-            - ROLE_APP_ADMIN_MAILJET_EMAIL_VIEW
+        ROLE_ADMIN_EMAIL:
```
blazarecki: You need to migrate the data.

app/config/config.yml
Remg: and `-w mailjet_campaign` -> `-w mailer_campaign`

app/config/security.yml
Remg: done

app/config/config.yml
atamano: yes Im on it

PR #1632, opened 2017-10-17
Parallelize test suite execution

Remg:


tests/Controller/EnMarche/AdherentControllerTest.php
```diff
@@ -23,6 +23,7 @@
 
 /**
  * @group functional
+ * @group controller-1
```
blazarecki: You can use meaningful name like `security`, `miscellaneous` etc.

Remg: you are right, but i just want to create 3 groups (or 2 tags), and these few controller class all have different meanings. and i dont want my command to be very long because i have to manage tags

Remg: +1

PR #1636, opened 2017-10-17
Add user password to rabbitmq connexion

atamano:


app/config/config.yml
```diff
@@ -211,6 +211,9 @@ old_sound_rabbit_mq:
     connections:
         default:
             host: "%env(RABBITMQ_HOST)%"
+            port: "%env(RABBITMQ_PORT)%"
+            user: "%env(RABBITMQ_USER)%"
+            password: "%env(RABBITMQ_PASSWORD)%"
             vhost: '/'
             lazy: false
             connection_timeout: 86400
```
blazarecki: One day for timeout ?

atamano: fixed

PR #1639, opened 2017-10-17
[BOARD_MEMBERS] Add link to access on board member area on navbar

NyKo24:


templates/base.html.twig
```diff
@@ -280,6 +285,11 @@
                                         <a href="{{ path('app_referent_users') }}">Espace référent</a>
                                     </li>
                                 {% endif %}
+                                {% if is_granted('ROLE_BOARD_MEMBER') %}
+                                    <li class="text--small">
+                                        <a href="{{ path('app_board_member_home') }}">Espace membres du conseil</a>
```
marineb: Singulier membre du conseil. :) On a change

PR #1648, opened 2017-10-18
Add correct php.ini for php-fpm

NyKo24:


atamano:
@NyKo24 is this issue also present in production ?

NyKo24:
Yes

PR #1653, opened 2017-10-18
Notify if no mail is sent to any referent

Remg:
- a committee can now be managed by more than one referent
- no mail is sent to referents after a committee approval if more than one referent is found
- admins should be notified about this eventual referent conflict

### DIFF
--- a/tests/Committee/CommitteeManagementAuthorityTest.php
+++ b/tests/Committee/CommitteeManagementAuthorityTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\AppBundle\Committee;
 
+use AppBundle\Collection\AdherentCollection;
 use AppBundle\Committee\CommitteeManagementAuthority;
 use AppBundle\Committee\CommitteeManager;
 use AppBundle\Committee\CommitteeUrlGenerator;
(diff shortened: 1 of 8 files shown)


tests/Committee/CommitteeManagementAuthorityTest.php
```diff
@@ -14,53 +15,121 @@
...
+        $urlGenerator = $this->createMock(CommitteeUrlGenerator::class);
+        $urlGenerator->expects($this->never())->method('generate')->with($this->anything());
+
+        $committeeManagementAuthority = new CommitteeManagementAuthority($manager, $urlGenerator, $mailer);
+        $committeeManagementAuthority->notifyReferentsForApproval($committee);
+    }
+
+    private function createCommittee(string $uuid, string $cityName): Committee
+    {
+        $committeeUuid = $this->createMock(Uuid::class);
```
hhamon: You should use the real ` Uuid`  class here. This is a value object so it's lightweight compared to a mock object.

tests/Committee/CommitteeManagementAuthorityTest.php
```diff
@@ -14,53 +15,121 @@
...
+        $committee = $this->createMock(Committee::class);
+        $committee->expects($this->any())->method('getUuid')->willReturn($committeeUuid);
+        $committee->expects($this->any())->method('getCityName')->willReturn($cityName);
+
+        return $committee;
+    }
+
+    private function createAnimator(string $uuid): Adherent
+    {
+        $animatorUuid = $this->createMock(Uuid::class);
```
hhamon: Same here, don't mock value objects.

tests/Committee/CommitteeManagementAuthorityTest.php
Remg: :+1:

PR #1657, opened 2017-10-19
[WIP] Add support for updating Adherent user account data.

hhamon:


app/config/config.yml
```diff
@@ -147,6 +147,12 @@ csa_guzzle:
     profiler: '%kernel.debug%'
     logger: true
     clients:
+        enmarche_api:
+            config:
+                base_uri: 'https://api.en-marche.dev'
```
hhamon: Must be set as env var.

PR #1668, opened 2017-10-23
Fix typo on committee page

NyKo24:


templates/coordinator/_committee_list.html.twig
```diff
@@ -28,7 +28,7 @@
             <td>
                 {% for messages in app.flashes(["error_#{committee.id}"]) %}
                     {% for message in messages %}
-                        <div class="form__errors">
```
marineb: Why are you changing this?

NyKo24: because the class name definition in CSS is form__error

marineb: I don't believe so. There is a `form__error` and a `form__errors`. Look for `__errors` and you'll find it.

templates/coordinator/_layout.html.twig
```diff
@@ -11,9 +11,9 @@
                     {{ block('page_title') }}
                 </h1>
                 <p class="text--center">
-                    L’espace coordinateur vous permet d'apprecier les comités des
+                    L’espace coordinateur vous permet d'apprécier les comités des
                     zones que vous gérez. Seule l’équipe du QG sera en mesure de
-                    voir vos appréciations.
+                    voir vos&nbsp;appréciations.
```
marineb: `&nbsp;` ?

PR #1681, opened 2017-10-24
Fix route generation of article

NyKo24:


illusionOfParadise:
This fix is also in my PR : https://github.com/EnMarche/en-marche.fr/pull/1684

PR #1682, opened 2017-10-24
fix throw exception on DynamicRedirectionsSubscriber

NyKo24:


src/Redirection/DynamicRedirectionsSubscriber.php
```diff
@@ -72,8 +74,16 @@ public function onKernelException(GetResponseForExceptionEvent $event)
                 continue;
             }
 
+            try {
+                $routeParams = $this->router->match($requestUri);
+            } catch (ResourceNotFoundException $resourceNotFoundException) {
```
illusionOfParadise: Instead of 2 catches you can write `catch (\RuntimeException $e)` because two exception classes extend this one. Or you can write `catch (ResourceNotFoundException | MethodNotAllowedException $e)` as we use PHP 7.1

PR #1697, opened 2017-10-26
[GROUPS] MOOC event categories

Remg:


src/Entity/BaseEventCategory.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        return $this->name ?: '';
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getName(): string
```
HeahDude: Constructor does allow null.

app/migrations/Version20171026170529.php
```diff
@@ -0,0 +1,27 @@
...
+class Version20171026170529 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('CREATE TABLE mooc_event_categories (id INT UNSIGNED AUTO_INCREMENT NOT NULL, name VARCHAR(100) NOT NULL, UNIQUE INDEX mooc_event_category_name_unique (name), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
+    }
+
+    public function postUp(Schema $schema)
+    {
+        foreach (LoadMoocEventCategoryData::MOOC_EVENT_CATEGORIES as $category) {
```
illusionOfParadise: Category datas seem as testing datas. Do we really need to insert them to base ?

Remg: It's needed if we want fixtures of event categories.

PR #1699, opened 2017-10-27
[GROUPS] Create entities and tables for groups

illusionOfParadise:


src/Entity/Adherent.php
```diff
@@ -789,6 +828,19 @@ public function setCoordinatorManagedAreaCodesAsString(string $codes = null): vo
...
     }
 
+    final public function getGroupMemberships(): GroupMembershipCollection
+    {
+        if ($this->groupMemberships instanceof Collection) {
+            if (!$this->groupMemberships instanceof GroupMembershipCollection) {
+                $this->groupMemberships = new GroupMembershipCollection($this->groupMemberships->toArray());
+            }
+        } else {
+            $this->groupMemberships = new GroupMembershipCollection((array) $this->groupMemberships);
```
HeahDude: What's supposed to be casted here? I don't think this is necessary to handle `iterable` like this. An `Iterator` would have to pas passed to `iterator_to_array`.
I suggest to remove this overingeenering.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        return Uuid::uuid5(Uuid::NAMESPACE_OID, static::canonicalize($name));
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public function getSlug(): string
```
HeahDude: Name and slug should be initialized as empty strings otherwise those getters will break.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public function getPostAddress(): PostAddress
+    {
+        return $this->postAddress;
+    }
+
+    public function getDescription(): string
+    {
+        return $this->description;
```
HeahDude: Same here.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public function isRefused(): bool
+    {
+        return self::REFUSED === $this->status;
+    }
+
+    public function getMembersCount(): int
+    {
+        return $this->membersCounts;
```
HeahDude: Property not initialized as 0, getter may break.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public function incrementMembersCount(int $increment = 1)
```
HeahDude: Missing `void`.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        $this->membersCounts += $increment;
+    }
+
+    public function decrementMembersCount(int $increment = 1)
```
HeahDude: Same here.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+
+    /**
+     * Marks this group as refused/rejected.
+     *
+     * @param string $timestamp
+     */
+    public function refused(string $timestamp = 'now')
+    {
+        $this->status = self::REFUSED;
+        $this->refusedAt = new \DateTime($timestamp);
```
HeahDude: We should use immutable objects when possible.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public static function canonicalize(string $name): string
+    {
+        return mb_strtolower($name);
+    }
+
+    public function getCreatedBy(): ?string
+    {
+        return $this->createdBy->toString();
```
HeahDude: Property not initialized, the `toString` call might break.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        return $this->phone;
+    }
+
+    /**
+     * Returns the approval date and time.
+     *
+     * @return \DateTime|null
+     */
+    public function getApprovedAt(): ?\DateTime
```
HeahDude: Please use the interface for type hints, so we can use both mutable and immutable date time objects.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        return $this->approvedAt;
+    }
+
+    public function updateSlug(string $slug)
```
HeahDude: Missing `void`.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        $this->slug = $slug;
+    }
+
+    public function equals(self $other): bool
+    {
+        return $this->uuid->equals($other->getUuid());
```
HeahDude: `equals` might break.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public function update(string $name, string $description, PostAddress $address, PhoneNumber $phone)
```
HeahDude: Missing `void`.

src/Entity/BaseGroup.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        $this->setName($name);
+        $this->description = $description;
+
+        if (!$this->postAddress->equals($address)) {
```
HeahDude: `equals` might break if `$this->postAddress` is `null`. Please consider adding a constructor in this class.

src/Entity/Group.php
```diff
@@ -0,0 +1,112 @@
...
+ *     @ORM\Index(name="group_status_idx", columns="status")
+ *   }
+ * )
+ * @ORM\Entity(repositoryClass="AppBundle\Repository\GroupRepository")
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class Group extends BaseGroup
+{
+    public function __construct(
```
HeahDude: Constructor should live in the base class.

src/Entity/Group.php
```diff
@@ -0,0 +1,112 @@
...
+        PostAddress $address = null,
+        PhoneNumber $phone = null,
+        string $slug = null,
+        string $status = self::PENDING,
+        string $approvedAt = null,
+        string $createdAt = 'now',
+        int $membersCount = 0
+    ) {
+        if ($approvedAt) {
+            $approvedAt = new \DateTime($approvedAt);
```
HeahDude: Please use immutable objects.

src/Entity/Group.php
```diff
@@ -0,0 +1,112 @@
...
+
+        return $group;
+    }
+
+    /**
+     * Marks this committee as approved.
+     *
+     * @param string $timestamp
+     */
+    public function approved(string $timestamp = 'now')
```
HeahDude: Missing `void`.

src/Entity/GroupFeedItem.php
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * @ORM\Table(name="group_feed_items")
+ * @ORM\Entity(repositoryClass="AppBundle\Repository\GroupFeedItemRepository")
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class GroupFeedItem
+{
+    const MESSAGE = 'message';
```
HeahDude: Missing scope.

src/Entity/GroupFeedItem.php
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * @var Adherent Any host of the group
+     *
+     * @ORM\ManyToOne(targetEntity="Adherent", inversedBy="groupFeedItems")
+     */
+    private $author;
+
+    /**
+     * @var BaseEvent
```
HeahDude: `BaseEvent|null`

src/Entity/GroupFeedItem.php
```diff
@@ -0,0 +1,180 @@
...
+     */
+    private $content;
+
+    /**
+     * @ORM\Column(type="boolean", options={"default": true})
+     */
+    private $published = true;
+
+    /**
+     * @var \DateTime|\DateTime|null
```
HeahDude: Weird, please use the interface instead.

src/Entity/GroupFeedItem.php
```diff
@@ -0,0 +1,180 @@
...
+    public function getContent(): ?string
+    {
+        if ($this->event instanceof Event) {
+            return $this->event->getDescription();
+        }
+
+        return $this->content;
+    }
+
+    public function getGroup(): ?Group
```
HeahDude: Does not seem to be nullable, please consider using a strict return type hint.

src/Entity/GroupMembership.php
```diff
@@ -0,0 +1,231 @@
...
+ *     @ORM\Index(name="groups_memberships_role_idx", columns="privilege")
+ *   }
+ * )
+ * @ORM\Entity(repositoryClass="AppBundle\Repository\GroupMembershipRepository")
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class GroupMembership
+{
+    const GROUP_ADMINISTRATOR = 'HOST';
```
HeahDude: Missing scope.

src/Entity/GroupMembership.php
```diff
@@ -0,0 +1,231 @@
...
+     * The group UUID.
+     *
+     * @var UuidInterface
+     *
+     * @ORM\Column(type="uuid")
+     */
+    private $groupUuid;
+
+    /**
+     * @var Adherent|null
```
HeahDude: Does not seem to be nullable.

src/Entity/GroupMembership.php
```diff
@@ -0,0 +1,231 @@
...
+    private $joinedAt;
+
+    /**
+     * Constructor.
+     *
+     * @param UuidInterface $uuid
+     * @param UuidInterface $groupUuid
+     * @param Adherent      $adherent
+     * @param string        $privilege
+     * @param string        $subscriptionDate
```
HeahDude: All these phpdocs can be removed as they do not bring any value.

src/Entity/GroupMembership.php
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    public static function checkPrivilege(string $privilege): void
+    {
+        if (!in_array($privilege, self::PRIVILEGES, true)) {
+            throw new \InvalidArgumentException(sprintf('Invalid privilege %s', $privilege));
+        }
+    }
+
+    private static function createUuid(UuidInterface $adherentUuid, UuidInterface $groupUuid): UuidInterface
```
HeahDude: Private static methods should be declared at last in the class.

src/Entity/GroupMembership.php
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Returns whether or not this membership enables the adherent to administrate a group.
+     */
+    public function canAdministrateGroup(): bool
+    {
+        return self::GROUP_ADMINISTRATOR === $this->privilege;
+    }
+
+    public function getAdherent(): ?Adherent
```
HeahDude: Should not be nullable.

src/Entity/GroupMembership.php
```diff
@@ -0,0 +1,231 @@
...
+        if (!$this->isDemotableAdministrator()) {
+            throw GroupMembershipException::createNotDemotableFollowerPrivilegeException($this->uuid);
+        }
+
+        $this->privilege = self::GROUP_FOLLOWER;
+    }
+
+    public function getSubscriptionDate(): \DateTimeImmutable
+    {
+        return new \DateTimeImmutable($this->joinedAt->format(DATE_RFC822), $this->joinedAt->getTimezone());
```
HeahDude:
```php
return \DateTimeImmutable::createFromMutable($this->joinedAt);
```

src/Exception/BaseGroupException.php
```diff
@@ -0,0 +1,27 @@
...
+    private $groupUuid;
+
+    public function __construct(UuidInterface $groupUuid, $message = '', \Exception $previous = null)
+    {
+        parent::__construct($message, 0, $previous);
+
+        $this->groupUuid = $groupUuid;
+    }
+
+    public function getCommitteeUuid(): UuidInterface
```
HeahDude: Does it make sense to have two methods that return the same value. The name is is misleading though.

src/Entity/BaseGroup.php
illusionOfParadise: Initialized in constructor.

src/Entity/BaseGroup.php
illusionOfParadise: Initialized in constructor.

src/Entity/BaseGroup.php
illusionOfParadise: Initialized in constructor.

src/Entity/BaseGroup.php
illusionOfParadise: Uuid is initialized in constructor.

src/Entity/Group.php
illusionOfParadise: Added.

PR #1704, opened 2017-10-28
[REFERENT] Keep track of filters and sent mails

Remg:


src/Consumer/ReferentMessageDispatcherConsumer.php
```diff
@@ -55,7 +53,10 @@ public function doExecute(array $data): bool
 
                 if ($i === MailjetService::PAYLOAD_MAXSIZE) {
                     $mailer->sendMessage(MailjetMessage::createFromModel($message, $chunk));
-                    $this->writeln($data['referent_uuid'], 'Message from '.$referent->getEmailAddress().' dispatched ('.$count.')');
+                    $this->writeln($data['uuid'], 'Message from '.$referent->getEmailAddress().' dispatched ('.$count.')');
+
+                    $referentMessage->incrementOffset($i);
+                    $manager->flush();
```
HeahDude: You should flush after the loop.

src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class ReferentManagedUsersMessage
+{
+    use EntityIdentityTrait;
+    use EntityTimestampableTrait;
+
+    /**
+     * @var Adherent
```
HeahDude: The mapping implies that it might be null in the future if the adherent is deleted.

src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -0,0 +1,196 @@
...
+     */
+    private $includeAdherentsInCommittee;
+
+    /**
+     * @ORM\Column(type="boolean")
+     */
+    private $includeHosts;
+
+    /**
+     * @ORM\Column(type="boolean")
```
HeahDude: All booleans above should use a default value if not nullable.

src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -0,0 +1,196 @@
...
+            $message->getFilter()->getQueryId()
+        );
+    }
+
+    public function incrementOffset(int $offset): void
+    {
+        $this->offset += $offset;
+    }
+
+    public function getFrom(): Adherent
```
HeahDude: Nullable.

src/Producer/ReferentMessageDispatcherProducer.php
```diff
@@ -9,6 +9,8 @@ class ReferentMessageDispatcherProducer extends Producer implements ReferentMess
 {
     public function scheduleDispatch(ReferentMessage $message): void
     {
-        $this->publish(\GuzzleHttp\json_encode($message->toArray()));
+        $this->publish(json_encode([
```
HeahDude: You should keep the guzzle one, it has a better error handling.

src/Repository/Projection/ReferentManagedUserRepository.php
```diff
@@ -43,7 +43,13 @@ public function createDispatcherIterator(Adherent $referent, ManagedUsersFilter
         $qb = $this->createFilterQueryBuilder($referent, $filter);
         $qb->andWhere('u.isMailSubscriber = 1');
 
-        return $qb->getQuery()->iterate();
+        $query = $qb->getQuery();
+
+        if ($filter) {
+            $query->setFirstResult($filter->getOffset());
```
HeahDude: Why not `$qb`? To keep the query call inlined.

src/Consumer/ReferentMessageDispatcherConsumer.php
Remg: Actually, this whole PR is for this flush.
if the consumer fails during the loop, we need to know the "offset" value for the next try

src/Consumer/ReferentMessageDispatcherConsumer.php
```diff
@@ -85,6 +82,29 @@ public function doExecute(array $data): int
...
+
+        try {
+            $this->mailer->sendMessage(MailjetMessage::createFromModel($message, $recipients));
+
+            $this->writeln(
+                $savedMessage->getUuid()->toString(),
+                'Message from '.$message->getFrom()->getEmailAddress().' dispatched ('.$count.')'
+            );
+
+            $this->referentMessageRepository->incrementOffset($savedMessage, count($recipients));
```
HeahDude: I would remove this of the try, otherwise this line still can be by-passed:
```php
try {
    $this->mailer->sendMessage(MailjetMessage::createFromModel($message, $recipients));

    $this->writeln(
        $savedMessage->getUuid()->toString(),
        'Message from '.$message->getFrom()->getEmailAddress().' dispatched ('.$count.')'
    );

    $this->getDoctrine()->getConnection()->commit();
} catch (\Throwable $error) {
    $this->getManager()->close();
    $this->getDoctrine()->getConnection()->rollback();
} finally {
    $this->referentMessageRepository->incrementOffset($savedMessage, count($recipients));
    $this->getDoctrine()->getManager()->flush();

    if (isset($error)) {
        throw $error;
    }
}
```
Consider keeping the reference of the connection to remove some calls.

src/Entity/ReferentManagedUsersMessage.php
Remg: not anymore: `onDelete="CASCADE"`

src/Consumer/ReferentMessageDispatcherConsumer.php
```diff
@@ -86,9 +82,32 @@ public function doExecute(array $data): int
...
+
+            $this->writeln(
+                $savedMessage->getUuid()->toString(),
+                'Message from '.$message->getFrom()->getEmailAddress().' dispatched ('.$count.')'
+            );
+
+            $this->getReferentMessageRepository()->incrementOffset($savedMessage, count($recipients));
+
+            $this->getManager()->getConnection()->commit();
+        } catch (\Exception $error) {
```
HeahDude: `\Throwable`

src/Controller/EnMarche/ReferentController.php
```diff
@@ -65,13 +64,13 @@ public function usersSendMessageAction(Request $request): Response
...
 
-        $message = new ReferentMessage($this->getUser(), $filter);
+        $message = $this->get('app.referent.message_notifier')->createMessage($this->getUser(), $filter);
 
         $form = $this->createForm(ReferentMessageType::class, $message);
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {
-            $this->get('old_sound_rabbit_mq.referent_message_dispatcher_producer')->scheduleDispatch($message);
+            $this->get('app.referent.message_notifier')->sendMessage($message);
```
HeahDude: Please keep the reference to avoid fetching the container twice.

src/Entity/ReferentManagedUsersMessage.php
HeahDude: Adressed considering your comment above.

app/config/services/platform.xml
```diff
@@ -354,6 +354,11 @@
             <argument type="service" id="router" />
         </service>
 
+        <service id="app.referent.message_notifier" class="AppBundle\Referent\ReferentMessageNotifier">
```
HeahDude: This service is public, you should use the FQCN as id.

src/Referent/ReferentMessageNotifier.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function __construct(
+        ObjectManager $manager,
+        ReferentMessageDispatcherProducerInterface $producer
+    ) {
+        $this->manager = $manager;
+        $this->producer = $producer;
+    }
+
+    public function createMessage(Adherent $referent, ManagedUsersFilter $filter)
```
HeahDude: This is not the responsibility of this class, you should add a static method in `ReferentMessage` instead.

src/Controller/EnMarche/ReferentController.php
```diff
@@ -65,13 +65,15 @@ public function usersSendMessageAction(Request $request): Response
             return $this->redirectToRoute('app_referent_users');
         }
 
-        $message = new ReferentMessage($this->getUser(), $filter);
```
HeahDude: So now you don't need this var anymore and can just get it only when the form is valid.

Remg: this `$message` is literally mapped to this form

HeahDude: Sorry I was targeting `$messageNotifier` in the other column of the diff view.

Remg: of course, good catch :+1:

Remg: fixed

PR #1719, opened 2017-10-31
Add role 'personality' on board member

NyKo24:


illusionOfParadise:
It's better to create a migration to add new role.

PR #1722, opened 2017-10-31
Change wording on IC form (front)

NyKo24:


src/CitizenInitiative/CitizenInitiativeCommand.php
```diff
@@ -276,4 +281,14 @@ public function setCapacity($capacity): void
 
         $this->capacity = $capacity;
     }
+
+    public function getPlace(): ?string
+    {
+        return $this->place;
+    }
+
+    public function setPlace(?string $place)
```
Remg: `public function setPlace(?string $place): void`

PR #1724, opened 2017-11-02
[1603] Fix registration path

blazarecki:


src/Controller/EnMarche/MembershipController.php
```diff
@@ -28,19 +29,28 @@ class MembershipController extends Controller
...
      * This action enables a guest user to adhere to the community.
      *
-     * @Route("/adhesion/{uuid}", name="app_register_fully")
+     * @Route("/adhesion", name="app_register_fully")
      * @Method("GET|POST")
+     *
+     * @Security("is_granted('ROLE_USER')")
      */
-    public function finaliseRegistrationAction(Request $request, Adherent $adherent): Response
+    public function finaliseRegistrationAction(Request $request): Response
```
HeahDude: not related to your PR, but typo: finalize (we use American English).

src/Entity/Adherent.php
```diff
@@ -516,6 +529,7 @@ public function updateMembership(MembershipRequest $membership, PostAddress $pos
         $this->phone = $membership->getPhone();
         $this->comEmail = $membership->comEmail;
         $this->comMobile = $membership->comMobile;
+        $this->adherent = true;
```
HeahDude: 🤔 So editing a profile is enough to be adherent?

templates/base.html.twig
```diff
@@ -6,7 +6,7 @@
 {% set default_sharer_width = 1905 %}
 {% set default_sharer_height = 1000 %}
 
-{% set user_is_adherent = is_granted('ROLE_ADHERENT') %}
+{% set user_is_user = is_granted('ROLE_USER') %}
```
HeahDude: `user_is_adherent`? the current naming looks weird.

HeahDude:
I'm not a fan of this new `adherent` profile property. What's the process now to be one or to stop being one?

templates/base.html.twig
blazarecki: Now you can be registered in en-marche without being an adherent. The naming is weird.

src/Entity/Adherent.php
blazarecki: No, finalize your registration in en-marche is enough

src/Controller/EnMarche/MembershipController.php
blazarecki: I'll rename it.

blazarecki:
You register in auth.en-marche.fr from en-marche.fr -> you became a `User`.

src/Entity/Adherent.php
HeahDude: So this should be removed, this makes it true for any edition (update) of the profile.

PR #1726, opened 2017-11-02
Add automatic deployment and migration

atamano:


scripts/deployment.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e
+
+# Get credentials
+sudo /opt/google-cloud-sdk/bin/gcloud --quiet components update kubectl
```
blazarecki: This is not needed, it's already done in the build script.

scripts/deployment.sh
```diff
@@ -0,0 +1,24 @@
...
+sudo /opt/google-cloud-sdk/bin/gcloud container clusters get-credentials $GCLOUD_CLUSTER --project $GCLOUD_PROJECT --zone europe-west1-d
+
+# Migrates database
+export GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcloud-service-key.json
+sudo /opt/google-cloud-sdk/bin/kubectl set image pod/staging-migrate-tasks enmarche=eu.gcr.io/$GCLOUD_PROJECT/app:$CIRCLE_SHA1
```
blazarecki: The `cache:clear --no-warmup` and `cache:warmup` of the entrypoint will be an issue for the migration.
The database schema will not be sync during the command, and the command will fail. 

I think we have 3 solutions:

1. We can run the new app image to only run the migration command migrate and deliberately don't run the cache:clear command. And then if migration is successful, update the app image.
2. We can update the entrypoint to run the migration command before the `cache:clear` command. (we also need to handle the rollback as well).
3. We can try to remove the `cache:clear` command of the entrypoint but it can be tricky because of the new app_host and amp_host env vars. 

What do you think of it ?

atamano: I thought the entrypoint is only applied when supervisord is beeing used.
https://github.com/EnMarche/en-marche.fr/blob/master/docker/prod/entrypoint.sh#L4

The pod `staging-migrate-tasks` I am using is simple kubernetes pod applying the command to migrate database `bin/console doctrine:migration:migrate -e prod --no-interaction`. It will live during migration and marked as `Completed` when finished. On next migration it will be replaced by a new pod with the new image

blazarecki: Fine ;) it'll work.

PR #1746, opened 2017-11-06
Fix https issue with HwiOauthBundle

blazarecki:


blazarecki:
The fix is in this PR https://github.com/EnMarche/en-marche.fr/pull/1844

PR #1771, opened 2017-11-13
[1597] Add adherent home page

ottaviano:


src/Controller/EnMarche/AdherentController.php
```diff
@@ -48,6 +48,15 @@ public function profileOverviewAction()
     }
 
     /**
+     * @Route("/accueil", name="app_adherent_home")
+     * @Method("GET")
+     */
+    public function homeAction()
```
illusionOfParadise: Missing return type hint.

ottaviano: @illusionOfParadise done

PR #1779, opened 2017-11-14
Add tags on adherent

NyKo24:


src/Admin/AdherentAdmin.php
```diff
@@ -383,6 +410,9 @@ protected function configureListFields(ListMapper $listMapper)
                 'label' => 'Type',
                 'template' => 'admin/adherent/list_status.html.twig',
             ])
+            ->add('tags', null, [
```
Remg: This could increase the number of queries very fast depending on the number of listed users.

Please consider overriding the query here like:
```
    public function createQuery($context = 'list')
    {
        $query = parent::createQuery($context);
        $query
            ->leftJoin($query->getRootAlias().'.tags', 't')
            ->addSelect('t')
        ;

        return $query;
    }
```
https://sonata-project.org/bundles/admin/master/doc/reference/action_list.html#customizing-the-query-used-to-generate-the-list

The same work should be done for the statuses actually, as it already generates a lot of additional queries.

Remg: :ok_hand:

PR #1780, opened 2017-11-15
replace our candidates by our deputy in the footer

guillaumedenhez:


templates/base.html.twig
```diff
@@ -428,7 +428,7 @@
                             </li>
                             <li>
                                 <a href="https://legislatives.en-marche.fr" target="_blank" rel="noopener noreferrer">
```
Remg: i know it's not related, but for legislatives routes we should now use the url generator:
`<a href="{{ path('legislatives_homepage') }}" target="_blank" rel="noopener noreferrer">`

PR #1786, opened 2017-11-16
add tag co-animateur on adherent list and filter ADMIN

NyKo24:


templates/admin/adherent/list_status.html.twig
```diff
@@ -18,7 +18,11 @@
             <span class="label label-primary" style="padding: 2px;">Procuration</span><br/>
         {% endif %}
 
-        {% if object.isHost %}
+        {% if object.isHostOnly %}
+            <span class="label label-primary" style="background-color: #90ade6 !important; padding: 2px;">Co-Animateur</span><br/>
```
thomasbisignani: What about to add a `padding: 2px` with a CSS selector for all span elements in this Twig block and apply the `background-color` only for this span in CSS ?

NyKo24: I keep the same logic for host & supervisor. Before, isHost check if adherent was Host (Co-AL) or supervisor (AL). Now we distinct with two different labels. I keep the same color and CSS Style ;)

PR #1789, opened 2017-11-18
Board members can not unregister

Remg:


src/DataFixtures/ORM/LoadAdherentData.php
```diff
@@ -81,6 +81,9 @@ public function load(ObjectManager $manager)
             'registered_at' => '2016-11-16 20:45:33',
         ]);
         $adherent2->disableCommitteesNotifications();
+        $roles = new ArrayCollection();
+        $roles->add($this->getReference('adherent'));
+        $adherent2->setBoardMember(BoardMember::AREA_ABROAD, $roles);
```
Remg: needed to have a board member who is neither a host nor a referent

PR #1794, opened 2017-11-20
Remove filter joins

Remg:


src/Admin/AdherentAdmin.php
```diff
@@ -32,9 +32,9 @@ class AdherentAdmin extends AbstractAdmin
         '_sort_by' => 'registeredAt',
     ];
 
-    public function createQuery($context = 'list')
```
Remg: `error_reporting | The $context argument of Sonata\AdminBundle\Admin\AbstractAdmin::createQuery is deprecated since 3.3, to be removed in 4.0.`

src/Admin/AdherentAdmin.php
```diff
@@ -363,7 +363,6 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                         return;
                     }
 
-                    $qb->leftJoin(sprintf('%s.memberships', $alias), 'ms');
                     $qb->andWhere(sprintf('ms.privilege', $alias).' = :privilege');
```
illusionOfParadise: After your modification the alias `ms` does not exist anymore.

src/Admin/AdherentAdmin.php
```diff
@@ -378,7 +377,6 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                         return;
                     }
 
-                    $qb->leftJoin(sprintf('%s.memberships', $alias), 'mss');
                     $qb->andWhere(sprintf('mss.privilege', $alias).' = :privilege');
```
illusionOfParadise: After your modification the alias `mss` does not exist anymore.

PR #1796, opened 2017-11-20
[1302] Delete account with auth

blazarecki:


src/Controller/Api/UserCanUnsubscribeController.php
```diff
@@ -0,0 +1,25 @@
...
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+class UserCanUnsubscribeController extends Controller
+{
+    /**
+     * @Route("/can_unsubscribe/{uuid}", name="api_can_unsubscribe")
```
ottaviano: you don't use this route?

blazarecki: Auth app use this route to check if the user can unsubscribe.

PR #1798, opened 2017-11-20
Refresh token roles on kernel.response event

ottaviano:


src/Security/RefreshTokenRolesListener.php
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        $token = $this->tokenStorage->getToken();
+
+        if (null === $token || !$token instanceof OAuthToken) {
+            return;
+        }
+
+        $user = $user = $token->getUser();
```
tony-tran: Duplicate `$user`

ottaviano: fixed 👍

PR #1802, opened 2017-11-21
Update mooc social image

marineb:


.gitignore
```diff
@@ -17,3 +17,4 @@
 /supervisord.log
 /supervisord.pid
 /yarn-error.log
+.DS_Store
```
Remg: this should be added in your global gitconfig.

PR #1803, opened 2017-11-21
UI improvements on profile edition form

ottaviano:


ottaviano:
I don't know how I could fix the issue of SLInsight :(
`address()` is the macro twig and I pass them one object `options` who could be transformed into JS object.

blazarecki:
@ottaviano it's not related to your PR. Don't fix it now.

PR #1828, opened 2017-11-29
[CP] Adminstrate citizen project categories

Remg:


app/config/security.yml
```diff
@@ -111,6 +111,7 @@ security:
             - ROLE_APP_ADMIN_CITIZEN_PROJECT_VIEW
             - ROLE_APP_ADMIN_CITIZEN_PROJECT_EDIT
             - ROLE_APP_ADMIN_CITIZEN_PROJECT_DELETE
+            - ROLE_APP_ADMIN_CITIZEN_PROJECT_CATEGORY_ALL
```
B-Galati: Why do we define a new role if it is not used ?

Remg: Sonata will dynamically use this role for the CitizenProjectCategoryAdmin

PR #1831, opened 2017-11-29
[CP] Administrate Citizen Projects

thomasbisignani:


src/Entity/CitizenProject.php
```diff
@@ -319,6 +320,12 @@ public function approved(string $timestamp = 'now'): void
     public function update(string $name, NullablePostAddress $address): void
     {
         $this->setName($name);
+        $this->setSubtitle($subtitle);
```
B-Galati: lots of variables are undefined here

src/Entity/CitizenProject.php
```diff
@@ -6,6 +6,7 @@
 use AppBundle\Exception\CitizenProjectAlreadyApprovedException;
 use Doctrine\ORM\Mapping as ORM;
 use libphonenumber\PhoneNumber;
+use Ramsey\Uuid\Uuid;
```
blazarecki: It's not used here.

src/Entity/CitizenProject.php
```diff
@@ -319,6 +320,12 @@ public function approved(string $timestamp = 'now'): void
     public function update(string $name, NullablePostAddress $address): void
     {
         $this->setName($name);
+        $this->setSubtitle($subtitle);
+        $this->category = $category;
+        $this->setAssistanceNeeded($assistanceNeeded);
```
blazarecki: Inside an Entity, try to avoid setter and use attribute instead

B-Galati: I disagree with you. Doing so would break the encapsulation principle.

blazarecki: You are inside the object, the encapsulation principle is already broken.

B-Galati: No, if one extends the behavior of the class by any way: inheritance, decorator, etc. This would break encapsulation. It's related to what it is explained in this article -> https://ocramius.github.io/blog/fluent-interfaces-are-evil/

src/Entity/CitizenProject.php
thomasbisignani: Thx, it's a mistake, I updated this method in another PR. :+1:

src/Entity/CitizenProject.php
```diff
@@ -316,9 +321,23 @@ public function approved(string $timestamp = 'now'): void
...
+    public function update(
+        string $name,
+        string $subtitle,
+        CitizenProjectCategory $category,
+        string $assistanceNeeded,
+        string $problemDescription,
+        string $proposedSolution,
+        string $requiredMeans,
+        NullablePostAddress $address): void
     {
```
Remg: NullablePostAddress $address
): void {

PR #1835, opened 2017-11-30
[CP] Add some columns and filters to citizen project list admin

B-Galati:


src/Admin/CitizenProjectAdmin.php
```diff
@@ -31,13 +35,15 @@ class CitizenProjectAdmin extends AbstractAdmin
     private $manager;
     private $citizenProjectMembershipRepository;
     private $cachedDatagrid;
+    private $citizenProjectRepository;
```
ottaviano: it's used?

src/Admin/CitizenProjectAdmin.php
```diff
@@ -277,9 +292,84 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                    }
+
+                    /** @var ProxyQuery|QueryBuilder $qb */
+                    if (!in_array('creator', $qb->getAllAliases(), true)) {
+                        $qb->leftJoin(Adherent::class, 'creator', Join::WITH, sprintf('creator.uuid=%s.createdBy', $alias));
+                    }
+
+                    $qb->andWhere(
+                            'LOWER(creator.emailAddress)=:emailAddress '.
+                            "OR $alias.createdBy=:uuid"
```
ottaviano: it's really necessary to search an email (`EmailType` field) in uuid column?

src/Admin/CitizenProjectAdmin.php
```diff
@@ -277,9 +292,84 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                        return;
+                    }
+
+                    /** @var ProxyQuery|QueryBuilder $qb */
+                    if (!in_array('creator', $qb->getAllAliases(), true)) {
+                        $qb->leftJoin(Adherent::class, 'creator', Join::WITH, sprintf('creator.uuid=%s.createdBy', $alias));
+                    }
+
+                    $qb->andWhere(
+                            'LOWER(creator.emailAddress)=:emailAddress '.
```
ottaviano: you can avoid `LOWER` function on email field and on `_ci` collate, it's already saved as lowercase string:
```
//MembershipRequest
public function setEmailAddress(string $emailAddress): void
    {
        $this->emailAddress = mb_strtolower($emailAddress);
    }
```

src/Admin/CitizenProjectAdmin.php
```diff
@@ -277,9 +292,84 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                            'LOWER(creator.firstName) LIKE :firstName '.
+                            "OR $alias.createdBy=:uuid"
+                        )
+                        ->setParameter('firstName', sprintf('%%%s%%', mb_strtolower($value['value'])))
+                        ->setParameter('uuid', $value['value']);
+
+                    return true;
+                },
+            ])
+            ->add('creatorLastName', CallbackFilter::class, [
```
ottaviano: maybe do juste one filter fied `Createur` for search in last and first name?

src/Admin/CitizenProjectAdmin.php
B-Galati: No thanks !

src/Admin/CitizenProjectAdmin.php
B-Galati: Yea great :-)

src/Admin/CitizenProjectAdmin.php
B-Galati: I think we must stick to the current layout with lastname and firstname in different filter

src/Admin/CitizenProjectAdmin.php
B-Galati: It's the way it is done and similar other field so I stick to it

ottaviano: why? 😕 
the input[type=email] field allows you to insert only a valid email address and it makes no sense to chech UUID column with an email.

B-Galati: I misunderstood what was done in the other fields.. all fixed now. Thank you very much :-)

PR #1837, opened 2017-12-03
Change referent message field types

Remg:


src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -62,17 +62,17 @@ class ReferentManagedUsersMessage
     private $includeSupevisors;
 
     /**
-     * @ORM\Column
+     * @ORM\Column(type="text")
```
HeahDude: Il n'y pas de contraintes quelque part ? (genre sur un VO)

Remg: good catch!
done: https://github.com/EnMarche/en-marche.fr/pull/1837/files#diff-0396f793d7ade106798f4f642e9c2e00

src/Referent/ManagedUsersFilter.php
```diff
@@ -23,14 +24,50 @@ class ManagedUsersFilter
     const PARAMETER_OFFSET = 'o';
     const PARAMETER_TOKEN = 't';
 
+    /**
+     * @Assert\Type("bool")
```
HeahDude: You can add an empty private constructor and remove those type assertions, since the good type is return from the VO line 77 and following.

PR #1840, opened 2017-12-04
[1779] Add AUTH and API image to EM

blazarecki:


src/Membership/AdherentFactory.php
```diff
@@ -65,6 +69,12 @@ public function createFromArray(array $data, bool $enabled = false): Adherent
             isset($data['isAdherent']) ? $data['isAdherent'] : true,
             isset($data['password']) ? $this->encodePassword($data['password']) : null
         );
+
+        if ($syncWithAuth) {
```
B-Galati: The Sync should be done after the entity is saved into DB no ?

blazarecki: It can be done before, this method is only used by fixtures.

PR #1841, opened 2017-12-04
[CP] Added Citizen projects skills data migration

thomasbisignani:


app/migrations/Version20171204142404.php
```diff
@@ -18,6 +22,13 @@ public function up(Schema $schema)
         $this->addSql('ALTER TABLE citizen_projects_skills ADD CONSTRAINT FK_B3D202D9EA64A9D0 FOREIGN KEY (citizen_project_skill_id) REFERENCES citizen_project_skills (id)');
     }
 
+    public function postUp(Schema $schema)
+    {
+        if ($this->container->get('app.migration.citizen_project_category_skill')) {
```
blazarecki: Please use the method `has` http://api.symfony.com/3.4/Symfony/Component/DependencyInjection/ContainerInterface.html#method_has

thomasbisignani: Oups! thx

app/migrations/Version20171204142404.php
```diff
@@ -18,6 +22,13 @@ public function up(Schema $schema)
         $this->addSql('ALTER TABLE citizen_projects_skills ADD CONSTRAINT FK_B3D202D9EA64A9D0 FOREIGN KEY (citizen_project_skill_id) REFERENCES citizen_project_skills (id)');
     }
 
+    public function postUp(Schema $schema)
+    {
+        if ($this->container->has('app.migration.citizen_project_category_skill')) {
```
B-Galati: It would be nice to add a little comment to explain why this test is here

src/Migration/CitizenProjectCategorySkillMigration.php
```diff
@@ -0,0 +1,129 @@
...
+                $citizenProjectCategorySkill = new CitizenProjectCategorySkill(
+                    $citizenProjectCategory, $citizenProjectSkill, $item['promotion']
+                );
+
+                $this->entityManager->persist($citizenProjectCategorySkill);
+            }
+        }
+    }
+
+    private static function categoryProvider(): array
```
HeahDude: Can't we reuse some fixtures constants like https://github.com/EnMarche/en-marche.fr/blob/master/src/DataFixtures/ORM/LoadCitizenInitiativeCategoryData.php#L11?

thomasbisignani: We can, but I wanted to separate the fixtures used only in local/dev and the migrations data for production.

HeahDude: The thing is that it adds all that in the cache anyway, so we could use prod values for dev fixtures, but this change would break some tests. It might not worth it, but if there aren't so many broken tests I suggest to do so.

thomasbisignani: I think it's a good idea for another features, in this case, this service will be removed as soon as the migration is run in production, and I think also this change would break somes tests.

PR #1842, opened 2017-12-05
[CP] Add citizen project search page

B-Galati:


app/config/services/platform.xml
```diff
@@ -427,11 +427,33 @@
             <argument type="service" id="cache.app"/>
         </service>
 
-        <service id="app.search.search_results_provider" class="AppBundle\Search\SearchResultsProvider">
+        <service id="app.search.search_results_providers_manager" class="AppBundle\Search\SearchResultsProvidersManager">
+            <call method="addProvider">
```
HeahDude: After an upgrade to 3.4 you could use:
```
<call method="addProviders">
    <argument type="tagged" tag="app.search_provider" />
</call>
```
And auto tag providers below thanks to their interface if any.

src/Controller/EnMarche/SearchController.php
```diff
@@ -32,7 +31,7 @@ public function searchEventsAction(Request $request)
         }
 
         try {
-            $results = $this->get('app.search.search_results_provider')->find($search);
+            $results = $this->get('app.search.search_results_providers_manager')->find($search);
```
HeahDude: Should use `Fqcn::class` here.

src/Controller/EnMarche/SearchController.php
```diff
@@ -76,11 +75,27 @@ public function searchCommitteesAction(Request $request)
...
      */
-    public function searchAction(Request $request): Response
+    public function searchCitizenProjectsAction(Request $request)
     {
         $this->disableInProduction();
 
-        return new Response();
+        $request->query->set(SearchParametersFilter::PARAMETER_TYPE, SearchParametersFilter::TYPE_CITIZEN_PROJECTS);
+
+        $search = $this->getSearch($request);
```
HeahDude: Did not look deeply into it, but having this line https://github.com/EnMarche/en-marche.fr/pull/1842/files#diff-77e84c6db7fe37094654486a56e642fbR126 apart from `handleRequest` seems weird, and if it could be moved inside the service, we could get rid of this private function. (Not really in your scope I know).

src/Controller/EnMarche/SearchController.php
```diff
@@ -76,11 +75,27 @@ public function searchCommitteesAction(Request $request)
...
+
+        try {
+            $results = $this->get('app.search.search_results_providers_manager')->find($search);
+        } catch (GeocodingException $exception) {
+            $errors[] = $this->get('translator')->trans('search.geocoding.exception');
+        }
+
+        return $this->render('search/search_citizen_projects.html.twig', [
+            'search_max_results' => $this->getParameter('search_max_results'),
```
HeahDude: :/, and here's another reason why this should be part of the search.

src/Controller/EnMarche/SearchController.php
```diff
@@ -76,11 +75,27 @@ public function searchCommitteesAction(Request $request)
...
+            'search_type' => SearchParametersFilter::TYPE_CITIZEN_PROJECTS,
```
HeahDude: The type is already part of the search why passing it twice to the view?

src/Repository/CitizenProjectRepository.php
```diff
@@ -63,4 +64,37 @@ public function findOneApprovedBySlug(string $slug): ?CitizenProject
...
+                ->createNearbyQueryBuilder($coordinates)
+                ->andWhere($this->getNearbyExpression().' < :distance_max')
+                ->setParameter('distance_max', $search->getRadius());
+        } else {
+            $qb = $this->createQueryBuilder('n');
+        }
+
+        if (!empty($query = $search->getQuery())) {
+            $qb->andWhere('n.name like :query');
+            $qb->setParameter('query', '%'.$query.'%');
```
HeahDude: `'%'.$query.'%'` => `"%$query%"`

src/Repository/CitizenProjectRepository.php
```diff
@@ -63,4 +64,37 @@ public function findOneApprovedBySlug(string $slug): ?CitizenProject
...
+        }
+
+        return $qb
+            ->andWhere('n.status = :status')
+            ->setParameter('status', CitizenProject::APPROVED)
+            ->setFirstResult($search->getOffset())
+            ->setMaxResults($search->getMaxResults())
+            ->getQuery()
+            ->getResult()
+            ;
```
HeahDude: CS: Wrong indent.

src/Search/CitizenProjectSearchResultsProvider.php
```diff
@@ -0,0 +1,34 @@
...
+        $this->citizenProjectRepository = $citizenProjectRepository;
+        $this->citizenProjectManager = $citizenProjectManager;
+    }
+
+    public function find(SearchParametersFilter $search): array
+    {
+        $result = $this->citizenProjectRepository->searchAll($search);
+
+        foreach ($result as $citizenProject) {
+            $citizenProject->setCreator($this->citizenProjectManager->getCitizenProjectCreator($citizenProject));
```
HeahDude: This is fu.... weird. I know you've already asked before, but why is this needed in the first place. We should maybe revise this model.

B-Galati: yes... I spent 1 or 2 hours looking for a decent solution but I was not able to find one reasonable..

app/config/services/platform.xml
B-Galati: I would love that 👍

src/Controller/EnMarche/SearchController.php
B-Galati: You are right

src/Controller/EnMarche/SearchController.php
B-Galati: Agree

src/Search/CitizenProjectSearchResultsProvider.php
HeahDude: Damned...

B-Galati:
all done @HeahDude 
Thank you !

src/Search/CitizenProjectSearchResultsProvider.php
B-Galati: Is there any way to change the default query performed by the EntityManager ?

B-Galati: Because I find something cool by mapping the creator field into Doctrine but it requires to use a specific DQL each time we need to retrieve the entity.

src/Controller/EnMarche/SearchController.php
```diff
@@ -123,8 +117,8 @@ public function resultsAction(Request $request)
     private function getSearch(Request $request): SearchParametersFilter
     {
         return $this
-            ->get('app.search.search_results_filter')
-            ->setMaxResults($this->getParameter('search_max_results'))
-            ->handleRequest($request);
+            ->get(\AppBundle\Search\SearchParametersFilter::class)
```
HeahDude: Please inline this call in controllers and get rid of this private method.

src/Search/CitizenProjectSearchResultsProvider.php
HeahDude: Can this DQL injection be abstracted in some base repository class or a trait?

B-Galati: For sure it could be

B-Galati:
@HeahDude Done 👍

app/config/services/platform.xml
```diff
@@ -422,16 +422,42 @@
         </service>
 
         <!-- Search -->
-        <service id="app.search.search_results_filter" class="AppBundle\Search\SearchParametersFilter">
+        <service id="AppBundle\Search\SearchParametersFilter" shared="false">
```
HeahDude: I fail to see when do we need to have it instantiated many times?

B-Galati: Yes normal, I did like this because, in theory, each object is like an entity with its own values.

B-Galati: It depends how you use these 2 instances. At the very beginning yes but then their values can  change independently.

HeahDude: So you're saying we have some stageful services?

B-Galati: What do you mean ?

HeahDude: I mean services should be stateless (sans état), and their configuration should not change during runtime to avoid undesired side effects.

>It depends how you use these 2 instances. At the very beginning yes but then their values can change independently.

I still got two questions:
 - when does it happens to handle 2 instances in the same runtime?
 - how can their values changes during runtime?

B-Galati: So yea it's stateful because the object is meant to be changed at runtime.

- When does it happens to handle 2 instances in the same runtime?
Never at the moment. Like I said, I did that from a theoretical point of view

- how can their values change during runtime?
At the moment, the value changed based on how you use setter methods and the HTTP request (see `handleRequest`).

HeahDude: I'm sorry I cannot see any reason to perform 2 search queries with the same Request. Calling the setter for maxResults does not seem necessary we could inject it in the constructor and use the default constant if no value is passed. The setter should only be used with a from type mapping attributes from the request.

Do I miss something?

HeahDude: Yet another question, do we really need a parameter hardcoded in the config file AND a default class constant, both using the same value... ?

B-Galati: > I'm sorry I cannot see any reason to perform 2 search queries with the same Request. Calling the setter for maxResults does not seem necessary we could inject it in the constructor and use the default constant if no value is passed. The setter should only be used with a from type mapping attributes from the request.

I know and I agree but I cannot rewrite the whole feature in this PR.

> Yet another question, do we really need a parameter hardcoded in the config file AND a default class constant, both using the same value... ?

Good question. I don't know why it was done like this. I guess it's to provide more flexibility just in case. Right now it looks useless.

HeahDude: Ok sorry again because I review an already implemented feature, and understand this is out of your scope. Now this is good to have this conversation on record.

Finally I suggest to use a shared service anyway.

B-Galati: You are very welcome ! Thank you for your review 👍

src/Search/SearchResultsProvidersManager.php
```diff
@@ -0,0 +1,36 @@
...
+        ));
+    }
+
+    /**
+     * @throws \RuntimeException If one provider already supports the type of search
+     */
+    public function addProvider(SearchResultsProviderInterface $provider): void
+    {
+        if (isset($this->providers[$provider->getSupportedTypeOfSearch()])) {
+            throw new \RuntimeException('This type of search is already supported by another provider');
```
HeahDude: To help debugging you should add the classes of the provider passed and of the one already set.

src/Admin/CitizenProjectDatagrid.php
```diff
@@ -26,13 +26,11 @@ public function __construct(DatagridInterface $decorated, CitizenProjectManager
     public function getResults()
     {
         if (!$this->cachedResults) {
-            /** @var CitizenProject[] $results */
```
HeahDude: Why moving the phpdoc below after the initialization? Maybe you can just remove it.

B-Galati: I don't know why I did that hahaha. In this context it is necessary to help the IDE.

src/Search/SearchResultsProvidersManager.php
B-Galati: done

PR #1845, opened 2017-12-05
CP add autocomplete skills on CitizenProject form

NyKo24:


src/CitizenProject/CitizenProjectCommand.php
```diff
@@ -185,4 +192,23 @@ public function setCitizenProject(CitizenProject $citizenProject): void
...
+    }
+
+    public function setSkills(iterable $skills): void
+    {
+        foreach ($skills as $skill) {
+            $this->addSkill($skill);
+        }
+    }
+
+    public function addSkill(CitizenProjectSkill $skill)
```
Remg: missing return type

templates/javascript.js.twig
```diff
@@ -102,3 +102,110 @@ $(document).ready(function() {
     removeSkill();
 });
 {% endmacro %}
+
+{% macro skillsProjectCitizen(idSkillsListField, idSkillSearchField ,urlSearch) %}
```
thomasbisignani: What about to keep the `CitizenProjectSkills` naming syntax in front ?

templates/javascript.js.twig
```diff
@@ -102,3 +102,110 @@ $(document).ready(function() {
...
+            $('.summary-skill > input[type=hidden]').each(function () {
+                skillCount = Math.max(this.id.replace(/[^0-9]/g, ''), skillCount);
+            });
+            skillCount++;
+
+            var newSkillWidget = skillList.attr('data-prototype'),
+                skillWidgetId = '#'+skillList.attr('id')+'_'+skillCount;
+            newSkillWidget = newSkillWidget.replace(/__name__/g, skillCount);
+
+            console.log(skillWidgetId);
```
thomasbisignani: Can you delete this line ? ;)

templates/javascript.js.twig
```diff
@@ -102,3 +102,110 @@ $(document).ready(function() {
...
+        var btn = $('.skill-remove');
+
+        btn.unbind('click');
+        btn.click(function () {
+            $(this).parent().remove();
+        });
+    };
+
+    var addSkill = function (item) {
+        console.log(item);
```
thomasbisignani: Can you delete this line ?

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -45,4 +49,35 @@ public function helpAction(): Response
 
         return new Response();
     }
+
+    /**
+     * @Route("/skills/autocompletion",
+     *     name="app_citizen_project_skills_autocomplete",
+     *     condition="request.isXmlHttpRequest()"
```
HeahDude: Please add `and request.query.get("category") and request.query.get("term")` to prevent useless queries, the javascript itself should not trigger an Ajax request with empty params.

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -45,4 +49,35 @@ public function helpAction(): Response
...
+        if (!$category = $this->getDoctrine()->getRepository(CitizenProjectCategory::class)->find($request->query->get('category'))) {
+            return new JsonResponse([], Response::HTTP_NOT_FOUND);
+        }
+
+        $citizenProjectCategorySkills = $this
+            ->getDoctrine()
+            ->getRepository(CitizenProjectCategorySkill::class)
+            ->findByCitizenProjectCategoryAndTerm($category, $request->query->get('term', ''));
+
+        $result = [];
```
HeahDude: `$results` and no need for the initialization use `return $results ?? [];` (this also compiles better optimized C code).

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -45,4 +49,35 @@ public function helpAction(): Response
...
+        /** @var CitizenProjectCategorySkill $citizenProjectCategorySkill */
```
HeahDude: There should be a `CitizenProjectCategorySkill[]` php doc for the return of the repository method instead.

src/Entity/CitizenProject.php
```diff
@@ -350,7 +350,8 @@ public function update(
         string $problemDescription,
         string $proposedSolution,
         string $requiredMeans,
-        NullablePostAddress $address
+        NullablePostAddress $address,
+        Collection $skills
```
HeahDude: I would use `iterable` here.

src/Entity/CitizenProject.php
```diff
@@ -385,4 +387,30 @@ public function getCreator(): ?Adherent
...
+
+    public function getSkills(): Collection
+    {
+        return $this->skills;
+    }
+
+    public function setSkills(iterable $skills): void
+    {
+        foreach ($skills as $skill) {
+            if ($skill instanceof CitizenProjectSkill) {
```
HeahDude: I'm not sure this check is needed. And if it is, it would be better to throw an `\InvalidArgument` when this is not true.

src/Entity/CitizenProject.php
```diff
@@ -385,4 +387,30 @@ public function getCreator(): ?Adherent
     {
         return $this->creator;
     }
+
+    public function getSkills(): Collection
```
HeahDude: It's ok (even better) to keep phpdocs which bring some actual value like `CitizenProjectSkill[]|Collection`.

src/CitizenProject/CitizenProjectCommand.php
```diff
@@ -185,4 +192,23 @@ public function setCitizenProject(CitizenProject $citizenProject): void
     {
         $this->citizenProject = $citizenProject;
     }
+
+    public function getSkills(): Collection
+    {
+        return $this->skills;
+    }
+
+    public function setSkills(iterable $skills): void
```
B-Galati: IMO, `set` means "ok let's start over with this data". Here it's not the case. It's an `addSkills` method.
So either rename it or modify it to match the semantic of the method.

src/CitizenProject/CitizenProjectUpdateCommandHandler.php
```diff
@@ -29,6 +29,27 @@ public function handle(CitizenProjectCommand $command)
...
+            $citizenProjectSkillIdsToSubmit[] = $skill->getId();
+        }
+
+        foreach ($citizenProject->getSkills() as $skill) {
+            $citizenProjectSkillIdsAlreadyAssociate[] = $skill->getId();
+        }
+
+        $citizenProjectSkillIdsToRemove = array_diff($citizenProjectSkillIdsAlreadyAssociate, $citizenProjectSkillIdsToSubmit);
+
+        foreach ($citizenProjectSkillIdsToRemove as $id) {
```
B-Galati: It looks cumbersome. Why not just replacing the whole skills collection with a new one ? Any idea @HeahDude ?

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -45,4 +49,34 @@ public function helpAction(): Response
 
         return new Response();
     }
+
+    /**
+     * @Route("/skills/autocompletion",
+     *     name="app_citizen_project_skills_autocomplete",
+     *     condition="request.isXmlHttpRequest() and request.query.get("category") and request.query.get("term")"
```
B-Galati: It's gonna do a 404 if we missed `category` and `term` parameters from the query string ? If yes it would be better to answer with a 400

src/Entity/CitizenProject.php
```diff
@@ -360,6 +361,7 @@ public function update(
         $this->setProblemDescription($problemDescription);
         $this->setProposedSolution($proposedSolution);
         $this->setRequiredMeans($requiredMeans);
+        $this->setSkills($skills->toArray());
```
B-Galati: The call to `toArray()` is useless.

templates/citizen_project/edit.html.twig
```diff
@@ -6,11 +6,14 @@
...
     <script src="{{ asset('ckeditor/ckeditor.js') }}"></script>
     <script src="{{ asset('ckeditor/lang/fr.js') }}"></script>
+    <script type="text/javascript" src={{ asset('bundles/sonataadmin/vendor/jquery/dist/jquery.min.js') }}></script>
+    <script type="text/javascript" src={{ asset('bundles/sonataadmin/vendor/jqueryui/ui/jquery-ui.js') }}></script>
     <script type="text/javascript">
         Kernel.onLoad(function() {
             {{ js.address(form.address) }}
             {{ js.ckeditor('citizen_project_proposed_solution') }}
             App.runCitizenProjectCreation();
+            {{ js.citizenProjectSkills('#citizen_project_skills', '#citizen_project_skills_search', path('app_citizen_project_skills_autocomplete')) }}
```
B-Galati: I don't understand why we need a macro for this.
We could achieve the same behavior with a Javascript file.

templates/javascript.js.twig
```diff
@@ -102,3 +102,107 @@ $(document).ready(function() {
     removeSkill();
 });
 {% endmacro %}
+
+{% macro citizenProjectSkills(idSkillsListField, idSkillSearchField ,urlSearch) %}
+$(document).ready(function() {
+    var skillList = $('{{ idSkillsListField|raw }}');
```
B-Galati: It's better to use `e('js')` instead of `raw`.

templates/javascript.js.twig
```diff
@@ -102,3 +102,107 @@ $(document).ready(function() {
     removeSkill();
 });
 {% endmacro %}
+
+{% macro citizenProjectSkills(idSkillsListField, idSkillSearchField ,urlSearch) %}
+$(document).ready(function() {
```
B-Galati: I don't think it is needed if the javascript is already at the end of the document no ?

tests/CitizenProject/CitizenProjectFactoryTest.php
```diff
@@ -65,5 +68,6 @@ public function testCreateCitizenProjectFromCitizenProjectCreationCommand()
         $this->assertSame($problemDescription, $citizenProject->getProblemDescription());
         $this->assertSame($proposedSolution, $citizenProject->getProposedSolution());
         $this->assertSame($requiredMeans, $citizenProject->getRequiredMeans());
+        $this->assertCount(1, $citizenProject->getSkills());
```
B-Galati: Would be better with something like `assertSame([$skill], $citizenProject->getSkills()->toArray())`;

src/CitizenProject/CitizenProjectUpdateCommandHandler.php
HeahDude: No I never really understood this part, I guess this is some business constraint I'm missing. I definitely on "cumbersome". I don't get why the CollectionType does not do this job.

src/Controller/EnMarche/CitizenProjectController.php
HeahDude: Yeah I'm pretty cool with this since this is internal api call that should not happen in the first place unless we want a full list to start but then it should be part of the first request unless the list is lazy loaded because the filed is optional. Anyway of to add some `if` in the controller as long as we don't trigger services for nothing.

src/CitizenProject/CitizenProjectCommand.php
NyKo24: If I remove all elements in $this->skills before call for each, it's good ?

templates/citizen_project/edit.html.twig
NyKo24: because in history other autocomplete JS functionality was wrote in macro, I follow the history...

tests/CitizenProject/CitizenProjectFactoryTest.php
NyKo24: ok, but why ? I would understand...

src/CitizenProject/CitizenProjectCommand.php
B-Galati: Yea I think so. You can also create a brand new Collection then add the one provided.

src/Controller/EnMarche/CitizenProjectController.php
B-Galati: Agree thank you ;-)

templates/citizen_project/edit.html.twig
B-Galati: I would say that it's never too late to change our habits :P. 
Perhaps it was done for a very good reason in the 1st place ? For this case I don't see any reason why we should do like this.
Anyway I think we can keep it like this but it would be cool considering not to use macro by default in the future.

tests/CitizenProject/CitizenProjectFactoryTest.php
B-Galati: Because here it is just the size of the returned value that is tested.
`assertSame` is more precise. It would strictly compare the type of collection and the data of the collection itself.
It's not a big deal but it's better 😉

src/CitizenProject/CitizenProjectUpdateCommandHandler.php
B-Galati: Thank you for the clarification !

src/Entity/CitizenProject.php
B-Galati: Please fix this one

src/Entity/CitizenProject.php
```diff
@@ -385,4 +387,34 @@ public function getCreator(): ?Adherent
...
+     * @return CitizenProjectSkill[]|Collection
+     */
+    public function getSkills(): Collection
+    {
+        return $this->skills;
+    }
+
+    public function setSkills(iterable $skills): void
+    {
+        foreach ($skills as $skill) {
```
B-Galati: The collection should be cleared at the beginning of the method

templates/javascript.js.twig
B-Galati: Please fix this one too and other call to `raw`

templates/javascript.js.twig
```diff
@@ -102,3 +102,107 @@ $(document).ready(function() {
...
+
+    var input = $('{{ idSkillSearchField|raw }}');
+
+    // Add skill autocomplete
+    input.autocomplete({
+        source: function(request, response) {
+            var term = request.term.trim(),
+                categoryId = $('#citizen_project_category').val();
+            if (term.length > 0 && categoryId) {
+                $.get("{{ urlSearch|raw }}?term=" + term+'&category='+categoryId, function(data) {
```
B-Galati: We could save up some HTTP calls by wrapping the whole function in a `setTimeout` function.
See https://github.com/EnMarche/en-marche.fr/pull/1842/files#diff-abdc72dfc05d1f015d1308f8b7bae8cbR50

templates/javascript.js.twig
NyKo24: No it's needed

src/Repository/CitizenProjectCategorySkillRepository.php
```diff
@@ -2,8 +2,21 @@
 
 namespace AppBundle\Repository;
 
+use AppBundle\Entity\CitizenProjectCategory;
 use Doctrine\ORM\EntityRepository;
 
 class CitizenProjectCategorySkillRepository extends EntityRepository
 {
+    public function findByCitizenProjectCategoryAndTerm(CitizenProjectCategory $category, string $term)
```
illusionOfParadise: Missing return type

src/Form/DataTransformer/CitizenProjectSkillTransformer.php
```diff
@@ -0,0 +1,49 @@
...
+            ];
+        }
+
+        return $newArray ?? [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function reverseTransform($ids)
```
illusionOfParadise: Missing return type

src/Form/DataTransformer/CitizenProjectSkillTransformer.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function __construct(CitizenProjectSkillRepository $citizenProjectSkillRepository)
+    {
+        $this->citizenProjectSkillRepository = $citizenProjectSkillRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function transform($arrayProjectCitizenSkill)
```
illusionOfParadise: Missing return type

src/Repository/CitizenProjectCategorySkillRepository.php
NyKo24: ✔️

src/Form/DataTransformer/CitizenProjectSkillTransformer.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        foreach ($arrayProjectCitizenSkill as $value) {
+            /* @var $value CitizenProjectSkill */
```
HeahDude: `/* @var CitizenProjectSkill $value */` before the foreach.

src/Repository/CitizenProjectCategorySkillRepository.php
```diff
@@ -2,8 +2,21 @@
...
+    {
+        return $this
+            ->createQueryBuilder('cpcs')
+            ->leftJoin('cpcs.skill', 's')
+            ->where('cpcs.category = :cpc')
+            ->andWhere('LOWER(s.name) LIKE :term')
+            ->setParameter('cpc', $category)
+            ->setParameter('term', '%'.strtolower($term).'%')
+            ->getQuery()
+            ->getResult();
```
HeahDude: Semi colon on a new line.

PR #1847, opened 2017-12-06
[CP] Add phone field to CP form

illusionOfParadise:


src/Entity/CitizenProject.php
```diff
@@ -364,6 +350,10 @@ public function update(
         if (null === $this->postAddress || !$this->postAddress->equals($address)) {
             $this->postAddress = $address;
         }
+
+        if (!$this->phone->equals($phone)) {
```
B-Galati: What is the usefulness of this ?

HeahDude: Persisting a clone while letting an orphan behind.

src/Admin/CitizenProjectAdmin.php
```diff
@@ -111,6 +111,12 @@ protected function configureShowFields(ShowMapper $showMapper)
                     'label' => 'Longitude',
                 ])
             ->end()
+            ->with('Phone', ['class' => 'col-md-5'])
```
NyKo24: Téléphone ?

illusionOfParadise: Changed

PR #1848, opened 2017-12-06
[CP] Send email to committee request support when CP is approve

NyKo24:


src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -77,4 +77,51 @@ public function committeeAutocompleteAction(Request $request)
 
         return new JsonResponse($result ?? []);
     }
+
+    /**
+     * @Route("/mon-comite-soutien/{slug}", name="app_citizen_project_committee_support")
+     * @Method("GET")
```
B-Galati: This action should be `POST`. 
There is also a security issue and a CSRF token should be used to fix it.

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -77,4 +77,51 @@ public function committeeAutocompleteAction(Request $request)
...
+    public function committeeSupportAction(CitizenProject $citizenProject): Response
+    {
+        if (!$this->getUser()->isSupervisor()) {
+            throw $this->createAccessDeniedException();
+        }
+
+        $committeeManager = $this->get('app.committee.manager');
+        $flashMessage = sprintf('Le projet citoyen %s n\'a pas demandé votre soutient', $citizenProject->getName());
+
+        foreach ($citizenProject->getCommitteeSupports() as $committeeSupport) {
```
B-Galati: Too much work performed in this controller. Try to move the business logic to the entity itself and/or manager/voter depending on the context.

templates/citizen_project/_sidebar.html.twig
```diff
@@ -64,14 +64,14 @@
...
-                    </li>
-                {% endif %}
+            {% for committeeSupport in citizen_project.approvedCommitteeSupports %}
+                <li class="text--body">
+                    {{ committeeSupport.committee }}
+                </li>
             {% endfor %}
         </ul>
-        <a href="#">Notre comité souhaite aussi soutenir ce projet</a>
+        {% if app.user and is_granted('ROLE_SUPERVISOR') %}
```
B-Galati: I think testing `if app.user` is implicit when using `is_granted`

src/Controller/EnMarche/CitizenProjectController.php
NyKo24: this URL is call on email send to committee supervisor

templates/citizen_project/_sidebar.html.twig
NyKo24: 👍

src/Controller/EnMarche/CitizenProjectController.php
B-Galati: Ok but we still need to fix the security issue.
I see 2 solutions : 
    - Create an intermediary page that would ask a confirmation to the user with a form (the simplest and best solution to me).
    - Use any kind of token (JWT, hmac time limited, a one time random secret, or whatever). Just like we do with password reset request.

src/CitizenProject/CitizenProjectMessageNotifier.php
```diff
@@ -65,6 +72,19 @@ private function scheduleCreationNotification(CitizenProject $citizenProject): v
         ]));
     }
 
+    private function sendAskCommitteeSupport(CitizenProject $citizenProject, Adherent $creator): void
+    {
+        foreach ($citizenProject->getPendingCommitteeSupports() as $committeeSupport) {
+            $this->mailer->sendMessage(
+                CitizenProjectRequestCommitteeSupportMessage::create(
+                    $citizenProject,
+                    $creator,
```
HeahDude: Passing the creator all along is not needed, just call `$citizenProject->getCreator()` in that `create` method.

src/Controller/EnMarche/CitizenProjectController.php
HeahDude: 👍 to send a temporary token, as done with account creation confirmation.

src/Entity/CitizenProjectCommitteeSupport.php
```diff
@@ -88,7 +88,7 @@ public function isPending(): bool
         return self::PENDING === $this->status;
     }
 
-    public function isApprove(): bool
+    public function isApproved(): bool
     {
         return self::APPROVE === $this->status;
```
illusionOfParadise: It's better to name a constant as `APPROVED`

src/Entity/CitizenProjectCommitteeSupport.php
```diff
@@ -98,4 +98,9 @@ public function approve(string $timestamp = 'now'): void
         $this->status = self::APPROVE;
         $this->approvedAt = new \DateTime($timestamp);
     }
+
+    public function getCitizenProject()
```
illusionOfParadise: Missing type hint

PR #1850, opened 2017-12-07
[CP] Comments section on CP show page

B-Galati:


src/Entity/CitizenProjectComment.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(type="text", nullable=true)
+     */
+    private $content;
+
+    /**
+     * @var \DateTime
```
HeahDude: `@var \DateTimeInterface`

src/Entity/CitizenProjectComment.php
```diff
@@ -0,0 +1,106 @@
...
+
+    public function __construct(
+        ?UuidInterface $uuid,
+        CitizenProject $citizenProject,
+        Adherent $author,
+        string $content,
+        string $createdAt = 'now'
+    ) {
+        if (empty($content)) {
+            throw new \InvalidArgumentException('Comment content cannot be empty');
```
HeahDude: The column is nullable 🤔

src/Entity/CitizenProjectComment.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->content = $content;
+        $this->createdAt = new \DateTimeImmutable($createdAt);
+    }
+
+    public function getCitizenProject(): CitizenProject
+    {
+        return $this->citizenProject;
+    }
+
+    public function getAuthor(): ?Adherent
```
HeahDude: What about an `AuthoredTrait` for this 3 methods?

src/Entity/CitizenProjectComment.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function getContent(): string
+    {
+        return $this->content;
+    }
+
+    public function getCreatedAt(): \DateTimeInterface
+    {
+        return $this->createdAt;
```
HeahDude:
```php
if ($this->createdAt instanceof \DateTime) {
    $this->createdAt = \DateTimeImmutable::createFromMutable($this->createdAt);
}

return $this->createdAt;
```

src/Entity/CitizenProjectComment.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $this->author = null;
+    }
+
+    public function getContent(): string
+    {
+        return $this->content;
+    }
+
+    public function getCreatedAt(): \DateTimeInterface
```
HeahDude: `\DateTimeImmutable`

src/CitizenProject/CitizenProjectCommentCommand.php
```diff
@@ -0,0 +1,57 @@
...
+    private $author;
+
+    /**
+     * @var string|null
+     *
+     * @Assert\NotBlank()
+     */
+    private $content;
+
+    public function __construct(CitizenProject $citizenProject, Adherent $author)
```
HeahDude: `?Adherent` because it might be null on editing.

src/CitizenProject/Voter/CommentCitizenProjectVoter.php
```diff
@@ -0,0 +1,21 @@
...
+
+class CommentCitizenProjectVoter extends AbstractCitizenProjectVoter
+{
+    protected function supports($attribute, $citizenProject)
+    {
+        return in_array($attribute, [CitizenProjectPermissions::COMMENT, CitizenProjectPermissions::SHOW_COMMENT], true)
+               && $citizenProject instanceof CitizenProject;
+    }
+
+    protected function doVoteOnAttribute(string $attribute, Adherent $adherent, CitizenProject $citizenProject): bool
```
HeahDude: /!\ I'm not sure you can add type hint on an abstract method signature.

src/Repository/CitizenProjectCommentRepository.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace AppBundle\Repository;
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\CitizenProject;
+use Doctrine\ORM\EntityRepository;
+
+class CitizenProjectCommentRepository extends EntityRepository
+{
+    public function findCitizenProjectComments(CitizenProject $citizenProject): array
```
HeahDude: I would call it `findForProject`, we already know we will find ...comments thanks to the ...CommentRepository.

Here you can add useful phpdoc like `@return ...Comment[]` to avoid duck type hinting in the code.

src/Repository/CitizenProjectCommentRepository.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return $this->findBy(['citizenProject' => $citizenProject], ['createdAt' => 'DESC']);
+    }
+
+    public function findCitizenProjectCommentsByAuthor(Adherent $author): array
```
HeahDude: `findForAuthor`.

src/Repository/CitizenProjectCommentRepository.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return $this->findBy(['author' => $author]);
+    }
+
+    public function removeAuthorItems(Adherent $author)
```
HeahDude: `removeForAuthor`

src/Twig/CitizenProjectRuntime.php
```diff
@@ -58,4 +58,14 @@ public function canSeeCitizenProject(CitizenProject $citizenProject): bool
     {
         return $this->authorizationChecker->isGranted(CitizenProjectPermissions::SHOW, $citizenProject);
     }
+
```
HeahDude: Not related to your PR but is it valuable to have a nullable argument in the constructor?

src/CitizenProject/CitizenProjectCommentCommand.php
B-Galati: Editing is not something we want yet.
But anyway, the author is mandatory at object creation only, then it can be set as Anonymous and edited without calling the constructor (Thank you Doctrine !)

src/CitizenProject/Voter/CommentCitizenProjectVoter.php
B-Galati: The method signature is not changed. Check the base class ;-)

src/Entity/CitizenProjectComment.php
B-Galati: Good catch thank you !

src/Repository/CitizenProjectCommentRepository.php
B-Galati: Agree

src/Twig/CitizenProjectRuntime.php
B-Galati: I don't think so.

src/CitizenProject/CitizenProjectCommentCommand.php
HeahDude: This is not the entity but the command.

src/CitizenProject/Voter/CommentCitizenProjectVoter.php
HeahDude: Ah yes! Sorry!

src/CitizenProject/CitizenProjectCommentCommand.php
B-Galati: Good point my bad !
Edit is not feature yet so we can let it go, no ?

src/CitizenProject/CitizenProjectCommentCommand.php
```diff
@@ -0,0 +1,57 @@
...
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\CitizenProject;
+use Symfony\Component\Validator\Constraints as Assert;
+
+class CitizenProjectCommentCommand
+{
+    /**
+     * @var CitizenProject
+     *
+     * @Assert\NotNull()
```
HeahDude: No need for empty parentheses (same below).

src/CitizenProject/CitizenProjectCommentCommand.php
HeahDude: Please do it now, it's not that easy to catch this business constraint.

src/CitizenProject/CitizenProjectCommentCommand.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->citizenProject = $citizenProject;
+        $this->author = $author;
+    }
+
+    public function getCitizenProject(): CitizenProject
+    {
+        return $this->citizenProject;
+    }
+
+    public function getAuthor(): Adherent
```
HeahDude: Nullable here too though.

B-Galati:
@HeahDude Done !

src/Twig/CitizenProjectRuntime.php
B-Galati: fixed

PR #1852, opened 2017-12-08
Handled citizen action creation

HeahDude:


app/config/services/platform.xml
```diff
@@ -226,6 +226,10 @@
         </service>
 
         <!-- Citizen projects -->
+        <service id="app.citizen_project.manager" class="AppBundle\CitizenProject\CitizenProjectManager">
```
thomasbisignani: This service is needed ? you can use action injection (enabled in `services.xml`)

thomasbisignani: I think this service is not conditionally used. He's used in the `return` statement : `$citizenProjectManager->getCitizenProjectAdministrators($project)`

HeahDude: Yes you're right. Good catch, thanks!

src/Controller/EnMarche/CitizenActionManagerController.php
```diff
@@ -0,0 +1,55 @@
...
+ * @Entity("action", expr="repository.findOneBySlug(slug)")
+ */
+class CitizenActionManagerController extends Controller
+{
+    /**
+     * @Route("/creer", name="app_citizen_action_manager_create", defaults={"slug": ""})
+     * @Method("GET|POST")
+     * @Security("is_granted('CREATE_CITIZEN_ACTION', project)")
+     */
+    public function createAction(Request $request, CitizenProject $project, CitizenProjectManager $citizenProjectManager): Response
```
Remg: argument `CitizenProjectManager $citizenProjectManager` shouldn't be there

src/Controller/EnMarche/CitizenActionManagerController.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * @Route("/projets-citoyens/{project_slug}/actions")
+ * @Entity("project", expr="repository.findOneApprovedBySlug(project_slug)")
+ * @Entity("action", expr="repository.findOneBySlug(slug)")
+ */
+class CitizenActionManagerController extends Controller
+{
+    /**
+     * @Route("/creer", name="app_citizen_action_manager_create", defaults={"slug": ""})
```
Remg: remove the default `slug` parameter

Remg: remove default `slug` parameter

app/config/services/platform.xml
```diff
@@ -368,6 +369,12 @@
             <argument type="service" id="app.repository.citizen_action"/>
         </service>
 
+        <service autowire="true" id="AppBundle\CitizenAction\CitizenActionCommandHandler">
+            <argument key="$mailer" type="service" id="app.mailer.transactional" />
```
B-Galati: Why not creating an alias for `app.mailer.transactional` and then use autowiring ?

src/Entity/CitizenAction.php
```diff
@@ -52,9 +44,9 @@ public function __construct(
         $this->category = $category;
         $this->description = $description;
         $this->postAddress = $address;
-        $this->capacity = $capacity;
         $this->participantsCount = $participantsCount;
-        $this->beginAt = $beginAt;
+        // We need a \DateTime object for now to work with Gedmo sluggable
+        $this->beginAt = $beginAt instanceof \DateTimeImmutable ? new \DateTime($beginAt->format(\DATE_ATOM)) : $beginAt;
```
B-Galati: Just Use https://secure.php.net/manual/en/datetimeimmutable.createfrommutable.php instead of `new \DateTime`

B-Galati: For the record, I misundertood it's ok :-)

PR #1856, opened 2017-12-11
[CP] Change wording on form & add checkbox

NyKo24:


src/Form/CitizenProjectCommandType.php
```diff
@@ -94,6 +94,17 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
         $citizenProject = $builder->getData()->getCitizenProject();
 
+        if (!$citizenProject instanceof CitizenProject) {
```
thomasbisignani: Can you place this logic in the listener on the `preSetData`? I think you can also place the `committees` and` committees_search` fields logic.

src/Form/CitizenProjectCommandType.php
```diff
@@ -143,5 +143,16 @@ public function preSetData(FormEvent $event)
                 'disabled' => true,
             ]);
         }
+
+        if (!$citizenProject instanceof CitizenProject) {
```
B-Galati: What do you think if we change this condition to 
```
if (!$citizenProject) {
```

or 

```
if (null === $citizenProject) {
```

app/config/config.yml
```diff
@@ -71,6 +71,7 @@ twig:
         legacy_adhesion:        "https://adherents.en-marche.fr/adherer.html"
         legislatives_platform:  "%env(LEGISLATIVES_PLATFORM)%"
         algolia_blacklist:      "%env(ALGOLIA_BLACKLIST)%"
+        url_help_en_marche:     "https://aide.en-marche.fr"
```
B-Galati: This URL should be in parameters.yml no ?
Any thought @Remg ?

src/Form/CitizenProjectCommandType.php
NyKo24: @HeahDude prefere instanceof... but for me I don't have preference...

app/config/config.yml
NyKo24: I hope no, because it's an external service who doesn't provide a staging env

src/Form/CitizenProjectCommandType.php
B-Galati: Ok as you wish guys, not a big deal.
I just thought it would be little bit more efficient like this.

app/config/config.yml
B-Galati: ok fine !

src/Form/CitizenProjectCommandType.php
HeahDude: >@HeahDude prefere instanceof... 

Not in all cases...

In fact I don't get why there is a pre set data listener, the condition is already there in `buildForm`. Please merge the logic in one place by removing this listener. 

The call here https://github.com/EnMarche/en-marche.fr/pull/1856/files#diff-76eb4209f0717100613e396a1c32ebc2R95 might break since `$builder->getData()` is no tested. Please add before something like:
```php
$command = $builder->getData();

if (!$commmand instanceof CitizenProjectCommand) {
    throw new InvalidConfigurationException('A pre set data is required in '.__CLASS__);
}
```

PR #1858, opened 2017-12-11
Use php-cs-fixer instead of styleci

B-Galati:


B-Galati:
@atamano done thank you !

PR #1866, opened 2017-12-12
Add email when adherent register on citizen action

NyKo24:


app/config/services/platform.xml
```diff
@@ -225,7 +225,7 @@
             <tag name="security.voter"/>
         </service>
 
-        <!-- Citizen projects -->
+        <!-- Citizen projectss -->
```
thomasbisignani: Can you fix this typo on this branch?

src/Event/EventRegistrationCommandHandler.php
```diff
@@ -40,11 +42,17 @@ public function handle(EventRegistrationCommand $command): void
...
-        $eventLink = $this->generateUrl('app_event_show', [
-            'slug' => $command->getEvent()->getSlug(),
-        ]);
-
-        $this->mailer->sendMessage(EventRegistrationConfirmationMessage::createFromRegistration($registration, $eventLink));
+        if ($command->getEvent() instanceof CitizenAction) {
+            $calendarExportLink = $this->generateUrl('app_event_export_ical', [
+                'slug' => $command->getEvent()->getSlug(),
+            ]);
+            $this->mailer->sendMessage(CitizenActionRegistrationConfirmationMessage::createFromRegistration($registration, $calendarExportLink));
```
B-Galati: `sendMessage` can be done outside after the if statement

templates/citizen_action/show.html.twig
```diff
@@ -0,0 +1,19 @@
+{% extends 'events/show.html.twig' %}
+
+{% block page_title citizen_action.name %}
+{% block canonical_url url('app_citizen_action_show', { slug: citizen_action.slug }) %}
+
+{% block json_ld %}
+    {% set event = citizen_action %}
```
B-Galati: I am not sure. Is it possible to set the `event` variable outside of block ? Like we could avoid the same line 3 times.

templates/citizen_initiative/show.html.twig
```diff
@@ -1,7 +1,7 @@
 {% extends 'events/show.html.twig' %}
 
 {% block page_title initiative.name %}
-{% block canonical_url url('app_citizen_initiative_show', {'slug': initiative.slug}) %}
+{% block canonical_url url('app_citizen_initiative_show', { slug: initiative.slug }) %}
```
B-Galati: cs

thomasbisignani: Can you remove this whitespaces ?

PR #1873, opened 2017-12-14
add AC search

NyKo24:


src/Repository/EventRepository.php
```diff
@@ -339,15 +341,21 @@ public function searchAllEvents(SearchParametersFilter $search): array
             $filterQuery = '';
         }
 
-        if ($category = $search->getEventCategory()) {
+        if ($category = $search->getEventCategory() && SearchParametersFilter::TYPE_CITIZEN_ACTIONS != $search->getEventCategory()) {
```
Remg: Please use strict comparison here

PR #1876, opened 2017-12-15
add follow / unfollow CP

NyKo24:


src/CitizenProject/CitizenProjectManagementAuthority.php
```diff
@@ -39,4 +40,18 @@ public function preApprove(CitizenProject $project): void
     {
         $this->manager->preApproveCitizenProject($project);
     }
+
+    public function followCitizenProject(Adherent $adherent, CitizenProject $citizenProject): void
+    {
+        $this->manager->followCitizenProject($adherent, $citizenProject);
+
+        if (!$hosts = $this->manager->getCitizenProjectAdministrators($citizenProject)->toArray()) {
```
B-Galati: I think this test should be moved in the listener(s) that don't want to do any kind of processing in this case. Because we will create new listeners in the future and perhaps the gonna want the event even if the CP is orphan.

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -199,4 +199,48 @@ public function listActorsAction(CitizenProject $citizenProject, CitizenProjectM
...
+    {
+        if (!$this->isCsrfTokenValid('citizen_project.follow', $request->request->get('token'))) {
+            throw $this->createAccessDeniedException('Invalid CSRF protection token to follow citizen project.');
+        }
+
+        $this->get('app.citizen_project.authority')->followCitizenProject($this->getUser(), $citizenProject);
+
+        return new JsonResponse([
+            'button' => [
+                'label' => 'Quitter ce projet citoyen',
```
B-Galati: Sounds weird to return view data from the action but perhaps it's done like this on this project.

src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -199,4 +199,48 @@ public function listActorsAction(CitizenProject $citizenProject, CitizenProjectM
...
+                'action' => 'quitter',
+                'csrf_token' => (string) $this->get('security.csrf.token_manager')->getToken('citizen_project.unfollow'),
```
B-Galati: As an alternative we could use the same CSRF for follow and unfollow.

tests/Controller/EnMarche/CitizenProjectControllerTest.php
```diff
@@ -200,6 +202,82 @@ public function testCommitteeSupportCitizenProject()
...
+
+        $crawler = $this->client->request(Request::METHOD_GET, $citizenProjectUrl);
+
+        $this->assertResponseStatusCode(Response::HTTP_OK, $this->client->getResponse());
+        $this->assertSame('2 acteurs', $crawler->filter('#followers > h3')->text());
+        $this->assertTrue($this->seeFollowLink($crawler));
+        $this->assertFalse($this->seeUnfollowLink($crawler));
+        $this->assertFalse($this->seeRegisterLink($crawler, 0));
+
+        // Emulate POST request to follow the committee.
```
B-Galati: Perhaps we should consider using a client that can interpret javascript one day.

src/CitizenProject/CitizenProjectManagementAuthority.php
NyKo24: ✔️

src/Controller/EnMarche/CitizenProjectController.php
NyKo24: yes, see the same logical on committee follow / unfollow

src/Controller/EnMarche/CitizenProjectController.php
NyKo24: I use the same logical on committee follow / unfollow and I prefer two differents tokens

tests/Controller/EnMarche/CitizenProjectControllerTest.php
NyKo24: YES !!! behat and selenium ❤️

PR #1881, opened 2017-12-16
[CP] add button to create citizen action

NyKo24:


src/CitizenAction/Voter/CreateCitizenActionVoter.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function voteOnAttribute($attribute, $subject, TokenInterface $token)
+    {
+        return
+            $token->getUser() instanceof Adherent
+            && $subject instanceof CitizenProject
```
thomasbisignani: This test is it usefull ? IMHO, it's already checked by the `supports` method.

PR #1887, opened 2017-12-18
fix voter service definition for citizen action

NyKo24:


app/config/services/services.xml
```diff
@@ -23,5 +23,7 @@
 		<service id="AppBundle\CitizenProject\Voter\AdministrateCitizenProjectVoter">
 			<tag name="security.voter" priority="-1000"/>
 		</service>
+
+		<prototype autowire="true" autoconfigure="true" namespace="AppBundle\CitizenAction\Voter\" resource="../../../src/CitizenAction/Voter/" public="false" />
```
thomasbisignani: The defaults options are already defined at line 7 :
```<defaults autowire="true" autoconfigure="true" public="false" />```

thomasbisignani:
:ok_hand:

PR #1910, opened 2017-12-21
[CP] Fix CP show page

illusionOfParadise:


src/Controller/EnMarche/CitizenProjectController.php
```diff
@@ -53,25 +53,11 @@ public function showIfNotAuthorizedDepartementAction(): Response
     public function showAction(Request $request, CitizenProject $citizenProject, CitizenProjectManager $citizenProjectManager): Response
```
thomasbisignani: The `$request` variable and the `POST` method annotation are useful ?

illusionOfParadise: No

PR #1922, opened 2017-12-26
[CP] Added committees in the CP edit form without status logic

thomasbisignani:


src/Form/CitizenProjectCommandType.php
```diff
@@ -117,30 +133,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-                    'filter_emojis' => true,
-                    'attr' => [
-                        'placeholder' => 'Ajouter le comité local',
-                    ],
-                ]);
-
-            $builder->get('committees')->addModelTransformer($this->committeeTransformer);
-        }
-
         if ($command instanceof CitizenProjectCommand && $command->isCitizenProjectApproved()) {
```
HeahDude: Not related but this if can be simplified to `if ($command->isCitizenProjectApproved()) {` due to the condition above.

thomasbisignani: Thanks you @HeahDude, good catch !

PR #1927, opened 2017-12-27
[CP] Updated the image upload handler

thomasbisignani:


src/CitizenProject/CitizenProjectManager.php
```diff
@@ -498,14 +498,51 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
         return $this->getCitizenProjectRepository()->hasCitizenProjectInStatus($adherent, $status);
     }
 
+    /**
+     * Uploads and saves the citizen project image to the cloud storage.
```
B-Galati: At this point, it does not matter if it's cloud or not, that's why we have an abstraction.

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -498,14 +498,51 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
...
     }
+
+    /**
+     * Removes the citizen project image from the cloud storage.
+     *
+     * @param CitizenProject $citizenProject
+     */
+    public function removeImage(CitizenProject $citizenProject): void
+    {
+        if ($citizenProject->isRemoveImage()) {
```
B-Galati: It does not feel like it's the responsibility of the manager to test this. IMO, the entity should not contain this information at all.

src/Form/CitizenProjectCommandType.php
```diff
@@ -133,6 +138,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder->get('skills')->addModelTransformer($this->citizenProjectSkillTransformer);
         $builder->get('committees')->addModelTransformer($this->committeeTransformer);
         $builder->get('address')->remove('address');
+
+        // Removes the "remove_image" field on the creation form or if the image was not upload
+        if (!$citizenProject instanceof CitizenProject || !$citizenProject->hasImageUploaded()) {
```
B-Galati: Let's invert the logic and only add the field if the CP has an image, what do you think ?

templates/citizen_project/_form.html.twig
```diff
@@ -43,6 +43,12 @@
         <div class="citizen_project_form__picture_preview">
             <img style="max-width: 400px" src="{{ asset(citizen_project.getAssetImagePath) }}" />
         </div>
+        <div class="form__new form__new__checkbox">
+            {{ form_widget(form.remove_image) }}
+            <label for="{{ form.remove_image.vars.id }}">
```
B-Galati: Why not just use `form_label` ?

app/migrations/Version20171227110138.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20171227110138 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE citizen_projects ADD image_name LONGTEXT DEFAULT NULL');
```
B-Galati: Is it really needed to have a `LONGTEXT` field for this one ? `VARCHAR` should be enough.

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -498,14 +498,51 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
...
+
+            // Clears the cache file
+            $this->glide->deleteCache($path);
+
             $citizenProject->setImageUploaded(true);
         }
     }
+
+    /**
+     * Removes the citizen project image from the cloud storage.
```
B-Galati: same

src/CitizenProject/CitizenProjectManager.php
B-Galati: I suggest to let the command handler manage if this method needs to be called or not.
An exception could be thrown if one try to remove an image on an entity that does not contain an image.

src/CitizenProject/CitizenProjectUpdateCommandHandler.php
```diff
@@ -48,13 +49,28 @@ public function handle(CitizenProjectCommand $command)
             $command->phone,
             $command->getSkills(),
             $command->getCommittees(),
-            $command->getImage()
+            $command->getImage(),
+            $command->isRemoveImage()
```
B-Galati: It's not needed to store this information in the entity

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -498,14 +498,51 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
...
-            $this->storage->put($pathImage, file_get_contents($citizenProject->getImage()->getPathname()));
-            $this->glide->deleteCache($pathImage);
+            $path = $citizenProject->getImagePath();
+
+            // Clears the old image if needed
+            if (null !== $citizenProject->getImageName() && $oldImagePath = $citizenProject->getImagePath()) {
+                $this->storage->delete($oldImagePath);
+            }
+
+            $citizenProject->setImageName($citizenProject->getImage());
```
HeahDude: Looks weird.

src/Form/CitizenProjectCommandType.php
```diff
@@ -31,6 +32,7 @@ public function __construct(CitizenProjectSkillTransformer $citizenProjectSkillT
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $command = $builder->getData();
+        $citizenProject = $command->getCitizenProject();
```
HeahDude: You cannot call this before https://github.com/EnMarche/en-marche.fr/pull/1927/files#diff-76eb4209f0717100613e396a1c32ebc2R37.

src/CitizenProject/CitizenProjectManager.php
HeahDude: The test is already there https://github.com/EnMarche/en-marche.fr/pull/1927/files#diff-ae8b6be9941e0be84d8476baa425862fR72, just throw a runtime exception.

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -499,50 +498,49 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
...
-            if (null !== $citizenProject->getImageName() && $oldImagePath = $citizenProject->getImagePath()) {
-                $this->storage->delete($oldImagePath);
-            }
+        // Clears the old image if needed
+        if (null !== $citizenProject->getImageName() && $oldImagePath = $citizenProject->getImagePath()) {
+            $this->storage->delete($oldImagePath);
+        }
 
-            $citizenProject->setImageName($citizenProject->getImage());
+        $citizenProject->setImageName($citizenProject->getImage());
```
HeahDude: You removed the check that ensure that `$citizenProject->getImage()` is an instance of `UploadedFile`, is this intended?

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -499,50 +498,49 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
...
+     * Removes the citizen project image.
      *
      * @param CitizenProject $citizenProject
      */
     public function removeImage(CitizenProject $citizenProject): void
     {
-        if ($citizenProject->isRemoveImage()) {
-            $path = $citizenProject->getImagePath();
+        if (null === $citizenProject->getImageName()) {
+            throw new \RuntimeException('This Citizen Project not contains an image.');
```
HeahDude: `This Citizen Project does not contain an image.`

src/Entity/CitizenProject.php
```diff
@@ -153,15 +153,10 @@ class CitizenProject extends BaseGroup implements CoordinatorAreaInterface
     /**
      * @var string|null
      *
-     * @ORM\Column(type="text", nullable=true)
+     * @ORM\Column(length=255, nullable=true)
```
HeahDude: To avoid a 500 you should trim the name when needed.

thomasbisignani: The `imageName` is hashed in the setter. Do you think is really needed to trim it ? The name will always be the same size (with the extension .jpg or .png).

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -498,14 +498,54 @@ public function hasCitizenProjectInStatus(Adherent $adherent, array $status): bo
...
     public function addImage(CitizenProject $citizenProject): void
     {
-        // Save citizen project image to cloud storage
-        if ($citizenProject->getImage() instanceof UploadedFile) {
-            $pathImage = $citizenProject->getImagePath();
-            $this->storage->put($pathImage, file_get_contents($citizenProject->getImage()->getPathname()));
-            $this->glide->deleteCache($pathImage);
-            $citizenProject->setImageUploaded(true);
+        if (!$citizenProject->getImage() instanceof UploadedFile) {
+            throw new \RuntimeException(sprintf('The image must be an instance of %s', UploadedFile::class));
```
HeahDude: Can't it be null? What about submitting a form without editing or removing the image, in such case null should be allowed, right?

thomasbisignani: If the image is not submitted, the `CitizenProjectCreationCommandHandler` and the `CitizenProjectUpdateCommandHandler`  not call this `addImage` method : 
```        // Uploads an image
        if (null !== $command->getImage()) {
            $this->citizenProjectManager->addImage($citizenProject);
        } 
```

HeahDude: Fait enough

PR #1928, opened 2017-12-27
Added an Election model

HeahDude:


src/Entity/Election.php
```diff
@@ -0,0 +1,106 @@
...
+    private $name = '';
+
+    /**
+     * @var string|null
+     *
+     * @ORM\Column(nullable=true)
+     *
+     * @Assert\Length(max=255)
+     */
+    private $place;
```
HeahDude: FYI I'm not sure yet of the usefulness of this property.

app/migrations/Version20171227161442.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20171227161442 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('CREATE TABLE elections (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(255) NOT NULL, place VARCHAR(255) DEFAULT NULL, UNIQUE INDEX UNIQ_1BD26F335E237E06 (name), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
+        $this->addSql('CREATE TABLE election_rounds (id INT AUTO_INCREMENT NOT NULL, election_id INT DEFAULT NULL, label VARCHAR(255) NOT NULL, description VARCHAR(255) DEFAULT NULL, date DATE NOT NULL, INDEX IDX_37C02EA0A708DAFF (election_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
```
B-Galati: election_id is not nullable no ?

src/Entity/ElectionRound.php
```diff
@@ -0,0 +1,117 @@
...
+ * @UniqueEntity({"election", "label"})
+ */
+class ElectionRound
+{
+    /**
+     * @var int|null
+     *
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue(strategy="IDENTITY")
```
B-Galati: Don't we prefer a UUID instead ?

src/Entity/Election.php
```diff
@@ -0,0 +1,106 @@
...
+class Election
+{
+    /**
+     * @var int|null
+     *
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue(strategy="IDENTITY")
+     */
+    private $id;
```
B-Galati: UUID ?

app/migrations/Version20171227161442.php
HeahDude: `nullable` is not an option for ManyToOne AFAIK. Do you know a work around?

src/Entity/ElectionRound.php
HeahDude: Not for now. This is a first draft, many PR will compose this feature and the model may change during the process.

app/migrations/Version20171227161442.php
HeahDude: Ok I didn't know this was settable on the joinColumn config.

HeahDude: Fixed.

src/Entity/ElectionRound.php
B-Galati: I am not convinced and don't really understand why it's not done now but ok :D

HeahDude: UUID has a performance cost and is an "extra" id because we always keep an internal primary ke. We should use it only when exposing resources via url to avoid leaking the "real" id.

I'm not convinced for now that we should add it on these entities.

B-Galati: Unrelated to this PR but what about UUID only instead of having both ? It looks like the guys from Doctrine does not like auto generated ids :D

HeahDude: I guess because performances with UUID only are worse. The choice was originally made by @tgalopin and @hhamon, maybe they can enlighten us more?

tgalopin: @B-Galati you should never rely on UUID for foreign keys, they are using much more memory in indexes. On the other hand, you should also never rely on internal integers for public pages because the can be counted so they enable easy scraping. To get the best from both worlds, you can have both an ID and a UUId: the UUID for WHERE queries and the ID for joins.

src/Entity/Election.php
tgalopin: Nope, see my other answer

src/Entity/ElectionRound.php
tgalopin: Note that in this specific PR, I don't see any usage of a WHERE query for elections, so adding an UUID does not seem necessary, I agree with @HeahDude

src/Entity/Election.php
```diff
@@ -0,0 +1,106 @@
...
+
+    public function addRound(ElectionRound $round): void
+    {
+        if (!$this->rounds->contains($round)) {
+            $this->rounds->add($round);
+            $round->setElection($this);
+        }
+    }
+
+    public function removeRound($round)
```
tgalopin: `: void`

src/Entity/ElectionRound.php
```diff
@@ -0,0 +1,118 @@
...
+     * @Assert\Length(max=255)
+     */
+    private $label = '';
+
+    /**
+     * @var string|null
+     *
+     * @ORM\Column(nullable=true)
+     *
+     * @Assert\Length(max=255)
```
tgalopin: Text perhaps?

HeahDude: Good call, I forward to Marine :p. Thanks!

src/Entity/ElectionRound.php
B-Galati: Thanks @tgalopin ;-)

PR #1931, opened 2017-12-28
Fix error during removing CitizenProjectSkill

NyKo24:


src/Entity/CitizenProjectSkill.php
```diff
@@ -19,8 +21,16 @@
  */
 class CitizenProjectSkill extends BaseSkill
 {
+    /**
+     * @var Collection|CitizenProjectCategorySkill
+     *
+     * @ORM\OneToMany(targetEntity="AppBundle\Entity\CitizenProjectCategorySkill", mappedBy="skill", orphanRemoval=true)
```
HeahDude: Are you making this relation bi-directional just to add the `orphanRemoval` directive?

If so you should create a migration instead, because this is only managed by doctrine, a delete query on the db will have the same problem:
```php
class CitizenProjectCategorySkill
{
    /**
     * ...
     * @ORM\JoinColumn(onDelete="CASCADE")
     */
    private $skill;

// ...
```
Will handle the case using doctrine or the db with the same behavior.

src/Entity/CitizenProjectSkill.php
```diff
@@ -19,8 +21,16 @@
  */
 class CitizenProjectSkill extends BaseSkill
 {
+    /**
+     * @var Collection|CitizenProjectCategorySkill
```
illusionOfParadise: Missing brackets : `CitizenProjectCategorySkill` -> `CitizenProjectCategorySkill[]`

tests/CitizenProject/CitizenProjectMessageNotifierTest.php
```diff
@@ -30,11 +30,6 @@ public function testOnCitizenProjectApprove()
         $committeeManager = $this->createMock(CommitteeManager::class);
         $router = $this->createMock(RouterInterface::class);
 
-        $producer->expects($this->once())->method('publish')->with(\GuzzleHttp\json_encode([
```
HeahDude: `$producer->expects($this->never())`?

B-Galati:
Rebase needed ? It looks like there are 2 extra commits no ?

HeahDude:
To rebase again.

PR #1932, opened 2017-12-28
Add remove committee support button

NyKo24:


src/CitizenProject/CitizenProjectManager.php
```diff
@@ -481,9 +482,22 @@ public function approveCommitteeSupport(Committee $committee, CitizenProject $ci
         }
 
         $committeeSupport->approve();
+        $this->getManager()->persist($committeeSupport);
```
HeahDude: There always a cascade persist set, you should call `$this->citizenProject->addCommitteeSupport($this)` in the support constructor instead.

src/Entity/CitizenProject.php
```diff
@@ -256,6 +256,17 @@ public function getPhone(): ?PhoneNumber
         return $this->phone;
     }
 
+    public function isSupportByCommitteeUuid(string $committeeUuid): bool
```
HeahDude: `isSupportedByCommitteeUuid`

src/Exception/CitizenProjectCommitteeSupportNotFoundException.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace AppBundle\Exception;
+
+use AppBundle\Entity\Committee;
+use Throwable;
+
+class CitizenProjectCommitteeSupportNotFoundException extends CitizenProjectCommitteeSupportException
```
HeahDude: Not sure this exception has a true value, what about throwing an `\InvalidArgumentException` or a `\RuntimeException` from the manager?

NyKo24: I create this exception to precise exception type and catch it more simply, I use this exception in `deleteCommitteeSupport` on Manager

B-Galati:
Rebase needed ?

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -483,7 +484,19 @@ public function approveCommitteeSupport(Committee $committee, CitizenProject $ci
...
         if ($flush) {
-            $this->getManager()->persist($committeeSupport);
+            $this->getManager()->flush();
+        }
+    }
+
+    public function deleteCommitteeSupport(Committee $committee, bool $flush = true): void
+    {
+        if (!$committeeSupport = $this->getCitizenProjectCommitteeSupportRepository()->findByCommittee($committee)) {
+            throw new CitizenProjectCommitteeSupportNotFoundException($committee);
```
HeahDude: >I create this exception to precise exception type and catch it more simply, I use this exception in deleteCommitteeSupport on Manager

Yes my point is that there is no specific catch for it, and this could be a more generic exception, I don't see the value of using a domain one here for now.

PR #1933, opened 2017-12-28
[REFERENT] Display label AL/Co-AL in referent managed users list

illusionOfParadise:


templates/referent/users_list.html.twig
```diff
@@ -188,7 +188,11 @@
                                 </small>
                                 <br />
                                 <small>
-                                    {{ user.isCommitteeSupervisor ? 'Animateur local' : user.isCommitteeHost ? 'Co-animateur local' : '' }}
+                                    {%- if user.isCommitteeSupervisor -%}
+                                        Animateur local
+                                        {{- user.isCommitteeHost ? ' / ' : '' -}}
```
HeahDude: I would declare a var before the if: `{% set is_host = user.isCommitteeHost %}`.

illusionOfParadise: Done

PR #1935, opened 2017-12-28
Add CKEditor on proposedSolution field on citizen project edit form

NyKo24:


src/Admin/CitizenProjectAdmin.php
```diff
@@ -154,6 +154,8 @@ protected function configureFormFields(FormMapper $formMapper)
                 ])
                 ->add('proposedSolution', null, [
                     'label' => 'Solution du problème',
+                    'filter_emojis' => true,
+                    'attr' => ['class' => 'ck-editor'],
```
HeahDude: missing `'purify_html' => true`?

B-Galati:
Rebase needed ?

src/Admin/CitizenProjectAdmin.php
NyKo24: it's a WISIWIG not purify HTML it's needed no ?

HeahDude: Never trust the client.

PR #1940, opened 2017-12-28
Add creation static page feature

tony-tran:


app/migrations/Version20171227163801.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20171227163801 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('UPDATE pages SET slug=\'emmanuel-macron\' WHERE slug= \'emmanuel-macron-ce-que-je-suis\'');
```
HeahDude: You must also handle 301s. See https://github.com/EnMarche/en-marche.fr/blob/281f460f0f5425c27787642d0bd0426381006120/src/Redirection/LegacyRedirectionsSubscriber.php#L35.

tony-tran: I don't need to do redirection, I change the slug for the `findOneBySlug`. Look changes on `PageController.php`

B-Galati: Why did you change all the slugs in that case ? I think I am missing something..

app/config/routing.yml
```diff
@@ -96,3 +96,12 @@ app_redirecting:
     host:     "%app_host%"
     resource: "@AppBundle/Controller/RedirectingController.php"
     type:     "annotation"
+
+# This route must be loaded the last to prevent an override
+app_static_page:
+    host:    "%app_host%"
+    path:    /s/{slug}
```
B-Galati: We really want this weird path ? Looks bad for SEO no ?

app/migrations/Version20171227163801.php
B-Galati: It is to simplify the logic inside `SitemapFactory.php`, that's it ?

tony-tran: I change all slugs because they are all wrong in database. And I need to change slug to generate the right URL in `SitemapFactory.php`.

app/config/routing.yml
tony-tran: There is already a route with `/{url}` for path. To prevent some errors I prefer use this url and it is write in the ticket of this feature ;)

app/migrations/Version20171227163801.php
HeahDude: I don't understand your answer, the Uris did change right?

HeahDude: I mean a changed uri should now respond a 404?

HeahDude: If "yes" is the answer to my questions above, you MUST handle 301.

B-Galati: No URLs are still the same ;-)

HeahDude: Ok, got it, thx!

PR #1945, opened 2017-12-29
Don’t show termine membership for non adherent

blazarecki:


HeahDude:
A test for this would be good.

PR #1946, opened 2017-12-29
Constraints must match DB ones

B-Galati:


src/Entity/EntityContentTrait.php
```diff
@@ -55,6 +57,7 @@
      * @var string|null
      *
      * @ORM\Column(length=255, nullable=true)
+     * @Assert\Length(max=255)
```
HeahDude: Please add an empty line between mapping and assertions.

B-Galati: done

PR #1947, opened 2017-12-29
Test sitemaps

tony-tran:


tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $sitemap = new Sitemap();
+        $this->createReflectionMethod('addArticles', $sitemap);
+
+        $this->assertEquals(1, $sitemap->getUrlCount());
+    }
+
+    public function testCreateContentSitemapWithoutHit()
+    {
+        $cacheItemInterface = $this->createMock(CacheItemInterface::class);
+        $cacheItemInterface->method('isHit')->willReturn(true);
```
HeahDude: CS: please go use a new line for each chained calls.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $categoriesRepository = $this->createMock(ArticleCategoryRepository::class);
+        $categoriesRepository->method('findAll')->willReturn([$this->category]);
+
+        $articleRepository = $this->createMock(ArticleRepository::class);
+        $articleRepository->method('findAllPublished')->willReturn([$this->article]);
+
+        $orderArticleRepository = $this->createMock(OrderArticleRepository::class);
+        $orderArticleRepository->method('findAllPublished')->willReturn([]);
+
+        $this->objectManager->method('getRepository')->will($this->returnValueMap([
```
HeahDude: Same.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->article->setSlug('article');
+        $this->article->setCategory($this->category);
+        $this->article->setUpdatedAt(new \DateTime());
+
+        $this->objectManager = $this->createMock(ObjectManager::class);
+        $this->routerInterface = $this->createMock(RouterInterface::class);
+        $this->cacheItemPoolInterface = $this->createMock(CacheItemPoolInterface::class);
+
+        $categoriesRepository = $this->createMock(ArticleCategoryRepository::class);
+        $categoriesRepository->method('findAll')->willReturn([$this->category]);
```
HeahDude: `->expects($this->any())` as a minimum, best would be a fixed count when possible.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+        $articleRepository = $this->createMock(ArticleRepository::class);
+        $articleRepository->method('findAllPublished')->willReturn([$this->article]);
```
HeahDude: Same.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+        $orderArticleRepository = $this->createMock(OrderArticleRepository::class);
+        $orderArticleRepository->method('findAllPublished')->willReturn([]);
```
HeahDude: Same.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->objectManager = null;
+        $this->routerInterface = null;
+        $this->cacheItemPoolInterface = null;
+        $this->article = null;
+        $this->category = null;
+
+        parent::tearDown();
+    }
+
+    private function createReflectionMethod(string $methodName, Sitemap $sitemap = null)
```
HeahDude: `invokeReflectionMethod`

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+class SitemapFactoryTest extends TestCase
+{
+    private $objectManager;
+    private $routerInterface;
+    private $cacheItemPoolInterface;
+    private $article;
+    private $category;
+
+    public function testAddArticle()
```
B-Galati: missing return type on some methods of this class

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->article = null;
+        $this->category = null;
+
+        parent::tearDown();
+    }
+
+    private function createReflectionMethod(string $methodName, Sitemap $sitemap = null)
+    {
+        $reflection = new \ReflectionMethod(SitemapFactory::class, $methodName);
+        $reflection->setAccessible(true);
```
B-Galati: This feels really hacky to me, unit tests should not care about private methods.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->category->setSlug('category');
+
+        $this->article = new Article();
+        $this->article->setSlug('article');
+        $this->article->setCategory($this->category);
+        $this->article->setUpdatedAt(new \DateTime());
+
+        $this->objectManager = $this->createMock(ObjectManager::class);
+        $this->routerInterface = $this->createMock(RouterInterface::class);
+        $this->cacheItemPoolInterface = $this->createMock(CacheItemPoolInterface::class);
```
B-Galati: You could use a simple implementation of the `CacheItemPoolInterface`. It would make our life easier.

tests/Sitemap/SitemapFactoryTest.php
blazarecki: I think he wants stub not mock, So it's ok for me without `expects($this->any)`.

HeahDude: `$this->any() !== $this->never()`, it cost nothing to add an assertion that breaks on behavioural change and has a real benefit IMO.

HeahDude: And I would say that using `method` and `willReturn` already make it a mock, not a stub to me.

tests/Sitemap/SitemapFactoryTest.php
tony-tran: In this projets, there is no return type on testXXX method.

tests/Sitemap/SitemapFactoryTest.php
tony-tran: The card of this PR is to test these privates functions. Is it better to change the visibility of these methods ?

HeahDude: No don't change the visibility of the methods because the tests look weird. The reflection classes are made for tests, hacks and hacky edge cases.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -27,34 +27,50 @@ class SitemapFactoryTest extends TestCase
...
 
         $this->assertEquals(1, $sitemap->getUrlCount());
     }
 
     public function testCreateContentSitemapWithoutHit()
     {
         $cacheItemInterface = $this->createMock(CacheItemInterface::class);
-        $cacheItemInterface->method('isHit')->willReturn(true);
-        $cacheItemInterface->method('get')->willReturn((string) new Sitemap());
+        $cacheItemInterface->expects($this->any())
```
HeahDude: CS: `expects()`and semi colon on a new line please.

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,134 @@
...
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Bundle\FrameworkBundle\Tests\TestCase;
+use Symfony\Component\Routing\RouterInterface;
+use Tackk\Cartographer\Sitemap;
+
+class SitemapFactoryTest extends TestCase
+{
+    private $objectManager;
+    private $routerInterface;
```
hhamon: `$router` ou `$urlGenerator`

tests/Sitemap/SitemapFactoryTest.php
```diff
@@ -0,0 +1,134 @@
...
+    private $cacheItemPoolInterface;
```
hhamon: `$cache`

tests/Sitemap/SitemapFactoryTest.php
B-Galati: Private methods are implementation details that we should not care in a test. In test we should rely on public API.
If it's needed to change the visibility of the methods to do the right test, it just means that the design is seriously bad.

HeahDude: I don't agree regarding unit tests everything can/should be tested.

B-Galati: Yes but through public API.

HeahDude:

Remg:
to rebase on `develop` once Behat is available

PR #1949, opened 2017-12-29
Validate Skill::name through Summary form

B-Galati:


src/Entity/BaseSkill.php
```diff
@@ -25,8 +25,8 @@
      *
      * @ORM\Column
      *
-     * @Assert\NotBlank
-     * @Assert\Length(min=2, max=200)
+     * @Assert\NotBlank(groups={"skills"})
```
HeahDude: This is good for other inclusion but I don't think that's not enough to fix the summary validation.
Note that from 3.4 we can use the `groups` option for the `Valid` constraint like here http://symfony.com/blog/new-in-symfony-3-4-groups-support-for-the-valid-constraint (ref https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/Summary.php#L164).
But my guess is that the best fix should be to add the `Default` group here https://github.com/EnMarche/en-marche.fr/blob/master/src/Form/SummaryType.php#L193, by doing:
```php
return ['Default', $options['step']];
```

PR #1956, opened 2018-01-02
[CP] Add order and empty option to the category field

illusionOfParadise:


src/CitizenProject/CitizenProjectCommand.php
```diff
@@ -53,6 +53,9 @@ class CitizenProjectCommand
      */
     public $phone;
 
+    /**
+     * @Assert\NotBlank
```
HeahDude: `NotNull` + `Type`

illusionOfParadise: Done

PR #1963, opened 2018-01-04
[CP] Added proposed solution and required means in show view

thomasbisignani:


thomasbisignani:
There are not tests for the "A propos" description bloc. I'm adding them. Thx !

PR #1965, opened 2018-01-04
[CP] Fixed proposal solution show field

thomasbisignani:


thomasbisignani:
@HeahDude The test for this field is here : https://github.com/EnMarche/en-marche.fr/blob/fix/cp-show-render-proposed-solution/tests/Controller/EnMarche/CitizenProjectControllerTest.php#L59

templates/citizen_project/show_base.html.twig
```diff
@@ -75,7 +75,7 @@
         </div>
 
         <div class="text--body text--dark" id="citizen-project-proposed-solution">
-            <p>{{ citizen_project.proposedSolution }}</p>
+            <p>{{ citizen_project.proposedSolution|raw }}</p>
```
B-Galati: No need for html purifier ?

thomasbisignani: Discussed in PM with @B-Galati.

PR #1967, opened 2018-01-04
Refactored procuration request summary

HeahDude:


tests/Controller/EnMarche/ProcurationControllerTest.php
```diff
@@ -616,6 +616,35 @@ public function testProcurationProposalManagerUuid()
         $this->isSuccessful($this->client->getResponse());
     }
 
+    public function testMyRequestRequiresProcessedRequest()
+    {
+        $this->client->request(Request::METHOD_GET, '/procuration/ma-demande/4/'.(Uuid::uuid4()->toString()));
+
+        $this->assertResponseStatusCode(Response::HTTP_NOT_FOUND, $this->client->getResponse());
```
B-Galati: Why do we need this obvious test case ?

HeahDude: When testing a controller, it's better to cover all returned status codes, in case a bug is introduced in vendors or if a condition or the behavior change then it will break 👍 
And here this is not obvious as it depends on a token generated by the entity matching the id.

B-Galati: Ok, why not but it's a fragile test. If the URL is changed the test would pass while it would test nothing. I know it's not going to happen and we hopefully gonna be smart and change this URL if it happens. 

One test of this kind should be enough no ? Why having 2 ?

HeahDude: One request is not valid (unprocessed).

PR #1968, opened 2018-01-04
[TIMELINE] Reindex related Themes on Algolia when a Measure is updated

Remg:
hotfix for Timeline backoffice


app/config/services/services.xml
```diff
@@ -41,6 +41,10 @@
 
 		<!-- Timeline -->
 		<service id="AppBundle\Command\ImportTimelineCommand" />
+		<service id="AppBundle\Timeline\MeasureManager" />
 		<service id="AppBundle\Timeline\TimelineFactory" />
+
+		<!-- Algolia -->
```
HeahDude: Please keep an alpha order for components.

src/Algolia/ManualIndexer.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace AppBundle\Algolia;
+
+use Algolia\AlgoliaSearchBundle\Indexer\Indexer;
+use Doctrine\ORM\EntityManagerInterface;
+
+class ManualIndexer implements ManualIndexerInterface
+{
+    /**
+     * @var \Algolia\AlgoliaSearchBundle\Indexer\ManualIndexer
```
HeahDude: CS: please add a use statement.

src/Algolia/ManualIndexer.php
```diff
@@ -0,0 +1,24 @@
...
+     */
+    private $indexer;
+
+    public function __construct(Indexer $algolia, EntityManagerInterface $manager)
+    {
+        $this->indexer = $algolia->getManualIndexer($manager);
```
HeahDude: Two possibilities:
- use a factory method for the service declaration and type hint `ManualIndexer` directly in the constructor
- keep both references and build the service only when `index()` is called

In this case I guess that if the service is construct the method will be called so the first one should be the way to go, not sure though.

app/config/services/services.xml
```diff
@@ -23,6 +23,12 @@
 		<service id="AppBundle\Membership\AdherentManager" />
 		<service id="AppBundle\Membership\AdherentRegistry" public="true" />
 
+		<!-- Algolia -->
+		<service id="AppBundle\Algolia\ManualIndexer">
+		    <factory class="Algolia\AlgoliaSearchBundle\Indexer\Indexer" method="getManualIndexer" />
+		    <argument type="service" id="Doctrine\ORM\EntityManagerInterface" />
```
HeahDude: Should be autowired

tests/Test/Algolia/DummyManualIndexer.php
```diff
@@ -0,0 +1,29 @@
...
+    public function __construct(LoggerInterface $logger = null)
+    {
+        $this->logger = $logger;
+    }
+
+    public function index($entities, array $options = []): void
+    {
+        if ($this->logger) {
+            if (is_array($entities)) {
+                $this->logger->info(sprintf('[algolia] indexing %d entities.', count($entities)));
```
B-Galati: Perhaps the logger context (2nd argument) is smart enough to dump `$entities` in a generic way so that this code could be simplified a lot.

Remg: i dont want to dump an array of entities if it's the case. i only want to know the count.
and i dont want to dump the entity if it's the case... only want to know which type of entity.

app/config/parameters.yml.dist
```diff
@@ -48,6 +48,7 @@ parameters:
     env(ALGOLIA_FRONT_API_KEY): ~
     env(ALGOLIA_BACK_API_KEY):  ~
     env(ALGOLIA_BLACKLIST):     ~
+    env(ALGOLIA_INDEX_PREFIX):  'dev'
```
HeahDude: `'dev'` => `'%kernel.environment%'`?

Remg: actually, no.
Algolia indexes are already suffixed by the APP_ENV environment variable.

I introduced this new env variable so we can have differents indexes for each environment.
(staging has APP_ENV=prod, like production., but we want different indexes for staging & prod !)

HeahDude: ok, was just asking ^^', thx!

PR #1973, opened 2018-01-05
[TIMELINE] Removing extra fields in many-to-many between Themes and Measures

Remg:
- ManyToMany between Theme and Measure does not need extra fields.

- property `Measure.global` is renamed to `Measure.major` to be consistent with the UI.

- the Themes should be flagged in the Measure admin, not the other way.

- some Timeline admin fix/improvements


src/Admin/TimelineMeasureAdmin.php
```diff
@@ -59,6 +62,27 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                'label' => 'Statut',
+                'show_filter' => true,
+                'field_type' => ChoiceType::class,
+                'field_options' => [
+                    'multiple' => true,
+                    'choices' => Measure::STATUSES,
+                ],
+            ])
+            ->add('major', null, [
+                'label' => 'Mise en avant (32)',
```
HeahDude: Why `(32)`? Forgotten debug or specific request?

src/Entity/Timeline/Theme.php
```diff
@@ -73,20 +74,29 @@ class Theme implements EntityMediaInterface
...
+     * @ORM\ManyToMany(targetEntity="AppBundle\Entity\Timeline\Measure", mappedBy="themes")
+     *
+     * @Algolia\Attribute
      */
     private $measures;
 
-    public function __construct()
-    {
+    public function __construct(
+        ?string $title = null,
```
HeahDude: The `?` is needless when the parameter is set to `null` by default.

src/Admin/TimelineMeasureAdmin.php
Remg: They just want `Mise en avant (32)` this to be the label of this specific field.

Remg: nice catch though, after update they are consistent in every admin now

PR #1976, opened 2018-01-08
[TIMELINE] i18n

Remg:


src/Admin/TimelineMeasureAdmin.php
```diff
@@ -23,14 +24,41 @@ public function __construct($code, $class, $baseControllerName, MeasureManager $
...
-                ->add('title', TextType::class, [
-                    'label' => 'Titre',
-                    'filter_emojis' => true,
+            ->with('Traductions', ['class' => 'col-md-6'])
+                ->add('translations', TranslationsType::class, [
+                    'by_reference' => false,
+                    'label' => false,
+                    'default_locale' => ['fr'],
+                    'locales' => ['fr', 'en'],
+                    //'required_locales' => ['fr', 'en'],
```
HeahDude: Please remove or uncomment, or add a reason why this is commented.

src/Controller/Admin/AdminTimelineMeasureController.php
```diff
@@ -2,12 +2,20 @@
...
 use AppBundle\Entity\Timeline\Measure;
 use Sonata\AdminBundle\Controller\CRUDController as Controller;
 use Symfony\Component\HttpFoundation\Request;
 
 class AdminTimelineMeasureController extends Controller
 {
+    /**
+     * @I18nDoctrine
+     */
+    public function listAction() {
```
HeahDude: CS

src/Entity/Timeline/Measure.php
```diff
@@ -307,4 +298,28 @@ public function profileIds(): array
             return $profile->getId();
         }, $this->profiles->toArray());
     }
+
+    /**
+     * @Algolia\Attribute
+     */
+    public function title(): array
```
HeahDude: Can you rename it `titles`?

src/Entity/Timeline/Measure.php
```diff
@@ -129,6 +122,11 @@ class Measure
 
     private $savedThemes;
 
+    /**
+     * @Assert\Valid
+     */
+    protected $translations;
```
HeahDude: I would name it `translatedTitles`.

src/Entity/Timeline/MeasureTranslation.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->locale = $locale;
+        $this->title = $title;
+    }
+
+    public function getTitle(): ?string
+    {
+        return $this->title;
+    }
+
+    public function setTitle(string $title): void
```
HeahDude: `?string`

templates/admin/form_translations.html.twig
```diff
@@ -0,0 +1,31 @@
...
+    {{ form_errors(form) }}
+    <div class="a2lix_translations tabbable">
+        <ul class="a2lix_translationsLocales nav nav-tabs">
+            {% for translationsFields in form %}
+                {% set locale = translationsFields.vars.name %}
+
+                <li class="{% if app.request.locale == locale %}active {% endif %}{% if not translationsFields.vars.valid %}error{% endif %}">
+                    <a href="#" data-toggle="tab" data-target=".{{ translationsFields.vars.id }}_a2lix_translationsFields-{{ locale }}">
+                        {{ locale|capitalize }}
+                        {% if form.vars.default_locale == locale %}[Default]{% endif %}
```
HeahDude: `==` => `is same as()`

src/Admin/TimelineMeasureAdmin.php
Remg: will remove it

src/Entity/Timeline/Measure.php
Remg: this translations are meant to hold all the translated field of this entity

also, as we are using the `A2lix\I18nDoctrineBundle\Doctrine\ORM\Util\Translation` trait, the property name should be `translations`.

src/Entity/Timeline/Measure.php
Remg: i can on my end yes.
i will wait for approval of the front who needs this algolia attribute under this name though

src/Entity/Timeline/Theme.php
```diff
@@ -85,59 +53,33 @@ class Theme implements EntityMediaInterface
      */
     private $measures;
 
+    /**
+     * @Assert\Valid
+     */
+    protected $translations;
```
HeahDude: Why using `protected` here?

src/Entity/Timeline/Theme.php
```diff
@@ -202,4 +144,58 @@ public function profileIds(): array
...
+        return $descriptions ?? [];
+    }
+
+    private function getTranslation(string $locale): ?ThemeTranslation
+    {
+        $translation = $this->translations->filter(function (ThemeTranslation $translation) use ($locale) {
+            return $locale === $translation->getLocale();
+        })->first();
+
+        return $translation ? $translation : null;
```
HeahDude: `return $translation ?: null;`

src/Entity/Timeline/Theme.php
Remg: i was following the documentation blindly.
i can make it private

src/Timeline/ProfileManager.php
```diff
@@ -0,0 +1,41 @@
...
+class ProfileManager
+{
+    private $algolia;
+
+    public function __construct(ManualIndexerInterface $algolia)
+    {
+        $this->algolia = $algolia;
+    }
+
+    public function postPersist(Profile $profile): void
```
HeahDude: Can you add an `AlgoliaIndexedEntityInterface` so we have only one `AlgoliaIndexedEntityManager`?

src/Admin/TimelineThemeAdmin.php
```diff
@@ -82,4 +157,28 @@ protected function configureListFields(ListMapper $listMapper)
             ])
         ;
     }
+
+    /**
+     * @param Theme $object
+     */
+    public function postPersist($object)
```
HeahDude: This class should use an `AlgoliaIndexedEntityAdminTrait` that handle an interface though a common manager to avoid adding these methods again and again.

src/Timeline/ProfileManager.php
Remg: reminder: `MeasureManager::unIndex` method slightly differs from others

HeahDude: Then in `AlgoliaIndexedEntityManager::index($entity)`, do `if ($entity instanceof Measure)`

HeahDude: Or maybe add an `getElementsToIndex` method in the `AlgoliaIndexedEntityInterface`?

Remg: i guess the functional logic is too specific.
`MeasureManager::unIndex` method needs to re-index related themes and un-index the measure.
when `MeasureManager::index` method needs to index related themes and index the measure.

I prefer your first option:
```
Then in AlgoliaIndexedEntityManager::index($entity), do if ($entity instanceof Measure)
```
:ok_hand:

src/Entity/Timeline/ProfileTranslation.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @ORM\Column(type="text")
+     *
+     * @Assert\NotBlank
+     *
+     * @Algolia\Attribute
+     */
+    private $description;
+
+    public function __construct(string $locale = null, string $title = null, string $slug = null, string $description = null)
```
thomasbisignani: CS : Can you declare this construct in multiples lignes for more visibility ?

src/Admin/AlgoliaIndexedEntityAdminTrait.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function postUpdate($object)
+    {
+        $this->algoliaManager->postUpdate($object);
+    }
+
+    /**
+     * @param AlgoliaIndexedEntityInterface $object
+     */
+    public function preRemove($object)
```
HeahDude: Missing `: void` on the three methods.

src/Admin/EmptyTranslationRemoverAdminTrait.php
```diff
@@ -0,0 +1,27 @@
...
+     * @var EmptyTranslationRemoverListener
+     */
+    private $emptyTranslationRemoverListener;
+
+    public function setEmptyTranslationRemoverListener(EmptyTranslationRemoverListener $listener): void
+    {
+        $this->emptyTranslationRemoverListener = $listener;
+    }
+
+    private function removeEmptyTranslationsOnSubmit(FormMapper $formMapper): void
```
HeahDude: You should pass the builder here.

src/Admin/TimelineMeasureAdmin.php
```diff
@@ -53,23 +57,61 @@ protected function configureFormFields(FormMapper $formMapper)
...
+            ])
             ->add('themes', null, [
                 'label' => 'Thèmes',
                 'show_filter' => true,
-            ], null, ['multiple' => true])
+            ], null, [
+                'multiple' => true,
+                'query_builder' => function (ThemeRepository $repository) {
+                    return $repository
+                        ->createQueryBuilder('theme')
```
HeahDude: You should use a custom method of the repo instead. Since the three are similar I guess you could use trait for them as well.

src/Admin/AlgoliaIndexedEntityAdminTrait.php
HeahDude: Yes indeed.

PR #1943, opened 2017-12-29
Timeline backoffice

Remg:

PR #1977, opened 2018-01-08
Refactored procuration matching

HeahDude:


src/Procuration/Filter/ProcurationProxyProposalFilters.php
```diff
@@ -37,20 +37,20 @@ public function apply(QueryBuilder $qb, string $alias): void
 
         $status = $this->getStatus();
         if (self::UNASSOCIATED === $status) {
-            $qb->andWhere(sprintf('%s.foundRequest IS NULL', $alias));
+            $qb->andWhere("$alias.foundRequest IS NULL");
```
thomasbisignani: Why remove these `sprintf` ?

templates/procuration/_request_reasons.html.twig
```diff
@@ -0,0 +1,17 @@
+<div class="profile-value" id="request-reason">
+    {% if request.reason == constant('REASON_RESIDENCY', request) %}
```
thomasbisignani: What do you think to refactor this elseif suite in one line with translations, for example :
`('procuration.request.reason.' ~ request.reason)|trans`

src/Procuration/Filter/ProcurationProxyProposalFilters.php
Remg: @thomasbisignani https://github.com/EnMarche/en-marche.fr/pull/1943#discussion_r159080614

templates/procuration/_request_reasons.html.twig
Remg: what is the reason is unknown or not translated ? the current logic does not provide a fallback label

HeahDude: This is a choice with a constraint, no need for a fallback.

HeahDude: Good call, done in https://github.com/EnMarche/en-marche.fr/pull/1977/commits/b5fe1fd7370a3e893082c5b6fb6d2e0fdab5c53e.

PR #1978, opened 2018-01-08
Added admin pages to handle elections

HeahDude:


app/config/security.yml
```diff
@@ -152,6 +152,13 @@ security:
             - ROLE_APP_ADMIN_PROCURATION_PROXY_EDIT
             - ROLE_APP_ADMIN_PROCURATION_PROXY_DELETE
 
+        ROLE_ADMIN_ELECTIONS:
+            - ROLE_APP_ADMIN_ELECTION_LIST
+            - ROLE_APP_ADMIN_ELECTION_VIEW
+            - ROLE_APP_ADMIN_ELECTION_EDIT
+            - ROLE_APP_ADMIN_ELECTION_CREATE
+            - ROLE_APP_ADMIN_ELECTION_DELETE
```
Remg: instead of adding every role, you could simply add `ROLE_APP_ADMIN_ELECTION_ALL`

HeahDude: I don't want all roles :p. But thanks!

PR #1987, opened 2018-01-09
[CP] Save geographic coordinates after modifying CP address

illusionOfParadise:


src/Geocoder/Subscriber/EntityAddressGeocodingSubscriber.php
```diff
@@ -99,6 +100,11 @@ public function onCitizenProjectCreated(CitizenProjectWasCreatedEvent $event): v
         $this->updateGeocodableEntity($event->getCitizenProject());
     }
 
+    public function onCitizenProjectUpdated(CitizenProjectWasUpdatedEvent $event): void
```
HeahDude: Method name should be meaningful about what it is doing, the register method already tells us when.

What about `updateCoordinates`?

illusionOfParadise: Done

PR #1991, opened 2018-01-09
UI tweaks on the procuration pages

marineb:


templates/procuration/choose_election.html.twig
```diff
@@ -6,74 +6,45 @@
...
 
         {{ form_end(elections_form) }}
+
+        <div class="text--gray">
+            {{ include('procuration/legal.html.twig', { elections: elections_form.vars.elections|join(', ') }, with_context=false) }}
+        </div>
+
     </section>
 
-    {{ include('procuration/legal.html.twig', { elections: elections_form.vars.elections|join(', ') }, with_context=false) }}
```
HeahDude: Be careful if you remove this, it is not included anywhere else for now.

marineb: I didn't take it out, I just moved it!

PR #1997, opened 2018-01-10
Fixed tests and admin

HeahDude:


src/Entity/ProcurationProxy.php
```diff
@@ -571,6 +551,21 @@ public function getAvailableRounds(): Collection
         });
     }
 
+    public function getAvailableRoundsAsString(): string
+    {
+        $availableRounds = $this->getAvailableRounds();
+
+        if ($this->electionRounds->count() === $availableRounds->count()) {
+            return 'Tous les tours proposés';
```
thomasbisignani: What about to use a const here ?

src/Entity/ProcurationProxy.php
```diff
@@ -571,6 +551,21 @@ public function getAvailableRounds(): Collection
...
+        }
+
+        if ($availableRounds->isEmpty()) {
+            return 'Aucun';
```
thomasbisignani: Same here, can you use a const ?

PR #2000, opened 2018-01-10
Copy and design updates on the PC

marineb:


front/style/layout/_forms.scss
```diff
@@ -559,11 +564,12 @@ input[type=date]::-webkit-inner-spin-button {
 
         input[type="checkbox"] {
             width: 20px;
+            flex: none;
         }
 
         label {
             order: 2;
-            @include text--small;
+            // @include text--small;
```
thomasbisignani: This comment is usefull ? can you remove this line ?

Remg: should you remove this line?

marineb: Yes, I saw that was doing it it now 🙊

templates/citizen_project/_form.html.twig
```diff
@@ -147,17 +147,38 @@
     </div>
 </div>
 
+<div class="form__new" id="citizen-project-phone">
+    <div class="register__form__phone">
+        {{ form_label(form.required_means, 'Votre téléphone :', {label_attr: {class: 'label__header--large'}}) }}
+        <div class="l__row l__row--top l__row--h-stretch">
+            {{ form_widget(form.phone.country) }}
+            {{ form_widget(form.phone.number, {attr: {placeholder: 'Téléphone' }}) }}
```
HeahDude: We should be consistent and always put spaces inside curly braces to improve the readability of hashes (here there is one at the end, but not at the start and not for both).

PR #2003, opened 2018-01-11
Add onDelete on 'foundBy' of ProcurationRequest

illusionOfParadise:


src/Entity/Adherent.php
```diff
@@ -165,7 +165,7 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
     /**
      * @var ProcurationManagedArea|null
      *
-     * @ORM\OneToOne(targetEntity="AppBundle\Entity\ProcurationManagedArea", mappedBy="adherent", cascade={"persist"})
+     * @ORM\OneToOne(targetEntity="AppBundle\Entity\ProcurationManagedArea", mappedBy="adherent", cascade={"persist", "remove"}, orphanRemoval=true)
```
HeahDude: `cascade={"all"}` would be better.

PR #1600, opened 2017-10-09
[EMAILS] Manage & test templates in the application

Remg:

PR #2005, opened 2018-01-11
[CP] Sends a summary email with approved citizen projects around adherents

thomasbisignani:


src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->logger = $container->get('logger');
+        $this->twig = $container->get('twig');
+
+        $this->adherentRepository = $this->entityManager->getRepository(Adherent::class);
+        $this->citizenProjectRepository = $this->entityManager->getRepository(CitizenProject::class);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $adherents = $this->adherentRepository->findBy(['status' => Adherent::ENABLED]);
```
HeahDude: You should batch, as it may break considering the number of results here.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,164 @@
...
+    {
+        return [
+            'name' => $citizenProject->getName(),
+            'subtitle' => $citizenProject->getSubtitle(),
+            'cityName' => $citizenProject->getCityName(),
+            'slug' => $citizenProject->getSlug(),
+        ];
+    }
+
+    private function createMessageFactory(
```
HeahDude: `createMessage`

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,164 @@
...
+            $this->logger->error(
+                '[Broadcasting citizen projects] An error occurred while broadcasting',
+                ['exception' => $e]
+            );
+        }
+    }
+
+    private function boadcastedCitizenProjectFactory(CitizenProject $citizenProject): array
+    {
+        return [
```
HeahDude: Could be returned by `CitizenProject::getSummary`

HeahDude: Actually this method seems useless, you should do an array query on those fields to save memory.

thomasbisignani: Actually the `findNearByCitizenProjectSummariesForAdherent` return an array of `CitizenProject` because the `broadcast` method iterate them and put the flag.

I think a `CitizenProject::getSummary` static method is a good idea. It's ok for you ?

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,155 @@
...
+            $this->entityManager->beginTransaction();
+
+            /** @var CitizenProject $citizenProject */
+            foreach ($citizenProjects as $citizenProject) {
+                if ($citizenProject->isBroadcasted()) {
+                    continue;
+                }
+
+                $citizenProject->setBroadcasted(true);
+                $this->entityManager->persist($citizenProject);
```
HeahDude: This line is useless, the entity is already tracked by doctrine since it has fetched it from the db.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            // Broadcast the citizen projects and returns them formatted in array
+            $broadcastedCitizenProjects = $this->broadcast($citizenProjects);
+
+            // Sends the message
+            $message = $this->createMessage($broadcastedCitizenProjects, $adherent);
+            $this->mailer->sendMessage($message);
+
+            $this->logger->info(
```
HeahDude: I would move the message handling in `broadcast` so this private method could easily be replaced by a service.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,155 @@
...
+        } catch (Exception $e) {
+            $this->logger->error(
+                '[Broadcasting citizen projects] An error occurred while broadcasting',
+                ['exception' => $e]
+            );
+        }
+    }
+
+    private function createMessage(
+        array $broadcastedCitizenProjects, Adherent $adherent): CitizenProjectApprovedSummaryMessage
```
HeahDude: CS

HeahDude: This method should be moved in a message factory. So a broadcaster could also depend on it.

src/Mailer/Message/CitizenProjectApprovedSummaryMessage.php
```diff
@@ -0,0 +1,28 @@
...
+final class CitizenProjectApprovedSummaryMessage extends Message
+{
+    public static function create(Adherent $adherent, string $template): self
+    {
+        $message = new self(
+            Uuid::uuid4(),
+            '289976',
+            $adherent->getEmailAddress(),
+            $adherent->getFullName(),
+            ' [Projets citoyens] Les projets citoyens près de chez vous !',
```
illusionOfParadise: Extra space in the beginning ?

templates/citizen_project/_email_broadcasting_content.html.twig
```diff
@@ -0,0 +1,11 @@
+<div>
+    {% for citizen_project in citizen_projects %}
+   <div>
+       <a href="{{ path('app_citizen_project_show', {slug: citizen_project.slug}) }}" title="{{ citizen_project.name }}">
```
illusionOfParadise: Please add spaces
https://github.com/EnMarche/en-marche.fr/pull/2000#discussion_r160898993

src/Repository/CitizenProjectRepository.php
```diff
@@ -120,6 +120,43 @@ public function searchAll(SearchParametersFilter $search): iterable
...
     }
 
+    /**
+     * Finds all citizen projects around an Adherent.
+     *
+     * @param Adherent $adherent
+     * @param int      $limit    Limit of citizen project
+     * @param int      $maxAge   Duration in days
+     *
+     * @return array List of CitizenProject
```
Remg: `@return array List of CitizenProject` => `@return CitizenProject[]`

but these docblocks are not very usefull imho

src/Repository/CitizenProjectRepository.php
```diff
@@ -120,6 +120,43 @@ public function searchAll(SearchParametersFilter $search): iterable
...
+    public function findNearByCitizenProjectSummariesForAdherent(Adherent $adherent, int $limit, ?int $maxAge): array
+    {
+        $qb = $this
+            ->createNearbyQueryBuilder(new Coordinates($adherent->getLatitude(), $adherent->getLongitude()))
+            ->andWhere($this->getNearbyExpression().' <= :distance_max')
+            ->setParameter('distance_max', $adherent->getCitizenProjectCreationEmailSubscriptionRadius());
+
+        if (null !== $maxAge && $maxAge > 0) {
+            $dateMin = new \DateTime('now');
+            $dateMin->modify('-'.$maxAge.' days');
```
Remg: `$dateMin->modify("-$maxAge days");`

src/Repository/CitizenProjectRepository.php
HeahDude: Specialized arrays return do help the IDE and avoid adding "duck type hint" in the code when manipulating entries (i.e in foreach loops).

src/Mailer/Message/CitizenProjectApprovedSummaryMessage.php
```diff
@@ -0,0 +1,28 @@
...
+            [
+                'target_firstname' => self::escape($adherent->getFirstName()),
+                'citizen_project_list' => self::escape($template),
+            ]
+        );
+
+        $message->setSenderEmail('projetscitoyens@en-marche.fr');
```
HeahDude: Another reason to use a message factory, this should be injected.

Remg: i agree.
this comment should be for #1600

src/Repository/CitizenProjectRepository.php
```diff
@@ -120,6 +120,44 @@ public function searchAll(SearchParametersFilter $search): iterable
         ;
     }
 
+    /**
+     * Finds all citizen projects around an Adherent.
+     *
+     * @param Adherent $adherent
+     * @param int      $limit         Limit of number of citizen projects to return
+     * @param int      $approvedSince Duration in format Time since the citizen projects has been approved
```
illusionOfParadise: `has been` -> `have been`

src/Entity/CitizenProject.php
```diff
@@ -161,7 +161,7 @@ class CitizenProject extends BaseGroup implements CoordinatorAreaInterface
      * @var CitizenAction|null
      */
     private $nextAction;
-
+
```
Remg: extra spaces

src/Repository/AdherentRepository.php
```diff
@@ -546,4 +547,17 @@ public function findAdherentsUuidByEmailAddress(string $emailAddress): array
...
+
+    public function findAllEnabledAdherents(int $offset = 0): IterableResult
+    {
+        $qb = $this->createQueryBuilder('a');
+
+        $query = $qb
+            ->andWhere('a.status = :status')
+            ->setParameter('status', Adherent::ENABLED)
+            ->setFirstResult($offset)
+            ->getQuery();
```
Remg:
```
            ->getQuery()
    ;
```

src/Repository/AdherentRepository.php
```diff
@@ -546,4 +547,17 @@ public function findAdherentsUuidByEmailAddress(string $emailAddress): array
...
+
+        return $query->iterate();
```
Remg: you could just chain the return i guess:
```
return $this
    ->createQueryBuilder('a')
    ->andWhere('a.status = :status')
    // ..
    ->getQuery()
    ->iterate()
;
```

Remg: then you can chain the return from the query builder

HeahDude: >I agree but to stay consistent

Consistent with what? we never did that before... I agree we should separate implementations though, but only in regard of the first line caller otherwise I don't see the benefit in case of a builder.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+
+    /** @var CitizenProjectRepository */
+    private $citizenProjectRepository;
+
+    protected function configure()
+    {
+        $this
+          ->setName(self::COMMAND_NAME)
+          ->addOption(
+              'approvedSince',
```
HeahDude: I think the convention for command arguments and options name is to use snake case.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+              'as',
```
HeahDude: Is this really needed? No so obvious IMO.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+              InputOption::VALUE_OPTIONAL,
+              'Duration in format Time since the citizen projects has been approved.',
+              self::DEFAULT_APPROVED_SINCE
+          )
+          ->addOption('offset', 'of', InputOption::VALUE_OPTIONAL, 'Offset to start the query.', self::DEFAULT_OFFSET)
```
HeahDude: I don't think we need aliases for custom commands, full names keeps cron tasks and others readable.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+
+        $adherentCount = 0;
+        $emailCount = 0;
+
+        foreach ($adherents as $chunk) {
+            ++$adherentCount;
+
+            /** @var Adherent $adherent */
+            $adherent = $chunk[0];
+            $this->logger->info(
```
HeahDude: This will add hundreds of thousands of lines of debug somewhere, I guess logging in the broadcaster, plus the final result here should be enough. Maybe one before the loop to warn that the script is launching if you want though.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+                sprintf(
+                    '[Broadcasting citizen projects] Treating adherent... (id #%d) (offset: %d)',
+                    $adherent->getId(),
+                    $adherentCount
+                )
+            );
+
+            // Finds all citizen projects near the Adherent
+            $citizenProjects = $this->citizenProjectRepository->findNearByCitizenProjectSummariesForAdherent(
```
HeahDude: Since you get the entity and not a special "summary" array query the method should be renamed `findNearByCitizenProjectsForAdherent`

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+                self::CITIZEN_PROJECTS_SUMMARY_LIMIT,
+                $input->getOption('approvedSince')
+            );
+
+            if (0 === count($citizenProjects)) {
+                continue;
+            }
+
+            // Broadcast the citizen projects
+            $this->broadcast->handle($citizenProjects, $adherent);
```
HeahDude: `$this->broadcaster->broadcast()`

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+
+            $this->entityManager->clear(CitizenProject::class);
```
HeahDude: It would be better to iterate over bunches and clear the manager for each bunch instead of each adherent.

src/CitizenProject/CitizenProjectBroadcast.php
```diff
@@ -0,0 +1,68 @@
...
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\CitizenProject;
+use AppBundle\Mailer\MailerService;
+use AppBundle\Mailer\Message\CitizenProjectApprovedSummaryMessage;
+use AppBundle\Mailer\Message\Message;
+use Psr\Log\LoggerInterface;
+use Twig\Environment;
+
+class CitizenProjectBroadcast
```
HeahDude: `CitizenProjectBroadcaster`

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,109 @@
...
+            );
+
+            // Finds all citizen projects near the Adherent
+            $citizenProjects = $this->citizenProjectRepository->findNearByCitizenProjectSummariesForAdherent(
+                $adherent,
+                self::CITIZEN_PROJECTS_SUMMARY_LIMIT,
+                $input->getOption('approvedSince')
+            );
+
+            if (0 === count($citizenProjects)) {
```
HeahDude: `if (!$citizenProjects)`

src/CitizenProject/CitizenProjectBroadcast.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->logger = $logger;
+    }
+
+    public function handle(array $citizenProjects, Adherent $adherent): void
+    {
+        try {
+            $broadcastedCitizenProjects = [];
+
+            /** @var CitizenProject $citizenProject */
+            foreach ($citizenProjects as $citizenProject) {
```
HeahDude: What's the purpose of this loop?

src/CitizenProject/CitizenProjectBroadcast.php
```diff
@@ -0,0 +1,68 @@
...
+    private $logger;
+
+    public function __construct(MailerService $mailer, Environment $twig, LoggerInterface $logger)
+    {
+        $this->mailer = $mailer;
+        $this->twig = $twig;
+        $this->logger = $logger;
+    }
+
+    public function handle(array $citizenProjects, Adherent $adherent): void
```
HeahDude: Imo, the broadcaster should be responsible for querying the projects, it then only needs one adherent to broadcast to.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
Remg: we need to keep a track of the offset in case the command crashed during execution.
we should be able to restart the command with approximately the last offset.

we should log this offset every 100 or 200 adherents

HeahDude: >we should log this offset every 100 or 200 adherents

Consistent with my advice to use bunches to iterate on.

HeahDude: Couldn't you add the culprit log in a `catch` instead?

src/CitizenProject/CitizenProjectBroadcast.php
HeahDude: Should be named `broadcastNearbySummariesToAdherent()`.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $container = $this->getContainer();
+
+        $this->entityManager = $container->get('doctrine.orm.entity_manager');
+        $this->logger = $container->get('logger');
+        $this->broadcaster = $container->get(CitizenProjectBroadcaster::class);
+
+        $this->adherentRepository = $this->entityManager->getRepository(Adherent::class);
+        $this->citizenProjectRepository = $this->entityManager->getRepository(CitizenProject::class);
```
HeahDude: Looks unused nom.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $adherents = $this->adherentRepository->findAdherentsByCitizenProjectCreationEmailSubscription(
```
HeahDude: You still don't batch here.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,87 @@
...
+            $input->getOption('offset')
+        );
+        $adherentCount = 0;
+
+        foreach ($adherents as $chunk) {
+            /** @var Adherent $adherent */
+            $adherent = $chunk[0];
```
HeahDude: You just need one column here the uuid, you should loop in batch array column result by adding a where clause to handle `isEligibleToCitizenProjectBroadcast` instead of building hundreds of thousands of objects and clear the manager on each iteration.

src/Command/MailerCitizenProjectValidatedSummaryCommand.php
```diff
@@ -0,0 +1,87 @@
...
+
+            if (!$adherent->isEligibleToCitizenProjectBroadcast()) {
+                continue;
+            }
+
+            $this->producer->scheduleBroadcast($adherent, $input->getOption('approved_since'));
```
HeahDude: Just pass the uuid here, and please move `getOption` before the loop.

templates/citizen_project/_email_summary.html.twig
```diff
@@ -0,0 +1,11 @@
+<div>
+    {% for citizen_project in citizen_projects %}
+   <div>
+       <a href="{{ path('app_citizen_project_show', { slug: citizen_project.slug }) }}" title="{{ citizen_project.name }}">
```
Remg: Since this template is rendered by a worker, you need to generate an absolute url.
Otherwise scheme and/or domain will be missing in the link depending on the configuration.

Please use `url()` instead of `path()`.

src/Consumer/CitizenProjectSummaryConsumer.php
```diff
@@ -0,0 +1,55 @@
...
+
+            if (null === $adherent) {
+                $this->writeln('Adherent not found', 'Adherent with '.$data['adherent_uuid'].' uuid not found');
+                $this->getLogger()->error('Adherent with '.$data['adherent_uuid'].' not found', $data);
+
+                return self::MSG_ACK;
+            }
+
+            $this->broadcaster->broadcast($adherent, $data['approved_since']);
+            $this->manager->clear();
```
HeahDude: You don't need this line anymore thanks to https://github.com/EnMarche/en-marche.fr/pull/2032.

src/Entity/Adherent.php
```diff
@@ -1019,4 +1019,9 @@ public function getCoordinatorManagedAreaCodesAsString(): string
             return $area->getCodesAsString();
         }, $this->coordinatorManagedAreas->toArray()));
     }
+
+    public function isEligibleToCitizenProjectBroadcast(): bool
```
HeahDude: I guess you can remove this method now.

src/Repository/AdherentRepository.php
```diff
@@ -546,4 +546,29 @@ public function findAdherentsUuidByEmailAddress(string $emailAddress): array
             return $uuid->toString();
         }, array_column($query->getArrayResult(), 'uuid'));
     }
+
+    public function findAdherentsUuidByCitizenProjectCreationEmailSubscription(?int $offset = 0): array
+    {
+        $qb = $this->createQueryBuilder('a');
+
+        $qb
```
HeahDude: You can remove those two lines, and chain directly.

src/Repository/AdherentRepository.php
```diff
@@ -546,4 +546,29 @@ public function findAdherentsUuidByEmailAddress(string $emailAddress): array
...
+            ->setParameter('radius', Adherent::DISABLED_CITIZEN_PROJECT_EMAIL)
+            ->andWhere('a.postAddress.latitude IS NOT NULL')
+            ->andWhere('a.postAddress.longitude IS NOT NULL')
+        ;
+
+        if (null !== $offset) {
+            $qb->setFirstResult($offset);
+        }
+
+        $query = $qb->getQuery();
```
HeahDude: Useless var.

templates/citizen_project/_email_summary.html.twig
```diff
@@ -0,0 +1,11 @@
+<div>
+    {% for citizen_project in citizen_projects %}
+   <div>
```
HeahDude: CS: please fix indent.

src/Consumer/CitizenProjectSummaryConsumer.php
Remg: #2032 is now merged, you can rebase and remove this `clear` \o/

src/CitizenProject/CitizenProjectBroadcaster.php
```diff
@@ -0,0 +1,64 @@
...
+            $adherent,
+            self::CITIZEN_PROJECTS_SUMMARY_LIMIT,
+            $approvedSince
+        );
+
+        if (!$citizenProjects) {
+            return;
+        }
+
+        $message = $this->createMessage($citizenProjects, $adherent);
```
HeahDude: Useless var.

src/CitizenProject/CitizenProjectBroadcaster.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->mailer->sendMessage($message);
+    }
+
+    private function createMessage(array $citizenProjects, Adherent $adherent): Message
```
HeahDude: Could be static.

src/CitizenProject/CitizenProjectBroadcaster.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $summary = $this->twig->render(
```
HeahDude: Please inline this call.

thomasbisignani: This call is too long for using inline.

src/CitizenProject/CitizenProjectBroadcaster.php
thomasbisignani: `$this` is used inside this method.

src/CitizenProject/CitizenProjectBroadcaster.php
HeahDude: Rename the var `$citizenProjects` to `$projects` :trollface:?

No just kidding :p

src/CitizenProject/CitizenProjectBroadcaster.php
```diff
@@ -55,9 +59,23 @@ private function createMessage(array $citizenProjects, Adherent $adherent): Mess
...
+                [],
+                UrlGeneratorInterface::ABSOLUTE_URL
+            ),
+        ];
+
         return CitizenProjectApprovedSummaryMessage::create(
             $adherent,
-            $summary
+            $summary,
+            $urls
```
Remg: keys `all_citizen_projects_url` and `email_notifications_url` belongs to the EmailTemplate logic, not the broadcaster.

can you please send 2 distincts `string` variables for these dynamic url?

HeahDude:

PR #2007, opened 2018-01-12
Fix 'Mes événements' page

illusionOfParadise:


guillaumedenhez:
@illusionOfParadise what is this for ?

PR #2009, opened 2018-01-12
 Auth - OAuth server 

B-Galati:


Makefile
```diff
@@ -6,7 +6,7 @@ PHPCSFIXER?=$(RUN) php -d memory_limit=1024m vendor/bin/php-cs-fixer
 
 .DEFAULT_GOAL := help
 .PHONY: help start stop reset db db-diff db-migrate db-rollback db-load watch clear clean test tu tf tj lint ls ly lt
-.PHONY: lj build up perm deps cc phpcs phpcsfix tty
+.PHONY: lj build up perm deps cc phpcs phpcsfix tty fix-keys-perms
```
blazarecki: `fix-keys-perms` ? I don't find it in the Makefile

app/config/config.yml
```diff
@@ -36,7 +36,7 @@ framework:
     form:            ~
     csrf_protection: ~
     validation:      { enable_annotations: true }
-    #serializer:      { enable_annotations: true }
+    serializer:      { enable_annotations: true }
```
blazarecki: Do you still need to enable symfony serialiser since you use JMS ?

blazarecki:
@B-Galati Behat is run on each container on CircleCI. You can ask @Remg how to use only one.

B-Galati:
@blazarecki @Remg for the record, it's alright like this because the parallel option is not set to true

PR #2010, opened 2018-01-12
Fix event description display

illusionOfParadise:


src/Form/BaseEventCommandType.php
```diff
@@ -15,6 +16,8 @@ class BaseEventCommandType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $needToPurifyHtml = $builder->getData() instanceof BaseEvent ? !$builder->getData()->isCitizenAction() : false;
```
HeahDude: You should override the option in the child form instead of adding overhead on all of them.
In https://github.com/EnMarche/en-marche.fr/blob/master/src/Form/CitizenActionCommandType.php#L18, you can add:
```diff
 $builder
     ->remove('capacity')
+    ->add('description', TextareaType::class, [
+        'filter_emojis' => true,
+        'purify_html' => true,
+    ])
 ;
```

templates/events/show.html.twig
```diff
@@ -166,7 +166,11 @@
             </ul>
             <div class="committee-event-description text--with-bullet b__nudge--bottom-huge l__wrapper--slim">
                 {%- block main_section -%}
-                    {{- event.description -}}
+                    {%- if event.isCitizenAction -%}
```
HeahDude:
```twig
{{ event.isCitizenAction ? event.description : event.description|raw }}
```

illusionOfParadise: Does not work as expected, always act like `{{ event.description }}`

src/Form/BaseEventCommandType.php
illusionOfParadise: Done

PR #2020, opened 2018-01-16
remove h1 on navbar desktop / mobile

NyKo24:


front/style/mixins/_text.scss
```diff
@@ -513,7 +513,7 @@ $text--size-extra-large: 48px;
             font-size: $text--size-default;
             letter-spacing: 0;
 
-            h1 {
+            .text--logo {
```
marineb: I think you could completely remove this class, no ? Since there's already a `text--logo` class that exists.

NyKo24: ah ? Where ?

marineb: Higher up in that same file. You have `--logo`. That's attached to `.text`.

NyKo24: ✅

PR #2021, opened 2018-01-16
add language attribute on html

NyKo24:


templates/base.html.twig
```diff
@@ -19,7 +19,7 @@
 |______|_| |_| |_|  |_|\__,_|_|  \___|_| |_|\___| (_)
 
 -->
-<html>
+<html lang="{{ app.request.locale }}">
```
HeahDude: I don't think we should have it dynamic, because the translator use `fr` as a fallback and we don't support other languages for now. I would hard code "fr" until then.

NyKo24: Soon we start translating on website

HeahDude: For now we don't. An HTML doc which says " my lang is en" but shows "fr" is invalid.

PR #2022, opened 2018-01-16
change meta content

NyKo24:


NyKo24:
Need more work : change meta for AMP version

PR #2025, opened 2018-01-16
Remove IC

NyKo24:


src/Entity/BaseEvent.php
```diff
@@ -334,11 +333,6 @@ public function isFull(): bool
 
     abstract public function getType();
 
-    public function isCitizenInitiative(): bool
-    {
-        return self::CITIZEN_INITIATIVE_TYPE === $this->getType();
```
Remg: is this constant still useful ?

PR #2026, opened 2018-01-16
Redirect already logged in users who try to reach membership form

Remg:


src/Controller/EnMarche/MembershipController.php
```diff
@@ -33,6 +33,10 @@ class MembershipController extends Controller
      */
     public function registerAction(Request $request): Response
     {
+        if ($this->get('security.authorization_checker')->isGranted('IS_AUTHENTICATED_FULLY')) {
+            return $this->redirectToRoute('homepage');
```
HeahDude: Why not redirect to the profile page instead?

Remg: This question was asked. This is a specific request :)

tests/Controller/EnMarche/MembershipControllerTest.php
```diff
@@ -72,6 +72,16 @@ public function provideEmailAddress()
         ];
     }
 
+    public function testCannotCreateMembershipAccountIfAlreadyLoggedIn()
+    {
+        $this->authenticateAsAdherent($this->client, 'gisele-berthoux@caramail.com', 'ILoveYouManu');
+
+        $this->client->request(Request::METHOD_GET, '/inscription');
+
+        $this->assertResponseStatusCode(Response::HTTP_FOUND, $this->client->getResponse());
```
HeahDude: This assertion is needless here since it's done in the following method.

src/Controller/EnMarche/MembershipController.php
Remg: the guru changed his mind \o/

blazarecki:
This will appear in the auth2 PR. We don't need to merge this now.

PR #2029, opened 2018-01-17
[TIMELINE] Add new attribute to search by measure titles + tests/refactoring

Remg:


src/Entity/Timeline/Measure.php
```diff
@@ -267,14 +279,13 @@ public function isDeferred(): bool
...
-    public function saveCurrentThemes(): void
-    {
-        $this->savedThemes = clone $this->themes;
-    }
-
     private function getSavedThemes(): Collection
     {
-        return $this->savedThemes ?? new ArrayCollection();
+        if (!$this->savedThemes) {
+            $this->savedThemes = new ArrayCollection();
```
blazarecki: You don't need to do that if `new ArrayCollection` is in the constructor.

Remg: Not when it comes from Doctrine.

blazarecki: Doctrine give you a `PersistedCollection`.

Remg: Yeah but this is not a collection managed by Doctrine

src/Entity/Timeline/Theme.php
```diff
@@ -120,6 +120,22 @@ public function measureIds(): array
...
     }
 
+    /**
+     * @Algolia\Attribute
+     */
+    public function measureTitles(): array
+    {
+        $titles = [];
+        foreach ($this->measures as $measure) {
+            /* @var $translation MeasureTranslation */
```
HeahDude: Class first

src/Command/ImportTimelineCommand.php
```diff
@@ -254,7 +254,7 @@ private function parseCSV(string $filepath): array
 
     private function getProfiles(): array
     {
-        /* @var $profile Profile */
+        /* @var Profile $profile */
```
illusionOfParadise: `/* @var */` => `/** @var */`

PR #2035, opened 2018-01-19
move CSS & JS call on bottom page

NyKo24:


templates/base.html.twig
```diff
@@ -553,6 +545,12 @@
             <div id="search-engine"></div>
         </div>
 
+        <link rel="stylesheet" type="text/css" href="{{ asset('css/cookieconsent.min.css') }}" />
```
Remg: this stylesheet was declared after `app.css`

NyKo24: ✅

PR #2037, opened 2018-01-19
Generates URL with http or https scheme from console

thomasbisignani:


B-Galati:
Perhaps it would be simpler with https://symfony.com/doc/current/security/force_https.html ?
Combined with the other trick @blazarecki pointed out.

app/config/config.yml
```diff
@@ -27,6 +27,7 @@ parameters:
     app_host:                    '%env(APP_HOST)%'
     amp_host:                    '%env(AMP_HOST)%'
     legislatives_host:           '%env(LEGISLATIVES_HOST)%'
+    router.request_context.scheme:  '%env(APP_SCHEME)%'
```
blazarecki: You can also add `.host`

Remg: No.
this app handles multiple host and each host is defined for each ressource in `routing.yml`

Remg: also, we don't want to force the host for several ressources (like the AssetsController), so that each domain handled by the application has access to these routes on the same domain.

Remg: after reflexion, this could be a good idea to act as a fallback for routes with undefined hosts.

`router.request_context.scheme:  '%env(APP_HOST)%'` would be great indeed

thomasbisignani: Good idea :ok_hand: I think we can make this fallback in another PR.

blazarecki: Or in this PR. It's not a big deal. 📈

thomasbisignani: Done. :ok_hand:

PR #2057, opened 2018-01-24
Auth - Add scope management to OAuth client admin

B-Galati:


B-Galati:
For the record we need this feature for `auth2` branch

PR #2074, opened 2018-01-26
Add checkbox to block committee name modification

illusionOfParadise:


src/Admin/CommitteeAdmin.php
```diff
@@ -152,6 +152,9 @@ protected function configureFormFields(FormMapper $formMapper)
                 ->add('name', null, [
                     'label' => 'Nom',
                 ])
+                ->add('noneditableName', null, [
```
thomasbisignani: What about using `nonEditableName` in camelCase ? or just  `uneditableName`

src/Entity/Committee.php
```diff
@@ -197,6 +202,16 @@ public function setAdminComment(string $adminComment): void
         $this->adminComment = $adminComment;
     }
 
+    public function hasNoneditableName(): ?bool
```
thomasbisignani: I think the `nullable` return type is useless here

src/Entity/Committee.php
```diff
@@ -86,6 +86,11 @@ class Committee extends BaseGroup implements CoordinatorAreaInterface
      */
     private $citizenProjectSupports;
 
+    /**
+     * @ORM\Column(type="boolean", options={"default": false})
+     */
+    private $noneditableName = false;
```
HeahDude: What about a `nameLocked` property with `isNameLocked/setNameLocked`

thomasbisignani: Good idea @HeahDude !

src/Committee/CommitteeFactory.php
```diff
@@ -43,6 +43,7 @@ public function createFromArray(array $data): Committee
             $data['created_at'] ?? 'now'
         );
 
+        $committee->setNameLocked(isset($data['name_locked']) ?: false);
```
Remg: not a big deal but in this case, if `$data['name_locked'] === false`, then `nameLocked` will still be set to true.

i would recommend the following behavior to avoid any future confusion using this method:
`$committee->setNameLocked(isset($data['name_locked']) ? $data['name_locked'] : false);`

HeahDude: Good catch, what about `$data['name_locked'] ?? false`?

PR #2076, opened 2018-01-29
Admin should not use OAuth

B-Galati:
403 them if they try to use OAuth

### DIFF
--- a/features/bootstrap/SecurityContext.php
+++ b/features/bootstrap/SecurityContext.php
@@ -1,7 +1,9 @@
 <?php
 
 use AppBundle\Entity\Adherent;
+use AppBundle\Entity\Administrator;
 use AppBundle\Repository\AdherentRepository;
+use AppBundle\Repository\AdministratorRepository;
 use Behat\Mink\Driver\BrowserKitDriver;
 use Behat\MinkExtension\Context\RawMinkContext;
 use Behat\Symfony2Extension\Context\KernelDictionary;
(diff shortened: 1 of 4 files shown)


src/Controller/OAuth/OAuthServerController.php
```diff
@@ -35,7 +35,7 @@ public function __construct(
     /**
      * @Route("/auth", name="app_front_oauth_authorize")
      * @Method("GET|POST")
-     * @Security("is_granted('IS_AUTHENTICATED_FULLY')")
+     * @Security("is_granted('IS_AUTHENTICATED_FULLY') and !is_granted('ROLE_ADMIN_DASHBOARD')")
```
HeahDude: we usually use `not` with EL.

features/bootstrap/SecurityContext.php
```diff
@@ -22,7 +24,19 @@ public function iAmLoggedAs(string $email): void
...
+        $this->logAs($user);
+    }
+
+    /**
+     * @When I am logged as :email admin
+     */
+    public function iAmLoggedAsAdmin(string $email): void
+    {
+        if (!$user = $this->getAdministratorRepository()->loadUserByUsername($email)) {
+            throw new \Exception(sprintf('Admin %s not found', $email));
```
HeahDude: What about using `webmozart/assert` in beta contexts? It's easier to throw exceptions wit it.

B-Galati: Why not But I would prefer to keep that for later.

PR #2081, opened 2018-01-29
[CP] Added show all in search form

thomasbisignani:


templates/search/_filters_radius.html.twig
```diff
@@ -0,0 +1,11 @@
+<div>
+    dans un rayon de
+    {# this widget will be embedded in the ribbon text #}
+    <select name="{{ constant('PARAMETER_RADIUS', search) }}">
+        {% for option in constant('RADII', search) %}
+            {% if option != 0 %}
```
HeahDude: The `if` can be put after the `for`:
```twig
{% for option in constant('RADII', search) if 0 != option %}
```

templates/search/_filters_radius_citizen_projects.html.twig
```diff
@@ -0,0 +1,9 @@
+<div>
+    <select name="{{ constant('PARAMETER_RADIUS', search) }}">
+        {% for option in constant('RADII', search) %}
```
HeahDude: `0` is allowed here?

thomasbisignani: Yes. For show all Citizen Projects, the value is `0`.

templates/search/_filters_radius.html.twig
thomasbisignani: Nice catch, thx !

src/Search/SearchParametersFilter.php
```diff
@@ -40,6 +40,7 @@ class SearchParametersFilter
         self::TYPE_CITIZEN_ACTIONS,
     ];
 
+    const RADIUS_ALL = 0;
```
blazarecki: What about the `RADIUS_NONE = -1` ?

HeahDude: "None" and "all" are the same in this context, I prefer "ALL" though.

src/Search/SearchParametersFilter.php
```diff
@@ -88,7 +90,13 @@ public function handleRequest(Request $request): self
     {
         $this->setQuery((string) $request->query->get(self::PARAMETER_QUERY));
         $this->setType($request->query->get(self::PARAMETER_TYPE, ''));
-        $this->setRadius($request->query->getInt(self::PARAMETER_RADIUS));
+
+        $radius = null !== $request->query->get(self::PARAMETER_RADIUS) ?
```
HeahDude: Could use `has` here, or store the result. But I would go for:
```php
$radius = $request->query->getInt(self::PARAMETER_RADIUS, self::DEFAULT_RADIUS);
```

src/Search/SearchParametersFilter.php
blazarecki: I don't think so, -1 is for user who don't want to saw citizen project. ping @thomasbisignani ?

thomasbisignani: Ok for `-1` ! Thx.

PR #2084, opened 2018-01-29
Display nice error message when max login attempts limit is reached

B-Galati:


app/config/services/services.xml
```diff
@@ -72,7 +69,6 @@
 
 		<!-- Security -->
 		<prototype namespace="AppBundle\Security\Voter\" resource="../../../src/Security/Voter" />
-		<service id="AppBundle\Security\FailedLoginAttemptFactory"/>
```
HeahDude: Hmm, I though I already did fix that file.

src/Entity/FailedLoginAttempt.php
```diff
@@ -45,13 +45,20 @@ public function __construct(string $signature, array $extra)
 
     public static function createFromRequest(Request $request): self
     {
+        $attempt = LoginAttemptSignature::createFromRequest($request);
+
         return new self(
-            (new FailedLoginAttemptSignature($request->get('_adherent_email'), $request->getClientIp()))(),
+            $attempt(),
```
HeahDude: We're loosing some meaning with this invocation, not a fan.

src/Security/LoginAttemptSignature.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return $this->login;
+    }
+
+    public function getIp(): string
+    {
+        return $this->ip;
+    }
+
+    public function __invoke(): string
```
HeahDude: Please use a meaningful method name here, like `formatArgs` or whatever.

app/config/services/services.xml
B-Galati: auth2 is not up to date so it's possible

src/Entity/FailedLoginAttempt.php
B-Galati: Yeap, replaced with `getSignature` method

app/config/security.yml
```diff
@@ -1,4 +1,7 @@
 security:
+    # Providers auth exceptions must not be hidden to customise the error message
+    # see usages of App\Security\Exception\BadCredentialsException
+    hide_user_not_found: false
```
HeahDude: We need to handle it for the main admin too then, only the main is covered here by modifying the user provider, as far as I can see.

B-Galati: It's ok because admin and user provider used the same class :

```
<service id="app.security.admin_provider" class="AppBundle\Security\UserProvider">
```

HeahDude: Ok I've forgot that, thx!

PR #2091, opened 2018-01-30
Address validity is not needed anymore;user is redirected to adherent homepage after adhesion

B-Galati:


src/Controller/EnMarche/AdherentController.php
```diff
@@ -32,6 +32,16 @@
  */
 class AdherentController extends Controller
 {
+    /**
+     * @Route("/accueil", name="app_adherent_home")
+     * @Method("GET")
+     * @Security("is_granted('ROLE_ADHERENT')")
```
HeahDude: No need thanks to the access control.

src/Form/UpdateMembershipRequestType.php
```diff
@@ -2,30 +2,55 @@
...
-    }
-
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefault('validation_groups', ['Default']);
+        $years = range((int) date('Y') - 15, (int) date('Y') - 120);
+
+        $resolver->setDefaults([
+            'data_class' => MembershipRequest::class,
+            'translation_domain' => false,
```
HeahDude: Why?

src/Membership/MembershipRequest.php
```diff
@@ -12,39 +12,40 @@
 use Symfony\Component\Validator\Constraints as Assert;
 
 /**
- * @AssertUniqueMembership(groups={"Default", "Registration"})
+ * @AssertUniqueMembership(groups={"Registration", "Update"})
```
HeahDude: Do we really need `Update` here? Seems we cannot edit the email.

B-Galati: No thanks

src/Form/UpdateMembershipRequestType.php
B-Galati: All labels are written directly in the view so we don't need to trigger the translator.

HeahDude: I heard they want to use translation soon, I understand the value but this is not worth it anymore.

src/Membership/MembershipRequest.php
B-Galati: In fact yes because we can edit the email when editing user profile.

src/Form/UpdateMembershipRequestType.php
```diff
@@ -2,30 +2,56 @@
...
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefault('validation_groups', ['Default']);
+        $years = range((int) date('Y') - 15, (int) date('Y') - 120);
+
+        $resolver->setDefaults([
+            'data_class' => MembershipRequest::class,
+            'years' => array_combine($years, $years),
+            'validation_groups' => ['Update'],
+            'disabled_name_and_email_fields' => true,
```
HeahDude: `disable_name_and_email_fields` or `name_and_email_fields_disabled`

src/Membership/MembershipRequest.php
HeahDude: I see 👌

src/Membership/MembershipRequest.php
```diff
@@ -76,26 +78,27 @@ class MembershipRequest implements MembershipInterface
     public $comEmail = false;
 
     /**
-     * @Assert\NotBlank(message="common.recaptcha.invalid_message", groups="Registration")
+     * @Assert\NotBlank(message="common.recaptcha.invalid_message", groups={"Registration"})
      * @AssertRecaptcha(groups={"Registration"})
      */
     public $recaptcha;
```
HeahDude: I've done a better handling of the recaptcha on `auth` repo, I'll import it as well to get rid of all these public properties in another PR.

B-Galati: Yeap !

PR #2095, opened 2018-01-31
Added ORM mapping and security checks in the CI and Makefile

HeahDude:


circle.yml
```diff
@@ -56,11 +56,14 @@ test:
             parallel: true
         - ./bin/console doctrine:schema:drop --force --env=test_mysql:
             parallel: true
-        - ./bin/console doctrine:schema:create --env=test_mysql:
+        - ./bin/console doctrine:migrations:migrate -n --env=test_mysql:
```
B-Galati: Could you also update the Makefile to use migrations in test (task `tfp`) instead of schema:create ?

HeahDude: Done.

circle.yml
```diff
@@ -72,6 +75,9 @@ test:
         # Try to build production JS
         - yarn run build-prod
 
+        # Check for vulnerable dependencies
+        - ./vendor/bin/security-checker security:check
```
Remg: 👍 
Should we remove the same check in the `build.sh` script then ?
https://github.com/EnMarche/en-marche.fr/blob/master/scripts/build.sh#L5
I don't see the point of checking it during the build 🤔

HeahDude: Done.

thomasbisignani:
Nice :ok_hand:

PR #2099, opened 2018-01-31
Prefent autofill for password during registration

blazarecki:


templates/membership/register.html.twig
```diff
@@ -57,7 +57,7 @@
                 <li>
                     {{ form_label(form.password,'Mot de passe',{ label_attr: {class: 'visually-hidden'} }) }}
                     {{ form_errors(form.password) }}
-                    {{ form_widget(form.password, {attr: {placeholder: 'Mot de passe'} }) }}
+                    {{ form_widget(form.password, {attr: {placeholder: 'Mot de passe', autocomplete: "new-password"} }) }}
```
illusionOfParadise: It should be `off`, is not it ? https://www.w3schools.com/tags/att_input_autocomplete.asp

HeahDude:
Typo in the commit: `Prefent` => `Prevent`

templates/membership/register.html.twig
blazarecki: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion 

> If an author would like to prevent the autofilling of password fields in user management pages where a user can specify a new password for someone other than themself, autocomplete="new-password" should be specified, though support for this has not been implemented in all browsers yet.

PR #2100, opened 2018-01-31
Send email for adherent on join and not on register

blazarecki:


src/Controller/EnMarche/MembershipController.php
```diff
@@ -85,7 +85,7 @@ public function joinAction(Request $request, EntityManagerInterface $manager): R
         ;
 
         if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
-            $this->get('app.membership_request_handler')->update($user, $membership);
+            $this->get('app.membership_request_handler')->join($user, $membership);
 
             $user->join();
```
B-Galati: Now it belongs to the Join method of the handler.

src/Membership/MembershipRequestHandler.php
```diff
@@ -43,17 +43,26 @@ public function __construct(
...
+    public function join(Adherent $adherent, MembershipRequest $membershipRequest)
+    {
+        $adherent->updateMembership($membershipRequest, $this->addressFactory->createFromAddress($membershipRequest->getAddress()));
+        $token = AdherentActivationToken::generate($adherent);
+
         $this->manager->persist($token);
         $this->manager->flush();
 
         $activationUrl = $this->generateMembershipActivationUrl($adherent, $token);
         $this->mailer->sendMessage(AdherentAccountActivationMessage::createFromAdherent($adherent, $activationUrl));
```
B-Galati: I think this message should be in `handle`

HeahDude: The naming has lost all its sense, `membership`, `AdherentAccountActivationMessage`,  ...

B-Galati: Agree

blazarecki:
@HeahDude @B-Galati can you review this PR ? Thanks.

src/Membership/MembershipRequestHandler.php
```diff
@@ -52,10 +60,25 @@ public function handle(MembershipRequest $membershipRequest)
...
         $this->dispatcher->dispatch(UserEvents::USER_CREATED, new UserEvent($adherent));
     }
 
+    public function join(Adherent $user, MembershipRequest $membershipRequest)
+    {
+        $user->updateMembership($membershipRequest, $this->addressFactory->createFromAddress($membershipRequest->getAddress()));
+        $user->join();
+        $this->manager->flush();
+
+        $this->mailer->sendMessage(AdherentAccountConfirmationMessage::createFromAdherent(
```
HeahDude: Could you send the message with a subscriber to `AdherentAccountWasCreatedEvent` instead? It would avoid adding the managers as deps here.

PR #2105, opened 2018-01-31
Don’t show Profil and My Activities for simple user

blazarecki:


src/DataFixtures/ORM/LoadUserData.php
```diff
@@ -0,0 +1,53 @@
...
+            'password' => 'enmarche',
+            'email' => 'simple-user@example.ch',
+            'gender' => 'female',
+            'first_name' => 'Simple',
+            'last_name' => 'User',
+            'address' => PostAddress::createForeignAddress('CH', '8057', 'Zürich', '32 Zeppelinstrasse', 47.3950786, 8.5361402),
+            'birthdate' => '1972-11-24',
+        ]);
+
+        PropertyReplacer::replaceProperty($user1, 'adherent', false);
```
B-Galati: Overkill, just add a parameters to the factory to disable the `join()` call in `createFromArray` method.

HeahDude: Why? we should not break the encapsulation of the entity because of this line in fixtures loading.

B-Galati: `\AppBundle\Membership\AdherentFactory::createFromArray` is only used for tests (which questionable too). So we can just add a new parameters to this method to conditionally called `$adherent->join();`. It's not breaking encapsulation.

HeahDude: >is only used for tests (which is questionable)
Yup, like many factories, this is in my list of improvements ;).

blazarecki: I fix it, but for me we can break the encapsulation for testing purpose.

B-Galati: Sure if you have a very good reason 👍

PR #2107, opened 2018-01-31
Add welcome message on adhesion form

B-Galati:


src/Form/UpdateMembershipRequestType.php
```diff
@@ -16,17 +16,21 @@ class UpdateMembershipRequestType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        if (!$options['is_adhesion_form']) {
```
HeahDude: What about a `UpdateUserType` and another that extends it to add this three fields?

B-Galati: Yeap let's do this. I will create a join type.

src/Controller/EnMarche/MembershipController.php
```diff
@@ -79,8 +79,9 @@ public function joinAction(Request $request, EntityManagerInterface $manager): R
             throw $this->createNotFoundException();
         }
 
+        $fromActivation = $request->query->getBoolean('from_activation');
         $membership = MembershipRequest::createFromAdherent($user);
-        $form = $this->createForm(UpdateMembershipRequestType::class, $membership)
+        $form = $this->createForm(UpdateMembershipRequestType::class, $membership, ['is_adhesion_form' => true])
```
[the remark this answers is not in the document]
B-Galati: Yea this is wrong >-<

src/Form/UpdateMembershipRequestType.php
B-Galati: done

src/Form/BecomeAdherentType.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace AppBundle\Form;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class BecomeAdherentType extends AbstractType
```
HeahDude: A membership is about becoming adherent. Please keep this one as `MembershipRequestType`.

src/Form/BecomeAdherentType.php
```diff
@@ -0,0 +1,24 @@
...
+            ->remove('firstName')
+            ->remove('lastName')
+            ->remove('emailAddress')
+            ->remove('position')
+        ;
+    }
+
+    public function getParent()
+    {
+        return MembershipRequestType::class;
```
HeahDude: Please rename this one instead, something like `UserProfileType` or `User...whatever`

B-Galati: Ok but it makes more sense to call it `AdherentType` as it's about modifying an `Adherent`.

src/Form/BecomeAdherentType.php
B-Galati: Ok

B-Galati: Just did it and I don't like it. It's more confusing IMO.
I understand your point but from an external point of view this name makes more sense.
It is really obvious to understand and it's the main goal here no ?
You can become an adherent with this form, that's it.
What do you think ?

HeahDude: `AdherentType` <3. Thanks!

src/Form/RegistrationType.php
```diff
@@ -12,7 +12,7 @@
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class NewMemberShipRequestType extends AbstractType
+class RegistrationType extends AbstractType
```
HeahDude: `UserRegistrationType`?

PR #2125, opened 2018-02-02
Fix migration updated fields

B-Galati:


app/migrations/Version20180131163109.php
```diff
@@ -11,7 +11,7 @@ class Version20180131163109 extends AbstractMigration
     public function up(Schema $schema)
     {
         foreach ($this->connection->executeQuery('SELECT id, first_name, last_name FROM adherents') as $adherent) {
-            $this->connection->executeQuery(
+            $this->addSql(
```
HeahDude: There will be too many output, maybe `executeUpdate` then?

PR #2128, opened 2018-02-02
Block modification of postal code/city/country fields if committee is valid

illusionOfParadise:


src/Form/AddressType.php
```diff
@@ -49,6 +54,7 @@ public function configureOptions(OptionsResolver $resolver)
         $resolver->setDefaults([
             'data_class' => Address::class,
             'error_bubbling' => false,
+            'disable_fields' => false,
         ]);
```
Remg: Should we add: `$resolver->setAllowedTypes('disable_fields', 'bool');` ?

src/Form/AddressType.php
```diff
@@ -46,9 +51,14 @@ function ($value) {
...
-            'error_bubbling' => false,
-        ]);
+        $resolver
+            ->setDefaults([
+                'data_class' => Address::class,
+                'disable_fields' => false,
+                'error_bubbling' => false,
+            ])
+            ->setAllowedTypes('disable_fields', 'bool')
+            ->setAllowedTypes('error_bubbling', 'bool')
```
HeahDude: To remove.

PR #2132, opened 2018-02-05
Remove City autocomplete feature

B-Galati:


src/Address/Address.php
```diff
@@ -31,13 +31,14 @@ class Address implements AddressInterface, GeocodableInterface
 
     /**
      * @Assert\Length(max=15, groups={"Default", "Update"})
-     * @Assert\NotBlank(groups={"Update"})
+     * @Assert\Expression(expression="this.getCountry() !== 'FR' or value", message="common.city_name.not_blank", groups={"Update"})
+     *
```
Remg: extra line

B-Galati: fixed

src/Address/Address.php
```diff
@@ -31,13 +31,13 @@ class Address implements AddressInterface, GeocodableInterface
 
     /**
      * @Assert\Length(max=15, groups={"Default", "Update"})
-     * @Assert\NotBlank(groups={"Update"})
+     * @Assert\Expression(expression="this.getCountry() !== 'FR' or value", message="common.city_name.not_blank", groups={"Update"})
```
Remg: should we use the constant `self::FRANCE` for the `FR` value ?

src/Address/Address.php
```diff
@@ -31,13 +31,13 @@ class Address implements AddressInterface, GeocodableInterface
...
      */
     private $city;
 
     /**
      * @Assert\Length(max=255, groups={"Default", "Update"})
-     * @Assert\NotBlank(message="common.city_name.not_blank", groups={"Update"})
+     * @Assert\Expression(expression="this.getCountry() === 'FR' or value", message="common.city_name.not_blank", groups={"Update"})
```
Remg: same here

B-Galati: fixed

src/Address/Address.php
B-Galati: fixed

PR #2133, opened 2018-02-05
[ADMIN] Add postal code filtering by multiple comma separated value

illusionOfParadise:


src/Admin/EventAdmin.php
```diff
@@ -197,8 +197,14 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                         return;
                     }
 
-                    $qb->andWhere(sprintf('%s.postAddress.postalCode', $alias).' LIKE :postalCode');
-                    $qb->setParameter('postalCode', $value['value'].'%');
+                    $value = array_map('trim', explode(',', strtolower($value['value'])));
+                    $idExpression = $qb->expr()->orX();
```
B-Galati: Why `idExpression` ?

src/Admin/AdherentAdmin.php
```diff
@@ -300,8 +300,14 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                         return;
                     }
 
-                    $qb->andWhere(sprintf('%s.postAddress.postalCode', $alias).' LIKE :postalCode');
-                    $qb->setParameter('postalCode', $value['value'].'%');
+                    $value = array_map('trim', explode(',', strtolower($value['value'])));
```
ottaviano: what about move this code into a class (ex `PostalCodeQueryBuilderUtils`) with static method for doing the stuff for avoiding the duplication?
`'callback' => [PostalCodeQueryBuilderUtils::class, 'doStuff']`

Remg: in this case a custom `AdminExtension` should be appropriate: 
https://symfony.com/doc/master/bundles/SonataAdminBundle/reference/extensions.html

src/Admin/EventAdmin.php
Remg: like our lord and savior @HeahDude once said: "copy pasted with the ass."

src/Admin/Extension/PostalCodeFilterAdminExtension.php
```diff
@@ -0,0 +1,39 @@
...
+    public function configureDatagridFilters(DatagridMapper $datagridMapper)
+    {
+        $datagridMapper
+            ->add('postalCode', CallbackFilter::class, [
+                'label' => 'Code postal (préfixe)',
+                'show_filter' => true,
+                'field_type' => TextType::class,
+                'callback' => function (ProxyQuery $qb, string $alias, string $field, array $value) {
+                    if (!$value['value']) {
+                        return;
```
Remg: `return false;`
This indicates whether the filter is applied on this field or not

PR #2137, opened 2018-02-06
Fixes for auth

B-Galati:


src/Membership/MembershipRequest.php
```diff
@@ -106,9 +105,6 @@ class MembershipRequest implements MembershipInterface
 
     public function __construct()
     {
-        $this->gender = Genders::MALE;
```
ottaviano: maybe assign the default values on property?
`private $gender = Genders::MALE`?

src/Membership/MembershipRequest.php
```diff
@@ -165,7 +161,7 @@ public function setEmailAddress(string $emailAddress): void
 
     public function getEmailAddress(): string
     {
-        return $this->emailAddress;
+        return $this->emailAddress ?: '';
```
ottaviano: `?? '';`

B-Galati: Why, in this case it's exactly the same thing no ?

src/Membership/MembershipRequest.php
B-Galati: We don't want a default value anymore in this case.

src/Membership/MembershipRequest.php
ottaviano: yes, in this case it do the same job, just for using php7

PR #2140, opened 2018-02-06
[ADMIN] Use an AdminExtension to synchronise AMP content

Remg:


src/Form/EventListener/AmpSynchronisedListener.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return [
+            FormEvents::SUBMIT => 'setAmpContent',
+        ];
+    }
+
+    public function setAmpContent(FormEvent $event): void
+    {
+        /** @var EntityContentInterface $entityContent */
+        if (!$entityContent = $event->getData()) {
```
ottaviano: `if (!$entityContent instanceof EntityContentInterface) { return; }`

app/config/services/admin.xml
```diff
@@ -577,6 +561,14 @@
 
         <service autowire="true" public="false" id="AppBundle\Form\DataTransformer\CommitteeTransformer"/>
 
+        <service id="app.admin.amp_synchronised_extension" class="AppBundle\Admin\Extension\AmpSynchronisedAdminExtension">
+            <argument type="service" id="AppBundle\Form\EventListener\AmpSynchronisedListener" />
+
+            <tag name="sonata.admin.extension" target="app.admin.article" />
```
B-Galati: Clarfication admin is missing no ?

Remg: this admin is not declared at all. i guess we have to remove it completely :)

thomasbisignani:
:ok_hand:

app/config/services/admin.xml
Remg: @guillaumedenhez confirmed: we should remove the `ClarificationAdmin` code for now as it is useless.

PR #2142, opened 2018-02-07
[2726] User must not subscribe to email notification by default

B-Galati:


app/config/assets_version.yml
```diff
@@ -1,2 +0,0 @@
-parameters:
-    assets_hash: d4b2bdcf09071f80c913
```
B-Galati: This file is generated automatically by `make assets-prod` and should not be committed. 
So I removed it and is now only required for prod and test.

B-Galati:
The changes on `var/cache/.gitkeep` are wanted too. This file is changing everytime you `make perm`. And it's not important if this one is 777.

src/Controller/EnMarche/CommitteeController.php
```diff
@@ -72,6 +73,7 @@ public function timelineAction(Request $request, Committee $committee): Response
         return $this->render('committee/timeline/feed.html.twig', [
             'committee' => $committee,
             'committee_timeline' => $timeline,
+            'has_role_adherent' => $this->getUser() instanceof Adherent ? $this->getUser()->isAdherent() : false,
```
ottaviano: or without terner: 
```
=> $this->getUser() instanceof Adherent && $this->getUser()->isAdherent(),
```

src/Twig/AssetRuntime.php
```diff
@@ -16,13 +16,17 @@ class AssetRuntime
...
-    public function __construct(Router $router, BaseAssetExtension $symfonyAssetExtension, string $secret, string $env, string $hash)
+    public function __construct(Router $router, BaseAssetExtension $symfonyAssetExtension, string $secret, string $env, ?string $hash)
     {
         $this->router = $router;
         $this->symfonyAssetExtension = $symfonyAssetExtension;
         $this->secret = $secret;
         $this->env = $env;
         $this->hash = $hash;
+
+        if ('dev' !== $env && !$this->hash) {
```
ottaviano: not fan of this.. it's just for deleting `assets_version.yml` file or for fixing a bug?

B-Galati: Just to be sure prod env is configured with a proper hash since the default value is now `null`.

src/Controller/EnMarche/CommitteeController.php
B-Galati: fixed

src/Twig/AssetRuntime.php
B-Galati: It's a really bad idea to have an `assets_version.yml` with a real value in the repo.. It could hide some real problem. Like this it's less error prone at least.

ottaviano: normally, the value of hash change on each build in production, isn't it?

B-Galati: yeap

src/Membership/AdherentFactory.php
```diff
@@ -60,15 +60,14 @@ public function createFromArray(array $data): Adherent
             $data['address'],
             $phone,
             Adherent::DISABLED,
-            isset($data['registered_at']) ? $data['registered_at'] : 'now'
+            isset($data['registered_at']) ? $data['registered_at'] : 'now',
```
Remg: `$data['registered_at'] ?? 'now'`

PR #2143, opened 2018-02-07
Fix some problemes with committee description

cacahouwete:
- fix display description in committee when there are apostrophe
- fix description edition in BO

### DIFF
--- a/src/Entity/Committee.php
+++ b/src/Entity/Committee.php
@@ -55,7 +55,7 @@ class Committee extends BaseGroup implements CoordinatorAreaInterface
      *
      * @Algolia\Attribute
      */
-    private $description;
+    protected $description;
 
     /**
      * The committee Facebook page URL.
--- a/src/Form/CommitteeCommandType.php
+++ b/src/Form/CommitteeCommandType.php
@@ -26,7 +26,6 @@ class CommitteeCommandType extends AbstractType
             ])
             ->add('description', TextareaType::class, [
                 'filter_emojis' => true,
-                'purify_html' => true,
             ])
             ->add('address', AddressType::class, [
                 'disable_fields' => $committee ? $committee->isApproved() : false,


src/Entity/Committee.php
```diff
@@ -205,6 +205,13 @@ public function getDescription(): string
         return $this->description;
     }
 
+    public function setDescription(string $description): self
+    {
+        $this->description = $description;
+
+        return $this;
```
ottaviano: remove this action pls, it's not a fluent interface here.

src/Entity/Committee.php
```diff
@@ -205,6 +205,13 @@ public function getDescription(): string
         return $this->description;
     }
 
+    public function setDescription(string $description): self
```
Remg: Why is this setter needed?

src/Form/CommitteeCommandType.php
```diff
@@ -26,7 +26,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ])
             ->add('description', TextareaType::class, [
                 'filter_emojis' => true,
-                'purify_html' => true,
```
ottaviano: why do you remove this? `purify_html` filter the content of the field and keep whitelisted html tags.

src/Entity/Committee.php
Remg: We don't use the fluent interface on this project.
Please return `void` on setters.

src/Entity/Committee.php
cacahouwete: Because if we don t have a setter we can't modify the description with an admin form

src/Form/CommitteeCommandType.php
cacahouwete: Because the content is display without raw and it cause some problems with aposthrophe

src/Entity/Committee.php
HeahDude: No, take a look at https://github.com/EnMarche/en-marche.fr/blob/develop/src/Entity/EntityCrudTrait.php.

src/Form/CommitteeCommandType.php
ottaviano: maybe you can just change the flag [here](https://github.com/EnMarche/en-marche.fr/blob/develop/src/Utils/HtmlPurifier.php#L25)
what do others think?

src/Entity/Committee.php
B-Galati: 😱

src/Form/CommitteeCommandType.php
B-Galati: @illusionOfParadise Don't we had the same sort of issue somewhere else in the past ?

illusionOfParadise: Yes, I did the same thing for citizen project.

cacahouwete: @ottaviano change the flag can maybe impact too many form no ? :s

PR #2155, opened 2018-02-08
[ch2705] Dynamic modification of articles URL after publications.

cacahouwete:
- Creation of doctrine listener to handle all object modifications
- Creation of Redirection Manager to handle redirect with cache for performances
- add some test to trigger and test redirection

https://app.clubhouse.io/enmarche/story/2705/pouvoir-modifier-les-url-des-articles-apr%C3%A8s-publication

### DIFF
--- a/src/Admin/ArticleAdmin.php
+++ b/src/Admin/ArticleAdmin.php
@@ -43,11 +43,6 @@ class ArticleAdmin extends AbstractAdmin
 
     protected function configureFormFields(FormMapper $formMapper)
     {
-        $slugEditable =
-            null === $this->getSubject()->getTitle()   // Creation
-            || !$this->getSubject()->isPublished()     // Draft
-        ;
-
         $formMapper
             ->with('Méta-données', ['class' => 'col-md-8'])
                 ->add('title', TextType::class, [
@@ -89,8 +84,7 @@ class ArticleAdmin extends AbstractAdmin
                 ])
                 ->add('slug', null, [
                     'label' => 'URL de publication',
-                    'disabled' => !$slugEditable,
-                    'help' => $slugEditable ? 'Ne spécifier que la fin : http://en-marche.fr/articles/[votre-valeur]<br />Doit être unique' : 'Non modifiable car publié',
+                    'help' => 'Ne spécifier que la fin : http://en-marche.fr/articles/[votre-valeur]<br />Doit être unique',
                 ])
                 ->add('category', null, [
                     'label' => 'Catégorie de publication',
--- a/src/DataFixtures/ORM/LoadArticleData.php
+++ b/src/DataFixtures/ORM/LoadArticleData.php
@@ -3,17 +3,13 @@
 namespace AppBundle\DataFixtures\ORM;
 
 use AppBundle\Entity\ArticleCategory;
-use Doctrine\Common\DataFixtures\FixtureInterface;
+use Doctrine\Bundle\FixturesBundle\Fixture;
 use Doctrine\Common\Persistence\ObjectManager;
 use Faker\Factory;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\File\File;
 
-class LoadArticleData implements FixtureInterface, ContainerAwareInterface
+class LoadArticleData extends Fixture
 {
-    use ContainerAwareTrait;
-
     public function load(ObjectManager $manager)
     {
         $faker = Factory::create('fr_FR');
--- a/src/DataFixtures/ORM/LoadOrderArticleData.php
+++ b/src/DataFixtures/ORM/LoadOrderArticleData.php
@@ -2,18 +2,13 @@
 
 namespace AppBundle\DataFixtures\ORM;
 
-use Doctrine\Common\DataFixtures\AbstractFixture;
-use Doctrine\Common\DataFixtures\FixtureInterface;
+use Doctrine\Bundle\FixturesBundle\Fixture;
 use Doctrine\Common\Persistence\ObjectManager;
 use Faker\Factory;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\File\File;
 
-class LoadOrderArticleData extends AbstractFixture implements FixtureInterface, ContainerAwareInterface
+class LoadOrderArticleData extends Fixture
 {
-    use ContainerAwareTrait;
-
     public function load(ObjectManager $manager)
     {
         $faker = Factory::create('fr_FR');
@@ -60,5 +55,10 @@ class LoadOrderArticleData extends AbstractFixture implements FixtureInterface, 
         ]));
 
         $manager->flush();
+    }
+
+    public function getDependencies()
+    {
+        return [LoadOrderSectionData::class];
     }
 }
--- a/src/EntityListener/ArticleListener.php
+++ b/src/EntityListener/ArticleListener.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace AppBundle\EntityListener;
+
+use AppBundle\Entity\Article;
+use AppBundle\Redirection\Dynamic\RedirectionManager;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
+use Symfony\Bundle\FrameworkBundle\Routing\Router;
+
+class ArticleListener
+{
+    private $redirectionManager;
+    private $router;
+
+    /**
+     * @var array[]
+     */
+    private $redirections = [];
+
+    public function __construct(RedirectionManager $redirectionManager, Router $router)
+    {
+        $this->redirectionManager = $redirectionManager;
+        $this->router = $router;
+    }
+
+    public function preUpdate(Article $article, PreUpdateEventArgs $preUpdateEventArgs): void
+    {
+        if (!$article->isPublished()
+            || (!$preUpdateEventArgs->hasChangedField('slug')
+                && !$preUpdateEventArgs->hasChangedField('category'))
+        ) {
+            return;
+        }
+
+        $articleSlugNew = $articleSlugOld = $article->getSlug();
+        if ($preUpdateEventArgs->hasChangedField('slug')) {
+            $articleSlugOld = $preUpdateEventArgs->getOldValue('slug');
+            $articleSlugNew = $preUpdateEventArgs->getNewValue('slug');
+        }
+        $categoryNew = $categoryOld = $article->getCategory();
+        if ($preUpdateEventArgs->hasChangedField('category')) {
+            $articleSlugOld = $preUpdateEventArgs->getOldValue('category');
+            $articleSlugNew = $preUpdateEventArgs->getNewValue('category');
+        }
+
+        $this->redirections[$article->getId()] = [
+            'source' => $this->router->generate('article_view',
+                [
+                    'articleSlug' => $articleSlugOld,
+                    'categorySlug' => $categoryOld->getSlug(),
+                ]),
+            'target' => $this->router->generate('article_view',
+                [
+                    'articleSlug' => $articleSlugNew,
+                    'categorySlug' => $categoryNew->getSlug(),
+                ]),
+        ];
+    }
+
+    public function postUpdate(Article $article): void
+    {
+        $redirection = $this->redirectionManager->setRedirection(
+            $this->redirections[$article->getId()]['source'],
+            $this->redirections[$article->getId()]['target']
+        );
+
+        $this->redirectionManager->optimiseRedirection($redirection);
+    }
+}
--- a/src/Redirection/Dynamic/RedirectToAdminPanelHandler.php
+++ b/src/Redirection/Dynamic/RedirectToAdminPanelHandler.php
@@ -2,7 +2,6 @@
 
 namespace AppBundle\Redirection\Dynamic;
 
-use AppBundle\Repository\RedirectionRepository;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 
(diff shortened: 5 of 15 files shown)


src/Redirection/Dynamic/RedirectToAdminPanelHandler.php
```diff
@@ -11,18 +11,16 @@
  */
 class RedirectToAdminPanelHandler implements RedirectToInterface
 {
-    private $provider;
-    private $redirectionRepository;
+    private $redirectionManager;
 
-    public function __construct(RedirectionsProvider $provider, RedirectionRepository $redirectionRepository)
+    public function __construct(RedirectionManager $redirectionManager, RedirectionRepository $redirectionRepository)
```
B-Galati: Let's remove the repository too ;-)

src/Redirection/Dynamic/RedirectionManager.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->cache = $cache;
+        $this->entityManager = $entityManager;
+        $this->serializer = $serializer;
+    }
+
+    public function refreshRedirectionCache(Redirection $redirection): void
+    {
+        $item = $this->cache
+            ->getItem('redirection-'.md5($redirection->getFrom()))
```
B-Galati: The prefixed is not required if you have a dedicated cache pool. And I don't think the `md5` is needed. The key of the item could be any string I guess.

src/Redirection/Dynamic/RedirectionManager.php
```diff
@@ -0,0 +1,66 @@
...
+            ->set($this->serializer->serialize($redirection, 'json'))
```
B-Galati: What do you think if we only serialize the needed information to perform the redirection ? Instead of the whole object ?

cacahouwete: In the object we have 3/5 attributes used for the redirection. I can configure a serialization group if you think it's important.

B-Galati: Yeap it would be more accurate 👍

src/Redirection/Dynamic/RedirectionManager.php
```diff
@@ -0,0 +1,66 @@
...
+        if ($redirection = $this->entityManager->getRepository(Redirection::class)->findOneByOriginUri($source)) {
+            $this->entityManager->remove($redirection);
+            $this->entityManager->flush();
+            $this->cache->deleteItem(md5($source));
+        }
+    }
+
+    public function getRedirection(string $source): ?Redirection
+    {
+        if (($item = $this->cache->getItem(md5($source))) && $item->isHit()) {
```
Remg: `&& $item->isHit()` : here `$item` is not assigned yet and therefore is undefined.

This should be:
```
$item = $this->cache->getItem(md5($source));

if ($item && $item->isHit()) {
```

Remg: my bad. i did not see the extra brackets 🤕

src/DataFixtures/ORM/LoadOrderArticleData.php
```diff
@@ -2,19 +2,14 @@
...
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\File\File;
 
-class LoadOrderArticleData extends AbstractFixture implements FixtureInterface, ContainerAwareInterface
+class LoadOrderArticleData extends Fixture
 {
-    use ContainerAwareTrait;
-
-    public function load(ObjectManager $manager)
+    public function load(ObjectManager $manager): void
```
Remg: method signature should not change.

src/DataFixtures/ORM/LoadOrderArticleData.php
```diff
@@ -61,4 +56,9 @@ public function load(ObjectManager $manager)
 
         $manager->flush();
     }
+
+    public function getDependencies(): array
```
Remg: method signature should not change.

src/DataFixtures/ORM/LoadArticleData.php
```diff
@@ -3,18 +3,14 @@
...
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\File\File;
 
-class LoadArticleData implements FixtureInterface, ContainerAwareInterface
+class LoadArticleData extends Fixture
 {
-    use ContainerAwareTrait;
-
-    public function load(ObjectManager $manager)
+    public function load(ObjectManager $manager): void
```
Remg: method signature should not change. (and will break as per PHP ^7.2)

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+use AppBundle\DataFixtures\ORM\LoadAdminData;
+use AppBundle\DataFixtures\ORM\LoadArticleData;
+use AppBundle\Entity\Article;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Tests\AppBundle\Controller\ControllerTestTrait;
+use Tests\AppBundle\MysqlWebTestCase;
+
+class ArticleAdminTest extends MysqlWebTestCase
```
Remg: Please add the `functional` tag for functional tests.

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+            $prefix.'[slug]' => 'outre-mer-new',
+            $prefix.'[description]' => 'Vous devez saisir au moins 10 caractères.',
+        ]);
+        $this->client->submit($form);
+
+        $this->assertStatusCode(Response::HTTP_FOUND, $this->client);
+
+        $this->client->request(
+            'GET',
+            $url = $this->getUrl('article_view', [
```
Remg: `$url` seems useless

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+                'articleSlug' => $article->getSlug(),
+                'categorySlug' => $article->getCategory()->getSlug(),
+            ])
+        );
+
+        $this->assertClientIsRedirectedTo('/articles/actualites/outre-mer-new', $this->client, false, true);
+
+        $this->client->request(
+            'GET',
+            $url = $this->getUrl('amp_article_view', [
```
Remg: `$url` seems useless

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+class ArticleAdminTest extends MysqlWebTestCase
+{
+    use ControllerTestTrait;
+
+    public function testCreateCategoryFail(): void
+    {
+        $this->authenticateAsAdmin($this->client, 'superadmin@en-marche-dev.fr', 'superadmin');
+
+        $crawler = $this->client->request('GET', $this->getUrl('admin_app_article_create'));
```
Remg: Please use `Symfony\Component\HttpFoundation\Request::METHOD_GET`
(and same for all such calls below)

Remg: Please hardcode urls in functional tests.

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+        $this->authenticateAsAdmin($this->client, 'superadmin@en-marche-dev.fr', 'superadmin');
+
+        /** @var EntityManagerInterface $em */
+        $em = $this->get('doctrine')->getManager();
+        /** @var Article $article */
+        $article = $em->getRepository(Article::class)->findOneBySlug('outre-mer');
+
+        $this->client->request(
+            'GET',
+            $this->getUrl('article_view', [
```
Remg: Please hardcode urls in functional tests.
(same for such calls below)

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+                'articleSlug' => $article->getSlug(),
+                'categorySlug' => $article->getCategory()->getSlug(),
+            ])
+        );
+
+        $this->assertStatusCode(Response::HTTP_OK, $this->client);
+
+        $crawler = $this->client->request(
+            'GET',
+            $this->getUrl('admin_app_article_edit', ['id' => $article->getId()])
```
Remg: `sprintf('/admin/app/article/%s/edit', $article->getId())`

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,121 @@
...
+            $this->client->getContainer()
+        );
+    }
+
+    public function testEditSlugToTriggerRedirectionListener(): void
+    {
+        $this->authenticateAsAdmin($this->client, 'superadmin@en-marche-dev.fr', 'superadmin');
+
+        /** @var EntityManagerInterface $em */
+        $em = $this->get('doctrine')->getManager();
```
Remg: you can use `$this->manager`, thanks to the `init()` method

src/DataFixtures/ORM/LoadOrderArticleData.php
cacahouwete: Good to know

src/Admin/ArticleAdmin.php
```diff
@@ -91,7 +91,6 @@ protected function configureFormFields(FormMapper $formMapper)
                 ])
                 ->add('slug', null, [
                     'label' => 'URL de publication',
-                    'disabled' => !$slugEditable,
                     'help' => $slugEditable ? 'Ne spécifier que la fin : http://en-marche.fr/articles/[votre-valeur]<br />Doit être unique' : 'Non modifiable car publié',
```
Remg: This help message should always be the same now that the field is never disabled

tests/Admin/ArticleAdminTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->client->request(
+            Request::METHOD_GET,
+            sprintf('//'.$this->hosts['app'].'/articles/%s/%s', $article->getCategory()->getSlug(), $article->getSlug())
+        );
+
+        $this->assertStatusCode(Response::HTTP_OK, $this->client);
+
+        $this->client->request(
+            Request::METHOD_GET,
+            sprintf('//'.$this->hosts['amp'].'/articles/%s/%s', $article->getCategory()->getSlug(), $article->getSlug())
```
Remg: Let's create a dedicated client for  `amp` subdomain. e.g.: 
`
$ampClient = $this->makeClient(false, ['HTTP_HOST' => $this->hosts['amp']]);`

This way we can get rid of every host declaration in urls, and network-like paths (starting  with `//`)

Makefile
```diff
@@ -133,7 +133,7 @@ tf: tfp
 	$(EXEC) vendor/bin/phpunit --group functional || true
 
 tfp:            ## Prepare the PHP functional tests
-tfp: vendor assets-amp assets-prod
+tfp: vendor assets-amp assets-prod up
```
B-Galati: Please remove this. It should be done only once while `tfp` task can be done a lot of time per day.

src/Redirection/Dynamic/RedirectionManager.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->cache = $cache;
+        $this->entityManager = $entityManager;
+        $this->serializer = $serializer;
+    }
+
+    public function refreshRedirectionCache(Redirection $redirection): void
+    {
+        $item = $this->cache
+            ->getItem(md5($redirection->getFrom()))
```
B-Galati: I think we can remove the `md5`. IMO it's not needed. Any reason you added it ?

cacahouwete: False we need it because of special chars like /? and some others reserved

src/EntityListener/ArticleListener.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function __construct(RedirectionManager $redirectionManager, Router $router)
+    {
+        $this->redirectionManager = $redirectionManager;
+        $this->router = $router;
+    }
+
+    public function preUpdate(Article $article, PreUpdateEventArgs $preUpdateEventArgs): void
+    {
+        if (!$article->isPublished()
```
ottaviano: if the article isn't published you'll creat one redirect entry wich the same URL for `source` and `target`, why?

src/Admin/ArticleAdmin.php
```diff
@@ -91,8 +86,7 @@ protected function configureFormFields(FormMapper $formMapper)
                 ])
                 ->add('slug', null, [
                     'label' => 'URL de publication',
-                    'disabled' => !$slugEditable,
```
ottaviano: prenons un exemple: article publié donc avec une URL (slug+category) indexée par les moteurs de recherche, cachée par des proxies, browsers et qui se trouve ds les bookmakers des gens. 
comment tu fais si l'admin modifie sa categorie, sauvegarde et ensuite modifie son slug et sauvegarde a nouveau? 
Pendant ces deux sauvegardes la toute premiere URL sera écrasée si je comprends bien le code.
Ne doit-on pas plutôt remplacer uniquement la `target` et laisser la `source` initiale d'une redirection?

cacahouwete: Je te prend un autre exemple ^^, x modifi le slug puis 2j plus tard il modifi la catégory, de fait la nouvelle URL commence aussi a etre indexé donc il faut une redirect de l ancienne et l ancienne de l ancienne , mais après j suis d'accord sur le fait qu il faudrait changer l ancienne redirect sur la nouvelle en plus de mettre une nouvelle redirect pour une question d'optimisation ^^,

ottaviano: c'est complètement faux ce que je dis, tu sauvegarde bien toutes les redirections 👍

src/Redirection/Dynamic/RedirectionManager.php
B-Galati: It's not very readable though. I prefer the way @Remg suggested.

src/Redirection/Dynamic/RedirectionManager.php
```diff
@@ -0,0 +1,66 @@
...
+            $this->entityManager->remove($redirection);
+            $this->entityManager->flush();
+            $this->cache->deleteItem(md5($source));
+        }
+    }
+
+    public function getRedirection(string $source): ?Redirection
+    {
+        if (($item = $this->cache->getItem(md5($source))) && $item->isHit()) {
+            $redirection = $this->serializer->deserialize($item->get(), Redirection::class, 'json');
```
B-Galati: For readability, you can return directly here then remove the else statement

B-Galati:
@cacahouwete I see 2 others points :
- I changed the slug of an article, then I changed it back to the old value (I did it by mistake). Does the app would behave correctly ?
- What about adding all redirections in a list in the Admin view ? It would be cool for everybody no ? (perhaps it's alreay possible through another admin, in that case nevermind)

Remg:
@B-Galati are you talking about `/admin/app/redirection/list` ?

B-Galati:
@Remg Sorry I was not precise enough. I talk about adding them in the article Admin directly so that we can see this very important information directly. But if we can see these from `/admin/app/redirection/list` with any kind of filter, it should be ok too (even if less ergonomic)

Remg:
@B-Galati i see what you mean :)
it could be a good idea, but maybe not a priority.

B-Galati:
@Remg agree, not a priority ;-)

B-Galati:
@cacahouwete Awesome, thank you !

PR #2160, opened 2018-02-09
Add some fixes proposals to docker-compose.yml.dist

B-Galati:


docker-compose.override.yml.dist
```diff
@@ -11,6 +11,9 @@ version: "2"
 #    app:
 #        volumes:
 #            - "~/.composer:/root/.composer"
+# Add this 2 volumes if you have weird permissions issues (But you won't gonna be able to see your cache on your host)
+#            - /app/var/cache
```
ottaviano: it works? it seems to be a named volume that doesn't exists

B-Galati: It creates a volume from this folder when starting, the result is that the folder is not mapped to file system anymore.

B-Galati:
@thomasbisignani @Remg @ottaviano I tested it once more and it's not working as I expected. The cache is not a filesystem but the permissions are the same...

PR #2161, opened 2018-02-09
Add command 'app:fix-title-case' to fix some text case

B-Galati:
https://app.clubhouse.io/enmarche/story/2751/command-to-fix-title-case

### DIFF
--- a/src/Command/FixTitleCaseCommand.php
+++ b/src/Command/FixTitleCaseCommand.php
@@ -1,0 +1,139 @@
+<?php
+
+namespace AppBundle\Command;
+
+use AppBundle\Doctrine\DBAL\BatchedConnection;
+use AppBundle\Form\TypeExtension\TextTypeExtension;
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class FixTitleCaseCommand extends Command
+{
+    private $connection;
+
+    /**
+     * @var SymfonyStyle
+     */
+    private $io;
+
+    public function __construct(Connection $connection)
+    {
+        parent::__construct();
+
+        $this->connection = new BatchedConnection($connection);
+    }
+
+    protected function configure()
+    {
+        $this
+            ->setName('app:fix-title-case')
+            ->setDescription('Fix case of adherents, citizen projects, committees and events for various property')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->io = new SymfonyStyle($input, $output);
+
+        $this->connection->setOnCommitCallback(function ($i) {
+            $this->io->success("$i queries processed in total");
+            usleep(500000); // 0.5s
+        });
+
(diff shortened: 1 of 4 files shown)


src/Command/FixTitleCaseCommand.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        $this
+            ->setName('app:fix-title-case')
+            ->setDescription('Fix case of adherents, citizen projects, committees and events for various property')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->io = new SymfonyStyle($input, $output);
```
Remg: a variable should be enough instead of a property, as it is only used locally

src/Command/FixTitleCaseCommand.php
```diff
@@ -0,0 +1,139 @@
...
+                [
+                    \PDO::PARAM_STR,
+                    \PDO::PARAM_STR,
+                    \PDO::PARAM_INT,
+                ]
+            );
+        }
+    }
+
+    protected function fixCommittesCase(): void
```
Remg: `fixCommitteesCase`

src/Command/FixTitleCaseCommand.php
```diff
@@ -0,0 +1,139 @@
...
+                    \PDO::PARAM_STR,
+                    \PDO::PARAM_INT,
+                ]
+            );
+        }
+    }
+
+    protected function fixCitizenProjectsCase(): void
+    {
+        foreach ($this->connection->executeQuery('SELECT id, `name`, subtitle FROM citizen_projects') as $project) {
```
Remg: Please be consistent with quotes around column names.
Either you use it, either you don't \o/

src/Command/FixTitleCaseCommand.php
```diff
@@ -0,0 +1,139 @@
...
+
+use AppBundle\Doctrine\DBAL\BatchedConnection;
+use AppBundle\Form\TypeExtension\TextTypeExtension;
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class FixTitleCaseCommand extends Command
```
Remg: Can we have simple functional tests ensuring the command runs without error? and maybe chewk a few outputs.

tests/Doctrine/DBAL/BatchedConnectionTest.php
```diff
@@ -0,0 +1,165 @@
...
+     * @var \PHPUnit_Framework_MockObject_MockObject|Connection
+     */
+    private $connection;
+
+    /**
+     * @var BatchedConnection
+     */
+    private $batchedConnection;
+
+    protected function setUp()
```
Remg: Please implement `tearDown()` method and clean properties to avoid memory leaks between tests

B-Galati:
@ottaviano @Remg 
I spoke with @ottaviano Quite a bit on this, as It's a one time command and the BatchedConnection is not that good. I'ill merge like this and remove everything once the command passed on prod.
Thanks guys :-) Let me know for any big issue (Specially with the SQL query, they are the important part)

PR #2182, opened 2018-02-14
Campus Re-wording / New btn link

Charlesdoiron:


templates/page/campus/home.html.twig
```diff
@@ -73,24 +73,16 @@
...
-                        À la fin de cette formation, vous aurez tous les outils
-                        et les conseils nécessaires pour que vos intuitions se
-                        transforment en vraies idées et en projets concrets
-                        sur votre territoire.
+                    <p>Ce MOOC est désormais terminé, mais vous pouvez accéder à l'ensemble du contenu et piocher tous les éléments qui peuvent vous être utiles !
                     </p>
                     <div class="text--center">
-                        <a href="{{ path('page_mooc') }}" class="btn btn--blue">
-                            Découvrir le MOOC
+                        <a href="https://auth.en-marche.fr/oauth/v2/auth?response_type=code&client_id=cadfecad-4b9a-4e29-b9ce-514222f8d028&redirect_uri=https://360l.en-marche.fr/auth&scope=&state=" class="btn btn--blue">
```
B-Galati: We cannot bring users to the MOOC website via OAuth.
The only thing we can do is updating the href with the new MOOC website.

templates/page/campus/home.html.twig
```diff
@@ -73,24 +73,16 @@
...
+                            Accéder au contenu du Mooc
```
marineb: "Mooc" should be "MOOC"

templates/page/campus/home.html.twig
marineb: 🤔 maybe we have a bug then? That's how we do it currently on this page. https://github.com/EnMarche/en-marche.fr/blob/57b674917d2dbe3748244213fe488b0430360eb7/templates/page/campus/mooc.html.twig#L28. What do you recommend we do @B-Galati?

B-Galati: Yea it is a bug.
We should just add a link to the Mooc website and that's all. It's up to them to redirect back if the user is not authenticated. We cannot call directly OAuth with their `client_id`. It would be a security issue. That's why we have the `state` parameters.

marineb: @atamano can you confirm ? I know 360 had given us that URL to fix another bug via you. I can't recall exactly what it was about.

templates/page/campus/home.html.twig
```diff
@@ -73,24 +73,16 @@
                 <div class="campus__text campus__text--center">
                     <h2>Passer à l'action</h2>
                     <p>
-                        Inscrivez-vous vite pour participer à notre première formation
-                        collective qui a pour thème "Agir près de chez moi".
+                        Nous proposons enfin des formations collectives en ligne sur plusieurs semaines, communément appelées MOOC ("Massive Open Online Courses").
```
Remg: Please split it in several lines to increase readability.
A line _should not_ exceed 80 characters.

Like @thomasbisignani  once said:
> scrolling horizontally is something i wouldn't wish on my worst enemy 👌

templates/page/campus/home.html.twig
atamano: @marineb you can add https://mooc.en-marche.fr => it is configured to redirect automatically

marineb: Thank you @atamano. @Charlesdoiron ^^ All good now. Let's wrap this up 🚀

templates/page/campus/home.html.twig
Charlesdoiron: ok : )

templates/page/campus/home.html.twig
Charlesdoiron: Done

templates/page/campus/home.html.twig
Charlesdoiron: Done

PR #2185, opened 2018-02-14
CP Admin : Added category filter and updated export fields

thomasbisignani:
For the Citizen Project Admin list view :
- Added category filter
- Updated export fields
- Fix a bug in the committee supports template (added getter for status property)

### DIFF
--- a/src/Admin/CitizenProjectAdmin.php
+++ b/src/Admin/CitizenProjectAdmin.php
@@ -226,6 +226,9 @@ class CitizenProjectAdmin extends AbstractAdmin
                 'label' => 'Nom',
                 'show_filter' => true,
             ])
+            ->add('category', null, [
+                'label' => 'Catégorie',
+            ])
             ->add('createdAt', DateRangeFilter::class, [
                 'label' => 'Date de création',
                 'field_type' => 'sonata_type_date_range_picker',
@@ -424,5 +427,26 @@ class CitizenProjectAdmin extends AbstractAdmin
                 'template' => 'admin/citizen_project/list_actions.html.twig',
             ])
         ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getExportFields()
+    {
+        return [
+            'Nom' => 'name',
+            'Date de création' => 'createdAt',
+            'Catégorie' => 'category',
+            'Créateur' => 'createdBy',
+            'Membres' => 'membersCounts',
+            'Ville' => 'postAddress.cityName',
+            'Compétences recherchées' => 'exportSkills',
+            'Compétences matchées' => 'matchedSkills',
+            'Demande d\'accomp.' => 'assistanceNeeded',
+            'Statut' => 'status',
+            'Comités en soutien' => 'exportCommitteeSupports',
+            'Coup de coeur' => 'featured',
+        ];
     }
 }
--- a/src/Entity/CitizenProject.php
+++ b/src/Entity/CitizenProject.php
@@ -672,4 +672,16 @@ class CitizenProject extends BaseGroup implements CoordinatorAreaInterface
     {
         $this->nextAction = $citizenAction;
     }
+
+    public function exportCommitteeSupports(): string
+    {
+        return \implode(', ', array_map(function (CitizenProjectCommitteeSupport $committeeSupport) {
+            return \sprintf('%s [%s]', $committeeSupport->getCommittee()->getName(), $committeeSupport->getStatus());
+        }, $this->committeeSupports->toArray()));
+    }
+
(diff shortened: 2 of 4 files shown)


src/Admin/CitizenProjectAdmin.php
```diff
@@ -425,4 +428,25 @@ protected function configureListFields(ListMapper $listMapper)
...
+    public function getExportFields()
+    {
+        return [
+            'Nom' => 'name',
+            'Date de création' => 'createdAt',
+            'Catégorie' => 'category',
+            'Créateur' => 'createdBy',
+            'Membres' => 'membersCounts',
+            'Ville' => 'postAddress.cityName',
+            'Compétences recherchées' => 'skills',
```
Remg: should you use `Comp.` to harmonize labels ?

src/Entity/CitizenProject.php
```diff
@@ -672,4 +672,18 @@ public function setNextAction(CitizenAction $citizenAction): void
...
+    public function exportCommitteeSupports(): string
+    {
+        return join(', ', array_map(function (CitizenProjectCommitteeSupport $committeeSupport) {
+            return $committeeSupport->getCommittee()->getName();
+        }, $this->committeeSupports->toArray()));
+    }
+
+    public function exportSkills(): string
+    {
+        return join(', ', array_map(function (CitizenProjectSkill $skill) {
```
Remg: `return join(', ', $this->skills->toArray());

thanks to `BaseSkill::__toString()`

thomasbisignani: Nice catch ! :+1: Thx

src/Entity/CitizenProject.php
```diff
@@ -672,4 +672,16 @@ public function setNextAction(CitizenAction $citizenAction): void
     {
         $this->nextAction = $citizenAction;
     }
+
+    public function exportCommitteeSupports(): string
+    {
+        return join(', ', array_map(function (CitizenProjectCommitteeSupport $committeeSupport) {
+            return $committeeSupport->getCommittee()->getName();
```
Remg: It would be nice to know the status of the committee on the exported sheet to avoid confusion.
Either we should only export approved supports, either we should display the status of each support.

templates/admin/citizen_project/list_committee_supports.html.twig
```diff
@@ -6,7 +6,7 @@
             {% for committeeSupport in object.committeeSupports %}
                 <li>
                     {{ committeeSupport.committee.name }}
-                    [{{ committeeSupport.status }}]
+                    [{{ committeeSupport.committee.status }}]
```
Remg: 🤔 we want to display `CitizenProjectComitteeSupport::$status` here, not the Comittee's status.

thomasbisignani: Ok so I add the `status` getter in `CitizenProjectComitteeSupport`, because there was a bug here.

src/Admin/CitizenProjectAdmin.php
thomasbisignani: After discussing with @guillaumedenhez , we will keep "Compétences" for both.

src/Admin/CitizenProjectAdmin.php
```diff
@@ -399,7 +402,7 @@ protected function configureListFields(ListMapper $listMapper)
                 'label' => 'Compétences recherchées',
             ])
             ->add('matchedSkills', null, [
-                'label' => 'Comp. matchées',
+                'label' => 'Compétences. matchées',
```
HeahDude: Remove the dot now :)

src/Entity/CitizenProject.php
```diff
@@ -672,4 +672,16 @@ public function setNextAction(CitizenAction $citizenAction): void
     {
         $this->nextAction = $citizenAction;
     }
+
+    public function exportCommitteeSupports(): string
+    {
+        return join(', ', array_map(function (CitizenProjectCommitteeSupport $committeeSupport) {
```
HeahDude: `join` => `\implode`?

src/Entity/CitizenProjectCommitteeSupport.php
```diff
@@ -85,6 +85,11 @@ public function getCommittee(): Committee
         return $this->committee;
     }
 
+    public function getStatus(): string
```
HeahDude: You don't have to add the getter if it's not used.

thomasbisignani: This getter is used by Sonata in the Citizen Project list view.

PR #2191, opened 2018-02-15
fix tests committee edit delete

cacahouwete:


src/Form/CommitteeFeedMessageType.php
```diff
@@ -40,6 +42,9 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefault('data_class', CommitteeMessage::class);
+        $resolver->setDefaults([
+            'data_class' => CommitteeMessage::class,
+            'need_subject' => true,
+        ]);
```
Remg: Please typehint custom options. e.g.: 
```
$resolver
    ->setDefaults([
        // ...
        'need_subject' => true,
    ])
    ->setAllowedTypes('need_subject', 'bool')
;
```

src/Form/CommitteeFeedMessageType.php
```diff
@@ -15,14 +15,16 @@ class CommitteeFeedMessageType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder
-            ->add('subject', TextType::class, [
+        if ($options['data_class'] === CommitteeMessage::class) {
```
Remg: Yoda notation you should use.

src/Form/CommitteeFeedMessageType.php
```diff
@@ -15,14 +15,16 @@ class CommitteeFeedMessageType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder
-            ->add('subject', TextType::class, [
+        if (CommitteeMessage::class === $options['data_class']) {
```
Remg: There should be a separate form that inherits from this one and just add the new field (or remove the extra field if it's the exception).

src/Form/CommitteeFeedMessageType.php
```diff
@@ -3,43 +3,31 @@
...
-        ;
+        $builder->add('subject', TextType::class, [
+            'label' => false,
+            'attr' => [
+                'class' => 'form--full b__nudge--bottom-medium',
+                'placeholder' => 'Entrez l\'objet de votre message',
+            ],
+        ]);
+
+        parent::buildForm($builder, $options);
```
B-Galati: Please use Form inheritance like shown here -> https://github.com/EnMarche/en-marche.fr/blob/develop/src/Form/BecomeAdherentType.php

This is not the normal way of doing inheritance in forms. Event if it's working like this right now, we don't know what sort of side effects it could do. Let's be safe.

The Form Component has its own way of managing inheritance for some reasons. 

You can do Php Inheritance for classes that are like `AbstractType` meaning that they are not a FormType. They are just a way of factorizing a bunch of code but I don't think it's a good solution either.

Really sorry about that.

See :
http://symfony.com/doc/master/form/create_form_type_extension.html#generic-form-type-extensions
https://stackoverflow.com/questions/22861044/symfony-formtype-getparent-vs-inheritance
https://github.com/symfony/symfony/issues/3879

HeahDude: There are cases when we can use PHP inheritance, but this one is not. Because we don't use those types as complex service or such, we want to keep configuration and template inheritance mechanism handled by the component.
Please use `getParent()` and remove calls to `parent::xx()`.

src/Form/CommitteeFeedMessageType.php
```diff
@@ -16,7 +16,7 @@ class CommitteeFeedMessageType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('subject', TextType::class, [
+                ->add('subject', TextType::class, [
```
thomasbisignani: Why this change ? it's not CS.

illusionOfParadise: Indent

PR #2192, opened 2018-02-15
Add Google IP ranges to trusted proxies

B-Galati:
https://app.clubhouse.io/enmarche/story/2784/l-ip-récupéré-sur-les-failed-login-attempt-les-donations-etc-n-est-pas-bon


web/index.php
```diff
@@ -38,6 +37,8 @@
     }
 }
 
+require __DIR__.'/../app/trusted_proxies.php';
```
cacahouwete: Why moving require here ?

B-Galati: Like this the require is behind the `Debug::enable()` statement in dev and you can see any errors from it.

PR #2194, opened 2018-02-15
Quoting string with % character in yml file

ottaviano:


Remg:
Could you please harmonize the whole file with double quotes for parameters ? 
e.g.: https://github.com/EnMarche/en-marche.fr/blob/59a6a87029a46ab7e5c0de2b2960092f3263c836/app/config/config.yml#L25-L31

circle.yml
```diff
@@ -56,6 +56,9 @@ test:
         - yarn lint
         - yarn test
 
+        # Ling YML
```
Remg: Lin**t***

PR #2202, opened 2018-02-19
WebHook - Move WebHook admin into OAuth Client

B-Galati:
Also simplify the logic of web hook creation : no more creation on the fly
https://app.clubhouse.io/enmarche/story/1770/auth-etq-admin-je-veux-pouvoir-définir-des-webhooks-pour-une-application-dans-le-backoffice


src/Entity/WebHook/WebHook.php
```diff
@@ -27,76 +27,83 @@ class WebHook
...
-    public function __construct(Event $event, $callbacks = [])
+    public function __construct(Client $client = null, Event $event = null, $callbacks = [])
     {
         $this->uuid = Uuid::uuid4();
-        $this->event = $event->getValue();
-        $this->callbacks = new ArrayCollection();
+        $this->client = $client;
+        $this->event = $event ? $event->getValue() : null;
 
+        $this->callbacks = [];
```
ottaviano: do this on declaration of the property `private $callbacks = [];`

src/Entity/WebHook/WebHook.php
```diff
@@ -27,76 +27,83 @@ class WebHook
...
+     * @var string[]
+     *
+     * @ORM\Column(type="json")
+     *
+     * @Assert\Count(min=1, minMessage = "Veuillez spécifier au moins une url de callback.")
      */
     private $callbacks;
 
-    public function __construct(Event $event, $callbacks = [])
+    public function __construct(Client $client = null, Event $event = null, $callbacks = [])
```
ottaviano: Why all arguments are optional?

src/Form/WebHookType.php
```diff
@@ -3,24 +3,33 @@
...
         $builder
-            ->add('client', EntityType::class, ['class' => Client::class])
-            ->add('urls', CollectionType::class, [
+            ->add('client', EntityType::class, [
+                'class' => Client::class,
+                'disabled' => true,
+            ])
+            ->add('event', ChoiceType::class, [
+                'placeholder' => '--',
+                'choices' => array_combine(Event::toArray(), Event::toArray()),
```
ottaviano: hm i don't see the real gain of `array_combine` and using 2x memory instead of `choice_label` option.

src/Entity/WebHook/WebHook.php
B-Galati: Because we are coupled to sonata and forms. So when creating a new web hook sonata needs a clean web hook instance. I know I can override the object creation in the SonataAdmin class but it would just enforce the `$client` property.

src/Entity/WebHook/WebHook.php
B-Galati: What's the advantages of doing so ?

src/Form/WebHookType.php
B-Galati: It's simpler to write IMO. In this case, it's not ideal. I'll change it.

src/Entity/WebHook/WebHook.php
ottaviano: I think it's more clearer and readable to use the construct for instantiating object / inject values and let the definition of scalar value on property declaration. 
```php
class Foo
{
    private $status;
    private $step;
    private $children;
    private $createdAt;

    function __construct()
    {
        $this->status = 'DEFAULT STATUS';
        $this->step = 1; // default step
        $this->children = [];
        $this->createdAt = new \DateTime;
    }
}
```

```php
class Foo
{
    private $status = 'DEFAULT STATUS';
    private $step = 1; // default step;
    private $children = [];
    private $createdAt;

    function __construct()
    {
        $this->createdAt = new \DateTime;
    }
}
```

B-Galati: IMO, it's simpler to put everything in the constructor so that you know it's all in the constructor and you don't need to look for anything anywhere else.

PR #2204, opened 2018-02-19
[ch2787] Finish implementation of switching follower to organisator.

cacahouwete:
- Implementation of changePrivilege method
- Remove organisator transformation button

https://app.clubhouse.io/enmarche/story/2787/pc-etq-admin-je-dois-pouvoir-changer-un-organisateur-en-follower-et-et-un-follower-en-organisateur

### DIFF
--- a/src/CitizenProject/CitizenProjectAuthority.php
+++ b/src/CitizenProject/CitizenProjectAuthority.php
@@ -41,6 +41,15 @@ class CitizenProjectAuthority
         if (!$membership = $this->getCitizenProjectMembership($adherent, $citizenProject)) {
             return;
         }
+        if (CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR === $privilege) {
+            $membershipAdmins = $this->membershipRepository->findPriviledgedMemberships(
+                $citizenProject->getUuid(),
+                [CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR]
+            );
+            foreach ($membershipAdmins as $membershipAdmin) {
+                $membershipAdmin->setPrivilege(CitizenProjectMembership::CITIZEN_PROJECT_FOLLOWER);
+            }
+        }
 
         $membership->setPrivilege($privilege);
     }
--- a/src/CitizenProject/CitizenProjectManager.php
+++ b/src/CitizenProject/CitizenProjectManager.php
@@ -186,6 +186,7 @@ class CitizenProjectManager
     {
         $citizenProject->approved();
 
+        /** @var Adherent $creator */
         $creator = $this->getAdherentRepository()->findOneByUuid($citizenProject->getCreatedBy());
         $this->projectAuthority->changePrivilege($creator, $citizenProject, CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR);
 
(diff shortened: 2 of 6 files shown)


src/CitizenProject/CitizenProjectManager.php
```diff
@@ -344,6 +344,26 @@ public function removeImage(CitizenProject $citizenProject): void
         $citizenProject->setImageUploaded(false);
     }
 
+    public function changePrivilege(CitizenProject $citizenProject, Adherent $adherent, string $privilege, bool $flush = true): void
```
ottaviano: you change the privilege of an adherent, for me adherent must be in the first position (+ for being as others `changePrivilege` methods of this project ;)

src/CitizenProject/CitizenProjectManager.php
```diff
@@ -344,6 +344,26 @@ public function removeImage(CitizenProject $citizenProject): void
         $citizenProject->setImageUploaded(false);
     }
 
+    public function changePrivilege(Adherent $adherent, CitizenProject $citizenProject, string $privilege, bool $flush = true): void
```
illusionOfParadise: It's confusing to have 2 almost the same functions. The function with the same name, almost the same list of arguments already exists https://github.com/EnMarche/en-marche.fr/blob/d5a32b4a5fddb28b585aeb058ef9dac67681c522/src/CitizenProject/CitizenProjectAuthority.php#L37
We should have only one `changePrivilege ` function.

templates/admin/citizen_project/members.html.twig
```diff
@@ -65,10 +65,6 @@
                                     {% endif %}
                                 </td>
                                 <td>
-                                    <a href="{{ path('app_admin_citizenproject_change_privilege', {citizenProject: citizen_project.id, adherent: administrator.id, privilege: constant('CITIZEN_PROJECT_FOLLOWER', membership), token: token}) }}"
-                                       class="btn btn-default btn-sm">
-                                        Transformer en follower
-                                    </a>
                                 </td>
```
Remg: Should we leave this `<td>` empty ?

ottaviano: need transform the entire table (all rows) on table of 2 cols and not 3 cols

src/CitizenProject/CitizenProjectAuthority.php
```diff
@@ -41,6 +41,16 @@ public function changePrivilege(Adherent $adherent, CitizenProject $citizenProje
         if (!$membership = $this->getCitizenProjectMembership($adherent, $citizenProject)) {
             return;
         }
+        if (CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR === $privilege) {
+            $membershipAdmins = $this->membershipRepository->findPriviledgedMemberships(
+                $citizenProject->getUuid(),
+                [CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR]
+            );
+            if (1 === \count($membershipAdmins)) {
```
Remg: Okay i think the functional behavior here is weird.

If more than 1 administrator is found: there won't be any privilege changed and nobody will be notified.
If count is different than 1, an Exception should be thrown to tell the user what's happening?
Or maybe a flash message should be displayed ?

cacahouwete: I will change it to put all admin as followers

PR #2213, opened 2018-02-20
Added user unregistration

thomasbisignani:


src/Twig/AdherentExtension.php
```diff
@@ -35,13 +41,22 @@ public function getMemberInterestLabel(string $interest)
...
     }
 
-    public function getUnregistrationReasonsLabel(string $reasons)
+    public function getUnregistrationAdherentReasonsLabel(string $reasons): string
+    {
+        if (!isset($this->unregistrationAdherentReasons[$reasons])) {
+            return '';
+        }
+
+        return $this->unregistrationAdherentReasons[$reasons];
```
Remg: `return $this->unregistrationAdherentReasons[$reasons] ?? '';`

src/Entity/Unregistration.php
```diff
@@ -60,19 +60,26 @@ class Unregistration
      */
     private $unregisteredAt;
 
+    /**
+     * @ORM\Column(type="boolean", options={"default": false})
+     */
+    private $adherent = false;
```
ottaviano: what about `isAdherent` (`is_adherent` in DB) for the column name too?
IMO it's clearer

src/Form/UnregistrationAdherentReasonsChoiceType.php
```diff
@@ -9,9 +9,9 @@
 use Symfony\Component\Form\FormEvents;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class UnregistrationReasonsChoiceType extends AbstractType
+class UnregistrationAdherentReasonsChoiceType extends AbstractType
```
ottaviano: this form can be used by two formTypes `UnregistrationAdherentType` & `UnregistrationUserType`, for me not need to rename it, no?

src/Form/UnregistrationUserReasonsChoiceType.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace AppBundle\Form;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class UnregistrationUserReasonsChoiceType extends AbstractType
```
ottaviano: IMO it's useless

src/Form/UnregistrationAdherentType.php
```diff
@@ -9,7 +9,7 @@
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class UnregistrationType extends AbstractType
+class UnregistrationAdherentType extends AbstractType
```
ottaviano: idem, I think it can be the same class for two type of unregistration because it has juste the choices option was changed

templates/admin/adherent/list_status.html.twig
```diff
@@ -30,6 +30,8 @@
             <span class="label label-primary" style="padding: 2px;">Membre du Conseil</span><br/>
         {% endif %}
     {% elseif object.isAdherent %}
-        <span class="label label-default">Adhérent</span>
+        <span class="label label-primary">Adhérent</span>
+    {% elseif object.isUser %}
```
ottaviano: `isUser === !isAdherent` so juste `{% else %}`

templates/user/terminate_membership.html.twig
```diff
@@ -0,0 +1,125 @@
...
+                <div class="l__wrapper--narrow">
+                    <h1 class="text--center text--larger">
+                        Je souhaite supprimer mon compte <em>En Marche</em><br/>
+                        Nous sommes tristes de vous voir partir 😢
+                    </h1>
+                </div>
+            </header>
+            <section class="l__wrapper--600 settings__unsub__intro">
+                <p>
+                    Merci de nous aider à comprendre pourquoi vous souhaitez quitter En Marche en choisissant
```
ottaviano: En marche -> `La République En Marche`?

templates/user/terminate_membership.html.twig
```diff
@@ -0,0 +1,125 @@
...
+            <header>
+                <h1 class="space--60-0 l__wrapper--slim text--center text--large">
+                    Votre compte En Marche a bien été supprimé et vos données personnelles effacées de notre base.
+                </h1>
+                <hr class="l__wrapper"/>
+            </header>
+            <section class="l__wrapper--600 text--body">
+                <p>
+                    Nous sommes tristes de vous voir partir. Vous serez
+                    toujours les bienvenus si vous souhaitez nous rejoindre.
```
ottaviano: `les` -> `le`?

app/config/adherent.yml
```diff
@@ -26,3 +26,10 @@ parameters:
         mouvement: Je ne me reconnais plus dans les actions du mouvement En Marche !
         comite: Je suis en désaccord avec l’activité de mon comité local
         autre: Autre
+    user_unregistration_reasons:
```
B-Galati: Let's use the translator instead of this ? 
We have something similar for the report feature with some reasons to check and it uses Translator. You can check it out. It's far easier to manage then.

src/Entity/Adherent.php
```diff
@@ -1051,6 +1051,11 @@ public function isAdherent(): bool
         return $this->adherent;
     }
 
+    public function isUser(): bool
+    {
+        return !$this->adherent;
```
B-Galati: It would be better to call the `isAdherent` method not to break encapsulation of this behavior.

thomasbisignani: I think it's more readable to have an explicit `isUser()` method to check if the person is an user.
Because everywhere you call the method `!$object->isAdherent()`, you will have to explain why it is a user.. it's not really logic.

B-Galati: > I think it's more readable to have an explicit isUser() method to check if the person is an user.
I agree

My point was to keep the method but use the method `isAdherent()` instead of the property directly.

thomasbisignani: Ah ok, nice idea :+1:

templates/user/terminate_membership.html.twig
thomasbisignani: I asked @guillaumedenhez for this question, we can keep "En Marche" for the moment :+1:

src/Form/UnregistrationType.php
```diff
@@ -18,23 +17,19 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-            'validation_groups' => function (FormInterface $form) {
-                $data = $form->getData();
-
-                if (in_array('autre', $data->getReasons())) {
-                    return ['Default', 'Reason'];
-                }
-
-                return ['Default'];
-            },
+            'reasons' => [],
```
ottaviano: need to keep the `data_class` option, no?

thomasbisignani: of course! thx :+1:

src/Admin/UnregistrationAdmin.php
```diff
@@ -86,7 +87,10 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                 'show_filter' => true,
                 'field_type' => ChoiceType::class,
                 'field_options' => [
-                    'choices' => array_flip($this->unregistrationReasons),
+                    'choices' => array_flip(
```
B-Galati: I am not sure but I don't think this would work as we would have integer in choices after `array_flip`. Let me know if I am wrong. I just like to be sure.

thomasbisignani: Nice catch! thx :+1:

B-Galati:
:100: :+1:

PR #2214, opened 2018-02-20
[ch2618] Modify meta-description for events

cacahouwete:
https://app.clubhouse.io/enmarche/story/2618/modification-des-meta-description-des-%C3%A9v%C3%A9nements

### DIFF
--- a/templates/events/show.html.twig
+++ b/templates/events/show.html.twig
@@ -3,6 +3,10 @@
 {% block page_title event.name %}
 {% block canonical_url url('app_event_show', {'slug': event.slug}) %}
 
+{% block meta_description %}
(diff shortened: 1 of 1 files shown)


templates/events/show.html.twig
```diff
@@ -3,6 +3,10 @@
 {% block page_title event.name %}
 {% block canonical_url url('app_event_show', {'slug': event.slug}) %}
 
+{% block meta_description %}
+    {{ event.name }} est un événement local du mouvement La République En Marche ! organisé par {{ event.organizer ? event.organizer.partialName : '' }} le {{ event.beginAt|localizeddate('full', 'none') }} à {{ event.beginAt|date('G\\hi') -}}, dans la ville de {{ event.postAddress.cityName }}. Visitez la page dédiée pour vous inscrire et obtenir toutes les informations concernant cet événement.
```
illusionOfParadise: If event organizer does not exist it means that adherent was unregistered. It's better to display `un ancien adhérent` instead of empty string.

PR #2218, opened 2018-02-21
Desindex un-needed entities from Algolia 

B-Galati:


src/Entity/Clarification.php
```diff
@@ -12,6 +13,8 @@
  *
  * @UniqueEntity(fields={"slug"})
  * @Gedmo\SoftDeleteable(fieldName="deletedAt", timeAware=false)
+ *
+ * @Algolia\Index(autoIndex=false)
```
Remg: `Clarification` entity should be indexed on Algolia.
see https://github.com/EnMarche/en-marche.fr/blob/develop/src/Command/AlgoliaSynchronizeCommand.php#L24

thomasbisignani:
:medal_military: :medal_sports:

PR #2219, opened 2018-02-21
[ADMIN][ch2672] Fix disappearing of coordinator managed area while Adherent edition

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/2672/etq-admin-si-je-mets-une-zone-r%C3%A9f%C3%A9rent-a-un-adh%C3%A9rent-qui-est-d%C3%A9j%C3%A0-coordinateur-son-r%C3%B4le-coordinateur-saute

### DIFF
--- a/src/Form/EventListener/CoordinatorManagedAreaListener.php
+++ b/src/Form/EventListener/CoordinatorManagedAreaListener.php
@@ -66,10 +66,6 @@ class CoordinatorManagedAreaListener implements EventSubscriberInterface
             return;
         }
 
-        $adherent->setCoordinatorManagedAreas(
-            $adherent->getCoordinatorManagedAreas()->filter(function (CoordinatorManagedArea $entity) {
-                return !empty($entity->getCodes()) && !empty($entity->getSector());
-            })
-        );
+        $adherent->removeEmptyCoordinatorManagedAreas();
     }
 }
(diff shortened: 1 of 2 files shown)


src/Form/EventListener/CoordinatorManagedAreaListener.php
```diff
@@ -66,10 +66,10 @@ public function onPostSubmit(FormEvent $event): void
             return;
         }
 
-        $adherent->setCoordinatorManagedAreas(
-            $adherent->getCoordinatorManagedAreas()->filter(function (CoordinatorManagedArea $entity) {
-                return !empty($entity->getCodes()) && !empty($entity->getSector());
-            })
-        );
+        foreach ($adherent->getCoordinatorManagedAreas() as $area) {
```
B-Galati: Let's move this to `Adherent` class instead. 
This listener would be lighter, it's easier to unit test and more reusable.
What do you think ?

B-Galati: With the right method name, it would also bring more semantic

illusionOfParadise: @B-Galati why not.
The name like `removeEmptyCoordinatorManagedAreas` ?

PR #2221, opened 2018-02-21
[ch2891] Design modification for button action in committee timeline

cacahouwete:
- scss modification to have correct display
- fix display next TimeLines and add some test

https://app.clubhouse.io/enmarche/story/2891/comit%C3%A9s-qqs-modifications-sur-la-suppression-ou-l-%C3%A9dition-des-messages

### DIFF
--- a/front/style/layout/_base.scss
+++ b/front/style/layout/_base.scss
@@ -26,26 +26,6 @@ svg polygon {
     border: none;
 }
 
-@mixin btn--as-link($color) {
-    background: none!important;
-    border: none;
-    padding: 0!important;
-    color: $color;
-    border-bottom: 1px solid transparent;
-
-    &:hover {
-        border: none;
-        border-bottom: 1px solid $color;
-        color: $color;
-    }
-}
-
-button {
-    &.link--deletered {
-        @include btn--as-link($red);
-    }
-}
-
 a {
     text-decoration: none;
     border-bottom: 1px solid black;
--- a/front/style/layout/_base.scss
+++ b/front/style/layout/_base.scss
@@ -26,6 +26,22 @@ svg polygon {
     border: none;
 }
 
+button {
+    &.link--deletered {
+        background:none!important;
+        border:none;
+        padding:0!important;
+        color: $red;
+        border-bottom: 1px solid transparent;
+
+        &:hover {
+            border: none;
+            border-bottom: 1px solid $red;
+            color: $red;
+        }
+    }
+}
+
 a {
     text-decoration: none;
     border-bottom: 1px solid black;
--- a/front/style/layout/_base.scss
+++ b/front/style/layout/_base.scss
@@ -26,19 +26,23 @@ svg polygon {
     border: none;
 }
 
+@mixin btn--as-link($color) {
+    background: none!important;
+    border: none;
+    padding: 0!important;
+    color: $color;
+    border-bottom: 1px solid transparent;
+
+    &:hover {
+        border: none;
+        border-bottom: 1px solid $color;
+        color: $color;
+    }
+}
+
 button {
     &.link--deletered {
-        background:none!important;
-        border:none;
-        padding:0!important;
-        color: $red;
-        border-bottom: 1px solid transparent;
-
-        &:hover {
-            border: none;
-            border-bottom: 1px solid $red;
-            color: $red;
-        }
+        @include btn--as-link($red);
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/Controller/EnMarche/CommitteeController.php
```diff
@@ -144,7 +144,9 @@ public function timelineAction(Request $request, Committee $committee): Response
         return $this->render('committee/timeline/feed.html.twig', [
             'committee' => $committee,
             'committee_timeline' => $timeline,
+            'committee_timeline_forms' => $this->createTimelineDeleteForms($timeline),
             'has_role_adherent' => $this->getUser() instanceof Adherent && $this->getUser()->isAdherent(),
+            'has_role_user' => $this->isGranted('ROLE_USER'),
```
Remg: This should not live in the controller; Twig environment has registered functions for this. (like `is_granted(xxx)`)

I see `has_role_adherent` was already badly used here.

Please remove `has_role_adherent` and `has_role_user` and retrieve these informations directly in the template.
See: https://github.com/EnMarche/en-marche.fr/blob/develop/templates/base.html.twig#L9

templates/committee/timeline/feed.html.twig
```diff
@@ -4,14 +4,16 @@
...
                      <li>
-                         <a href="{{ path('app_committee_timeline_edit', { slug: committee.slug, id: item.id }) }}">
+                         <a class="text--body link--newblue" href="{{ path('app_committee_timeline_edit', { slug: committee.slug, id: item.id }) }}">
                              Modifier le message
                          </a>
                      </li>
                      <li>
-                         {{ form(committee_timeline_forms[item.id]) }}
+                         {{ form_start(committee_timeline_forms[item.id]) }}
+                         {{ form_widget(committee_timeline_forms[item.id].delete, {'attr': {'class': 'link--deletered'}}) }}
```
Remg: CS for twig associative arrays we follow on this project:
- no quotes around keys
- spaces inside brackets

`{'attr': {'class': 'link--deletered'}}` => `{ attr: { class: 'link--deletered' } }`

src/Controller/EnMarche/CommitteeController.php
cacahouwete: Not possible

cacahouwete: This template is used with ajax call and include call :s

Remg: indeed, my bad.
this template is used as an inclusion in the layout and also as a single template.

ottaviano: do we really need to call `createTimelineDeleteForms` and generate N forms for delete each feed item if the current user is not a adherent (hasn't ROLE_USER)?

what about check the role in the controller and build the forms if user has the good role, and in the template check only if `committee_timeline_forms` is defined?

front/style/layout/_base.scss
```diff
@@ -26,6 +26,22 @@ svg polygon {
     border: none;
 }
 
+button {
+    &.link--deletered {
```
ottaviano: maybe it'll be nice to have a more generic class name to transform one button to a link and in `_button.scss` file?
ex:
```css
@mixin btn--as-link {
  /* your css here */
}

button {
  & .btn--as-link {
    @include btn--as-link; 
  }
}

input {
  &[type="submit"],
  &[type="button"],
  &[type="reset"] {
    & .btn--as-link {
      @include btn--as-link;
    }
  }
}
```

front/style/layout/_base.scss
```diff
@@ -26,6 +26,22 @@ svg polygon {
...
+        background:none!important;
+        border:none;
+        padding:0!important;
```
ottaviano: CS `option: value;`

B-Galati:
I like @ottaviano's idea.

> do we really need to call createTimelineDeleteForms and generate N forms for delete each feed item if the current user is not a adherent (hasn't ROLE_USER)?
> what about check the role in the controller and build the forms if user has the good role, and in the template check only if committee_timeline_forms is defined?

src/Controller/EnMarche/CommitteeController.php
cacahouwete: To create delete form, I can have a twig extension like create_delete_form(form) and in twig template have
```twig
{% set delete_form = create_delete_form(path('app_committee_timeline_delete', { id: item.id, slug: item.committee.slug }), 'committee_feed_delete') %}
{{ form_start(committee_timeline_forms[item.id]) }}
{{ form_widget(committee_timeline_forms[item.id].delete, { attr: { class: 'link--deletered'}}) }}
{{ form_end(committee_timeline_forms[item.id]) }}
```

PR #2223, opened 2018-02-22
[CA] Change link displaying + add link to mail + fix error message

illusionOfParadise:
The field `contacts` was deleted from Type class and added directly in template to avoid displaying error message when array of `contacts` passed to controller from previous page and not from page with concerned form.

### DIFF
--- a/templates/citizen_action_manager/contact_participants.html.twig
+++ b/templates/citizen_action_manager/contact_participants.html.twig
@@ -39,7 +39,7 @@
 
                 {{ form_start(form) }}
 
-                   {{ form_widget(form.contacts, { value: contacts|json_encode }) }}
+                    <input type="hidden" name="contacts" value="{{ contacts|json_encode }}" />
 
                     <div class="form__row subject">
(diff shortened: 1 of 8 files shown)


templates/citizen_action_manager/contact_participants.html.twig
```diff
@@ -39,7 +39,7 @@
 
                 {{ form_start(form) }}
 
-                   {{ form_widget(form.contacts, { value: contacts|json_encode }) }}
+                    <input type="hidden" name="contacts" value="{{ contacts|json_encode }}" />
```
B-Galati: Why removing the field from the form type ? Is it just a way of giving some data to some javascript code ?
In that we should completely remove this field no ? 
If we keep the form. Is the name need to be the same as the form ? (in that case we should `vars` from the `form`)

illusionOfParadise: I explained in the PR comment why I do this. 
We need `contacts` in the controller action. The datas posted to this action are different depending on page accessing this action. Actually I can't find more propre/simple solution than this one.

PR #2228, opened 2018-02-26
[Emails] Refactor & test messages

Remg:
- Message classes do not hold Template & Subject anymore.
    - Template is defined through the MessageRegistry
    - Subject is defined in the message template
- Unit test for all Message class
- Message & Test harmonization


app/config/services/admin.xml
```diff
@@ -598,5 +598,15 @@
...
         </service>
 
+        <!--
+        ######################################################
+        # Menu
+        ######################################################
+         -->
+
+        <service id="app.admin.menu.email" class="AppBundle\Admin\EventListener\AdminEmailMenuListener">
+            <argument type="service" id="doctrine.orm.entity_manager"/>
```
thomasbisignani: Why inject the entity manager ? it's useless IMO.

PR #2230, opened 2018-02-26
[PROCURATION][ch2470] Show only datas concerning upcoming elections

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/2470/procuration-etq-respo-procuration-d-une-%C3%A9lection-sur-le-filtre-tour-concern%C3%A9-je-ne-veux-voir-appara%C3%AEtre-que-les-tours-de

### DIFF
--- a/src/Procuration/Filter/ProcurationFilters.php
+++ b/src/Procuration/Filter/ProcurationFilters.php
@@ -156,8 +156,11 @@ abstract class ProcurationFilters
             }
         }
 
-        $qb->leftJoin("$alias.electionRounds", 'rounds');
-        $qb->addSelect('rounds');
+        $qb
+            ->leftJoin("$alias.electionRounds", 'rounds')
+            ->addSelect('rounds')
+            ->andWhere('rounds.date >= CURRENT_DATE()')
+        ;
 
         if ($this->electionRound) {
             $qb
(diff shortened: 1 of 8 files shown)


src/Procuration/Filter/ProcurationFilters.php
```diff
@@ -157,7 +157,11 @@ public function apply(QueryBuilder $qb, string $alias): void
         }
 
         $qb->leftJoin("$alias.electionRounds", 'rounds');
-        $qb->addSelect('rounds');
+        $qb
```
thomasbisignani: You can write directly : 
```
        $qb
            ->leftJoin("$alias.electionRounds", 'rounds')
            ->addSelect('rounds')
            ->andWhere('rounds.date >= :now')
            ->setParameter('now', (new \DateTime())->format('Y-m-d'))
        ;
```

src/Repository/ElectionRoundRepository.php
```diff
@@ -40,4 +40,21 @@ public function getAllRoundsAsChoices(): array
 
         return $rounds ?? [];
     }
+
+    public function getUpcomingElectionRounds(): array
+    {
+        $results = $this->createQueryBuilder('r')
```
ottaviano: `$this->createQueryBuilder('r', 'r.id')` and don't need the `foreach`, the result array will indexed by id
but each item of array will contents two columns: id and label

PR #2235, opened 2018-02-27
[REFERENT][ch2964] Add managed committee postal code and send referent mails to AL of committees in the referent managed zone

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/2964/etq-qu-animateur-local-si-mon-code-postal-ne-match-pas-celui-du-comite-que-je-manage-et-de-mon-referent-je-ne-recois-pas


src/Entity/Projection/ReferentManagedUserFactory.php
```diff
@@ -12,6 +12,7 @@ public function createFromArray(array $data): ReferentManagedUser
             $data['original_id'],
             $data['email'],
             $data['postal_code'],
+            isset($data['committee_postal_code']) ? $data['committee_postal_code'] : null,
```
Remg: `$data['committee_postal_code'] ?? : null,`

src/Entity/Projection/ReferentManagedUser.php
```diff
@@ -73,6 +73,15 @@ class ReferentManagedUser
      */
     private $postalCode;
 
+    /**
+     * @var string|null
+     *
+     * The postal code is filled only for committee supervisors
```
cacahouwete: Can we have description first in comment bloc ?

PR #2237, opened 2018-02-27
[ch3001] Fix display in "my account" for un-adherent

cacahouwete:
https://app.clubhouse.io/enmarche/story/3001/la-partie-adherente-depuis-x-devrait-etre-visible-uniquement-si-la-personne-est-adherente

### DIFF
--- a/templates/user/my_account.html.twig
+++ b/templates/user/my_account.html.twig
@@ -8,7 +8,7 @@
         <span class="settings__username"><strong>Email</strong> {{ app.user.emailaddress }}</span><br />
         <span class="settings__username"><strong>Code postal</strong> {{ app.user.postAddress.postalCode }}</span><br/>
 
-        {% if app.user.activatedAt %}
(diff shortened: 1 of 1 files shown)


templates/user/my_account.html.twig
```diff
@@ -8,7 +8,7 @@
         <span class="settings__username"><strong>Email</strong> {{ app.user.emailaddress }}</span><br />
         <span class="settings__username"><strong>Code postal</strong> {{ app.user.postAddress.postalCode }}</span><br/>
 
-        {% if app.user.activatedAt %}
+        {% if has_role_adherent and app.user.activatedAt %}
```
Remg: just wondering if `{% if has_role_adherent %}` wouldn't be enough here?
can a `User` be `Adherent` without prior activation ?

cacahouwete: There are a test where he check if an adherent without activation can see the page as a non adherent. It s weird :s

PR #2238, opened 2018-02-27
Update of Adherent with Admin must trigger RabbitMQ events

B-Galati:
https://app.clubhouse.io/enmarche/story/2990/une-modification-de-l-user-sur-l-admin-n-entraine-pas-une-modification-du-user-sur-l-api

### DIFF
--- a/src/Admin/AdherentAdmin.php
+++ b/src/Admin/AdherentAdmin.php
@@ -12,6 +12,8 @@ use AppBundle\Form\EventListener\BoardMemberListener;
 use AppBundle\Form\EventListener\CoordinatorManagedAreaListener;
 use AppBundle\Form\GenderType;
 use AppBundle\Intl\UnitedNationsBundle;
+use AppBundle\Membership\UserEvent;
+use AppBundle\Membership\UserEvents;
 use Misd\PhoneNumberBundle\Form\Type\PhoneNumberType;
 use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\ListMapper;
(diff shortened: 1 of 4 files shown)


src/Admin/AdherentAdmin.php
```diff
@@ -384,6 +396,30 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
         ;
     }
 
+    /**
+     * @param Adherent $adherent
+     */
+    public function postPersist($adherent)
```
Remg: This is not needed.
Adherents can't be created through backoffice.

src/Admin/AdherentAdmin.php
```diff
@@ -384,6 +396,30 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+     */
+    public function postUpdate($adherent)
+    {
+        $this->dispatcher->dispatch(UserEvents::USER_UPDATED, new UserEvent($adherent));
+    }
+
+    /**
+     * @param Adherent $adherent
+     */
+    public function postRemove($adherent)
```
Remg: Same here, an Adherent can not be deleted through backoffice.

B-Galati:
@Remg Fixed, thanks a lot ;-)

src/Admin/AdherentAdmin.php
```diff
@@ -384,6 +397,14 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
         ;
     }
 
+    /**
+     * @param Adherent $adherent
+     */
+    public function postUpdate($adherent)
```
Remg: method signature should not change

B-Galati: I read about this yesterday and PHP does not enforce the name of parameters. It only enforces the number of **required** parameter and their types. Thus, you can add any optional argument as you want and also add return type if the parent method does not specify one.

Anyway I'll change it as you proposed because I am not a big fan either :+1:

PR #2251, opened 2018-03-02
[ch2850] Fix fk contraint when we want delete a committee in BO

cacahouwete:
- add set null on delete for event committee link.

https://app.clubhouse.io/enmarche/story/2850/corriger-le-bug-qui-emp%C3%AAche-la-suppression-de-comit%C3%A9s-en-prod


ottaviano:
is it one hotfix? -> `master`

B-Galati:
@cacahouwete @thomasbisignani @ottaviano Why not completely delete the event instead ?

cacahouwete:
image

PR #2268, opened 2018-03-12
Remove autodeploys from circleci

atamano:


.circleci/config.yml
```diff
@@ -169,7 +169,6 @@ jobs:
                 set -x
 
                 bash scripts/build.sh ${CIRCLE_BRANCH}
```
B-Galati: Everything can be put on 1 line now :-) No need for the `set -x` anymore

thomasbisignani:
:medal_sports:

.circleci/config.yml
B-Galati: @atamano The pipe is also useless now ;-)

```
- run: bash scripts/build.sh ${CIRCLE_BRANCH}
```

PR #2276, opened 2018-03-14
fix iphone 5 & js export

Charlesdoiron:


front/style/pages/_grande-marche-europe.scss
```diff
@@ -138,6 +139,10 @@
                 flex-direction: column;
             }
 
+            @media (max-width: $mobile) {
```
marineb: This could be added to the `@media (max-width: $mobile) {` above, non ?

PR #2294, opened 2018-03-20
Sync Commitee & Event with api

Remg:


src/Event/EventCreatedEvent.php
```diff
@@ -2,37 +2,6 @@
...
 namespace AppBundle\Event;
 
-use AppBundle\Committee\CommitteeEvent as BaseEvent;
-use AppBundle\Entity\Adherent;
-use AppBundle\Entity\Committee;
-use AppBundle\Entity\Event;
-use AppBundle\Geocoder\GeocodableInterface;
-
-class EventCreatedEvent extends BaseEvent
+class EventCreatedEvent extends EventBaseEvent
```
B-Galati: As a general comment for the PR : an "event object" should not be aware of the type of event (=event name).

The goal of an event object is to carry the information about the subject of the dispatched event.

I.e. :

```
$this->dispatcher->dispatch(Events::USER_CREATED, new UserEvent($user));
$this->dispatcher->dispatch(Events::USER_UPDATED, new UserEvent($user));
```

At the end we should have one event object that carries a committee and one event object that carries an event (related to EM stuff).

src/Entity/Event.php
```diff
@@ -31,6 +32,9 @@ class Event extends BaseEvent implements UserDocumentInterface
      */
     private $committee;
 
+    /**
+     * @JMS\Exclude
```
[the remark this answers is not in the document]
Remg: Because JMS tries to serialize this from the parent, and this does not exist in the parent, and it's a weird behavior and this is the only workaround i found.

This field is doctrine specific so JMS shouldn't know about it.

thomasbisignani:
Just don't forget to add some tests :1st_place_medal:

src/Entity/Event.php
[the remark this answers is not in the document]
B-Galati: Alright !

src/Entity/BaseEvent.php
```diff
@@ -53,6 +54,9 @@
      * @ORM\Column(length=100)
      *
      * @Algolia\Attribute
+     *
+     * @JMS\Groups({"public"})
+     * @JMS\SerializedName("name")
```
ottaviano: Why do you need to use SerializedName ? It isn't by default the property name  ?

Remg: indeed. just a habit to explicitely declare this.
I will remove all default/useless declaration

PR #2295, opened 2018-03-20
[ch3206] Add a lien in referent space

cacahouwete:
https://app.clubhouse.io/enmarche/story/3206/ajouter-un-lien-dans-l-espace-r%C3%A9f%C3%A9rent

image

### DIFF
--- a/templates/referent/_layout.html.twig
+++ b/templates/referent/_layout.html.twig
@@ -39,6 +39,11 @@
                             </a>
                         </li>
                     {% endfor %}
+                    <li>
+                        <a href="https://en-marche.helprace.com/" target="_blank">
+                            Vos retours <i class="fa fa-external-link" aria-hidden="true"></i>
+                        </a>
+                    </li>
                 </ul>
 
                 <hr />


front/style/pages/_referent.scss
```diff
@@ -1,3 +1,16 @@
+.remark {
```
marineb: We're using the BEM styleguide for CSS. So you shouldn't just add a class on its own. More here : http://getbem.com/naming/

front/style/pages/_referent.scss
```diff
@@ -1,3 +1,16 @@
...
+    text-align: right;
+    width: 90%;
+    position: absolute;
+    margin: 0 auto;
+    a {
```
marineb: Add a space above

front/style/pages/_referent.scss
```diff
@@ -1,3 +1,16 @@
...
+        background: #2ABAFF;
```
marineb: We're using variables for colors. So this should be `$blue`.

templates/referent/_layout.html.twig
```diff
@@ -6,6 +6,9 @@
 {% block content %}
     <main>
         <section class="referent procuration-manager b__nudge--bottom-huge">
```
marineb: I don't understand. Why is there a `procuration-manager` class in there ?

marineb:
@cacahouwete @guillaumedenhez and I looked at the interface. Probably better if it's with the other links underneath with `envoyer un message`, `evenements`, etc. @guillaumedenhez thoughts the language could change to something shorter such as `Vos retours` with an font awesome icon of something that opens in a new window https://fontawesome.com/v4.7.0/icon/external-link/

If that's too much, I can help :P

PR #2297, opened 2018-03-21
Add Session TTL with Remember me feature

ottaviano:
image

### DIFF
--- a/src/Security/AdherentLoginTimestampRecorder.php
+++ b/src/Security/AdherentLoginTimestampRecorder.php
@@ -4,9 +4,7 @@ namespace App\Security;
 
 use App\Entity\Adherent;
 use Doctrine\Common\Persistence\ObjectManager;
-use Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token\JWTUserToken;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken;
 use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
 use Symfony\Component\Security\Http\SecurityEvents;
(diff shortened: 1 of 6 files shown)


src/Security/AdherentLoginTimestampRecorder.php
```diff
@@ -28,8 +28,8 @@ public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
             return;
         }
 
-        if (!$token instanceof UsernamePasswordToken) {
-            throw new \RuntimeException(sprintf('Authentication token must be a %s instance.', UsernamePasswordToken::class));
+        if (!$token instanceof TokenInterface) {
```
B-Galati: This test is basically useless now as `$event` is always returning a `TokenInterface`.
Let's remove it no ?

B-Galati:
@ottaviano Perhaps the checkbox should be checked by default

templates/security/adherent_login.html.twig
```diff
@@ -27,7 +27,12 @@
 
                     <a href="{{ path('forgot_password') }}" class="form__forgotten-pw">Oublié ?</a>
                 </li>
-
+                <li>
+                    <div class="b__nudge--top text--small form__checkbox form__checkbox--large" id="field-conditions">
+                        <input type="checkbox" id="remember_me" name="_remember_me"/>
+                        <label class="form form__label" for="remember_me">Se souvenir de  moi</label>
```
thomasbisignani: There is a useless whitespace in the text

templates/security/adherent_login.html.twig
```diff
@@ -27,7 +27,12 @@
 
                     <a href="{{ path('forgot_password') }}" class="form__forgotten-pw">Oublié ?</a>
                 </li>
-
+                <li>
+                    <div class="b__nudge--top text--small form__checkbox form__checkbox--large" id="field-conditions">
```
thomasbisignani: The `field-conditions` id is really needed here ?

app/migrations/Version20180321121625.php
```diff
@@ -0,0 +1,25 @@
...
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20180321121625 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('CREATE TABLE `rememberme_token` (
+            `series` char(88) UNIQUE PRIMARY KEY NOT NULL,
+            `value` char(88) NOT NULL,
+            `lastUsed` datetime NOT NULL,
```
hhamon: `last_used` no?

templates/security/adherent_login.html.twig
ottaviano: fixed, thank yoy @thomasbisignani :D

templates/security/adherent_login.html.twig
ottaviano: nop, removed 👍

app/migrations/Version20180321121625.php
ottaviano: no, we should keep it in camelCase, it is used here: https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/Security/RememberMe/DoctrineTokenProvider.php#L84

illusionOfParadise (approved):

PR #2299, opened 2018-03-21
improve 404

Charlesdoiron:
• Add Home Button 
• Add Header
• Add Footer

### DIFF
--- a/templates/TwigBundle/Exception/error404.html.twig
+++ b/templates/TwigBundle/Exception/error404.html.twig
@@ -11,7 +11,7 @@
                 <h1>
                     Cette page n'existe pas !
                 </h1>
-                <p class="text--body text--black">
+                <p class="text--body">
                     Vous trouverez peut-être ce que vous cherchiez sur notre page d'accueil.
                 </p>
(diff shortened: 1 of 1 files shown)


templates/TwigBundle/Exception/error404.html.twig
```diff
@@ -2,31 +2,16 @@
...
-            </h1>
-        </div>
         <section class="l__wrapper--slim">
             <div class="error__box error__box--404"></div>
             <div>
                 <h1>
                     Cette page n'existe pas !
                 </h1>
-                <p class="text--body text--gray">
+                <p class="text--body text--black">
```
marineb: You don't need `text--black` here since the text is by default black .

Charlesdoiron: https://app.clubhouse.io/enmarche/story/557/etq-utilisateur-je-ne-veux-que-la-page-404-ne-soit-pas-un-cul-de-sac-je-veux-pouvoir-revenir-facilement-a-en-marche-et-savoir

Ok for the text--black

PR #2301, opened 2018-03-26
[ch3248] Change google analytics location

cacahouwete:
https://app.clubhouse.io/enmarche/story/3248/corriger-placement-des-balises-de-tracking-sur-toutes-les-pages


templates/base.html.twig
```diff
@@ -41,6 +41,15 @@
         {% block extra_link '' %}
 
         <title>{% block title %}{% block page_title '' %} | {% block app_title 'La République En Marche !' %}{% endblock %}</title>
+        {% if app.environment == 'prod' %}
+            <!-- Google Tag Manager -->
+            <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
B-Galati: Should we add this script at the end of the page to avoid slowing down the page loading ?
Same question for the other one.

JulHb: Nop, as high as possible actually, to let the tracker be loaded before the tracked elements.

B-Galati: Alright :+1:

PR #2304, opened 2018-03-27
Add Mailer consumer for scheduled mails

ottaviano:
the test is coming

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -295,6 +295,13 @@ old_sound_rabbit_mq:
             callback:          AppBundle\Consumer\ProjectCitizenCreationNotificationConsumer
             qos_options:       { prefetch_size: 0, prefetch_count: 1, global: false }
 
+        mailer:
+            connection:        default
+            exchange_options:  { name: 'mails', type: direct }
+            queue_options:     { name: 'mails', durable: false }
+            callback:          AppBundle\Consumer\MailerConsumer
+            qos_options:       { prefetch_size: 0, prefetch_count: 1, global: false }
+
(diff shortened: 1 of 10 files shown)


app/config/config.yml
```diff
@@ -295,6 +295,13 @@ old_sound_rabbit_mq:
             callback:          AppBundle\Consumer\ProjectCitizenCreationNotificationConsumer
             qos_options:       { prefetch_size: 0, prefetch_count: 1, global: false }
 
+        mailer:
+            connection:        default
+            exchange_options:  { name: 'mails', type: direct }
+            queue_options:     { name: 'mails', durable: false, routing_keys: [mailer.scheduled_mail]}
```
B-Galati: Why do we need to specify a routing key ? it could be anything no ?

src/Consumer/AbstractConsumer.php
```diff
@@ -76,10 +76,15 @@ protected function getManager(): EntityManagerInterface
         return $this->manager;
     }
 
+    protected function allowExtraFields(): bool
```
B-Galati: `isExtraFieldsAllowed` would be a better name IMO.
This one looks an "action" and/or "behavior" if you know what I mean.

Remg: i know what you mean.

ottaviano: agree ;)

app/config/config.yml
ottaviano: removed

thomasbisignani:
:medal_sports:

PR #2308, opened 2018-03-28
[ch3011] Fix button visual and text and Canary part of the feature

cacahouwete:
Add a few visual and language tweaks to the notification page
Add a canary

https://app.clubhouse.io/enmarche/story/3011/am%C3%A9liorations-de-la-gestion-des-notifications

### DIFF
--- a/src/Entity/Adherent.php
+++ b/src/Entity/Adherent.php
@@ -452,20 +452,47 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
         if ($this->hasCitizenProjectCreationEmailSubscription()) {
             $subscriptions[] = AdherentEmailSubscription::SUBSCRIBED_EMAILS_CITIZEN_PROJECT_CREATION;
         }
+//      CANARY !
+        if ($this->localHostEmailsSubscription) {
+            $subscriptions[] = AdherentEmailSubscription::SUBSCRIBED_EMAILS_LOCAL_HOST;
+        }
 
         return $subscriptions;
     }
 
     public function setEmailsSubscriptions(array $emailsSubscriptions): void
     {
-        if ($key = array_search(AdherentEmailSubscription::SUBSCRIBED_EMAILS_CITIZEN_PROJECT_CREATION, $emailsSubscriptions, true)) {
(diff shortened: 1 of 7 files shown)


front/style/pages/_profile.scss
```diff
@@ -134,3 +134,9 @@
         }
     }
 }
+
+#adherent_email_subscription_emails_subscriptions {
```
marineb: We don't generally use IDs in CSS. Need to use a class. Also if it's just a margin, you can use one of our reusable classes.

marineb:
@cacahouwete we should put all this in canary mode for now since it's not connected to anything. Also I couldn't find where you're infusing the `h3`. Maybe you can tell me on Monday. Thank you!

templates/user/set_email_notifications.html.twig
```diff
@@ -2,20 +2,30 @@
...
-
         <div class="b__nudge--top">
             {{ form_widget(form.submit, { attr: { class: 'btn--blue' } }) }}
         </div>
 
+        <p class="text--small">
+            Pour ne plus recevoir les notifications des <strong>animateurs de comités</strong> ou de <strong>porteurs de projets citoyens</strong>, rendez-vous sur leurs pages respectives et arretez de les suivre.
+        </p>
+
+
```
illusionOfParadise: An extra line

src/Entity/Adherent.php
```diff
@@ -452,20 +452,45 @@ public function getEmailsSubscriptions(): array
         if ($this->hasCitizenProjectCreationEmailSubscription()) {
             $subscriptions[] = AdherentEmailSubscription::SUBSCRIBED_EMAILS_CITIZEN_PROJECT_CREATION;
         }
+//      CANARY !
+        if ($this->localHostEmailsSubscription) {
+            $subscriptions[] = AdherentEmailSubscription::SUBSCRIBED_EMAILS_LOCAL_HOST;
```
ottaviano: `LOCAL_HOST` => `LOCALHOST` :trollface:

cacahouwete: :D I know ^^, it's one of the old fields removed ^^,

thomasbisignani:
why don't using the `CanaryControllerTrait` to disable code in production ?

PR #2313, opened 2018-04-03
Test Api synchronization commands

Remg:


thomasbisignani:
:man_teacher: :1st_place_medal:

PR #2314, opened 2018-04-03
Improve btn Delete & Modify on Comitee's page

Charlesdoiron:


front/style/layout/_buttons.scss
```diff
@@ -444,9 +444,11 @@ input {
     }
 }
 
-button.btn {
-    display: inline-block;
-    padding: 10px 15px 9px;
+button {
+    .btn {
```
marineb: Why change this? This is used in a lot of places.

Charlesdoiron: As we said, I just removed this piece of code.
💥

PR #2322, opened 2018-04-05
Handle ClientException when sending an email

Remg:


src/Consumer/AbstractMailerConsumer.php
```diff
@@ -42,6 +43,14 @@ protected function doExecute(array $data): int
             }
 
             return $delivered ? ConsumerInterface::MSG_ACK : ConsumerInterface::MSG_REJECT_REQUEUE;
+        } catch (ClientException $clientException) {
+            $this->writeln($data['uuid'], 'Client Exception');
+            $this->getLogger()->error(
+                'Error while sending a mail with UUID '.$data['uuid'],
```
B-Galati: It would be nice to have the exception message in the log message too.

What do you think if we replace the beginning of the message by `Http client error` instead of `Error` ?

PR #2329, opened 2018-04-06
[ch3324] Modify errors message in adhesion page.

cacahouwete:
https://app.clubhouse.io/enmarche/story/3324/modifier-les-messages-d-erreur-sur-la-page-d-adh%C3%A9sion

### DIFF
--- a/templates/membership/join.html.twig
+++ b/templates/membership/join.html.twig
@@ -63,7 +63,7 @@
                                         {{ form_errors(form.lastName) }}
                                     </div>
                                 </div>
-                                 <div class="l__row l__row--h-stretch register__form--full">
+                                 <div class="l__row--h-stretch register__form--full">
                                      {{ form_widget(form.emailAddress.first, { attr: { placeholder: 'E-mail' } }) }}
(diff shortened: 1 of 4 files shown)


src/Controller/EnMarche/MembershipController.php
```diff
@@ -134,6 +137,7 @@ private function joinAdherent(Request $request): Response
         return $this->render('membership/join.html.twig', [
             'membership' => $membership,
             'form' => $form->createView(),
+            'formIsValid' => $isValid,
```
B-Galati: As an alternative we can do `$isValid ?? null;` and then remove the line 122 which init the var.

templates/membership/join.html.twig
```diff
@@ -163,6 +163,9 @@
                         {% endif %}
 
                         <div class="register__form--full">
+                            {% if formIsValid is same as(false) %}
```
B-Galati: What about using form.vars.valid instead ?

cacahouwete: I didn't see this var. I think it's better than doing this trix ^^,. Thx

PR #2344, opened 2018-04-11
[ch3092] Create an instant validation for multiple form

cacahouwete:
- create a route api to dynamicly validate form send
- create js to hand any form and display errors

https://app.clubhouse.io/enmarche/story/3092/int%C3%A9grer-l-instant-field-validation-sur-les-formulaires-de-don-et-d-adh%C3%A9sion

### DIFF
--- a/front/pages/join.js
+++ b/front/pages/join.js
@@ -1,6 +1,8 @@
 import validateEmail from '../validator/emailValidator';
+import formValidator from '../validator/formValidator';
 
-export default () => {
+export default (formType) => {
+    const form = dom('form[name="adherent_registration"]') || dom('form[name="become_adherent"]');
     const emailField = dom('#adherent_registration_emailAddress_first');
     const confirmEmailField = dom('#adherent_registration_emailAddress_second');
     let zipCodeField = dom('#adherent_registration_address_postalCode');
@@ -45,4 +47,6 @@ export default () => {
 
     on(zipCodeField, 'input', displayCaptcha);
     zipCodeField.dispatchEvent(new Event('input'));
+
+    formValidator(formType, form);
 };
--- a/front/pages/registration.js
+++ b/front/pages/registration.js
@@ -1,6 +1,8 @@
 import validateEmail from '../validator/emailValidator';
+import formValidator from '../validator/formValidator';
 
-export default () => {
+export default (formType) => {
+    const form = dom('form[name="user_registration"]');
     const emailField = dom('#user_registration_emailAddress_first');
     const confirmEmailField = dom('#user_registration_emailAddress_second');
     const zipCodeField = dom('#user_registration_address_postalCode');
@@ -36,4 +38,6 @@ export default () => {
 
     emailField.dispatchEvent(new Event('input'));
     zipCodeField.dispatchEvent(new Event('input'));
+
+    formValidator(formType, form);
 };
--- a/front/validator/formValidator.js
+++ b/front/validator/formValidator.js
@@ -1,0 +1,57 @@
+import reqwest from 'reqwest';
+
+function displayErrors(node, id) {
+    if (Object.prototype.hasOwnProperty.call(node, 'children')) {
+        Object.keys(node.children).forEach((key) => {
+            displayErrors(node.children[key], `${id}_${key}`);
+        });
+    }
+    let htmlErrors = dom(`#${id}_errors`);
+    if (Object.prototype.hasOwnProperty.call(node, 'errors')) {
+        if (!htmlErrors) {
+            const htmlInput = dom(`#${id}`);
+            if (!htmlInput) {
+                return;
+            }
+            htmlInput.parentNode.insertAdjacentHTML(
+                'beforeend',
+                `<ul id="${id}_errors" class="form form__errors"></ul>`
+            );
+            htmlErrors = dom(`#${id}_errors`);
+        }
+        htmlErrors.innerHTML = '';
+        node.errors.forEach((value) => {
+            htmlErrors.insertAdjacentHTML('beforeend', `<li class="form__error">${value}</li>`);
+        });
+    } else if (htmlErrors) {
+        htmlErrors.parentNode.removeChild(htmlErrors);
+    }
+}
+
+export default (formType, form) => {
+    formType = encodeURI(formType);
+    form.addEventListener('change', () => {
+        const url = `/api/form/validate/${formType}`;
+
+        const formData = new FormData(form);
+        const emptyFieldNames = [];
+        for (const field of formData.entries()) {
+            if (!field[1]) {
+                emptyFieldNames.push(field[0]);
+            }
+        }
+        for (const fieldName of emptyFieldNames) {
+            formData.delete(fieldName);
+        }
+
+        reqwest({
+            url,
+            method: 'post',
+            data: formData,
+            processData: false,
+            success(resp) {
+                displayErrors(resp, form.name);
+            },
+        });
+    });
+};
--- a/src/Controller/Api/FormController.php
+++ b/src/Controller/Api/FormController.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace AppBundle\Controller\Api;
+
+use JMS\Serializer\SerializerInterface;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormTypeInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @Route("/form")
+ */
+class FormController extends Controller
+{
+    /**
+     * @Route("/validate/{formType}", name="api_form_validate", condition="request.isXmlHttpRequest()")
+     * @Method("POST")
+     */
+    public function validate(Request $request, SerializerInterface $serializer, string $formType): Response
+    {
+        if (!is_subclass_of($formType, FormTypeInterface::class)) {
+            throw $this->createNotFoundException('Form not found');
+        }
+
+        $form = $this->createForm($formType, null, ['csrf_protection' => false]);
+
+        if (!$request->request->has($form->getName())) {
+            throw new BadRequestHttpException(sprintf('No parameters with key "%s" founded.', $form->getName()));
+        }
+
+        $form->submit($request->request->get($form->getName()), false)->isValid();
+
+        // We use jms serializer handler for errors.
+        return new Response($serializer->serialize(
+            [
+                'children' => array_filter($form->all(), function (FormInterface $form) {
+                    return $form->isSubmitted() && !$form->isValid();
+                }),
+            ],
+            'json'
+        ));
+    }
+}
--- a/src/Donation/DonationRequestUtils.php
+++ b/src/Donation/DonationRequestUtils.php
@@ -7,6 +7,7 @@ use AppBundle\Entity\Donation;
 use AppBundle\Exception\InvalidDonationCallbackException;
 use AppBundle\Exception\InvalidDonationPayloadException;
 use AppBundle\Exception\InvalidDonationStatusException;
+use AppBundle\Exception\InvalidPayboxPaymentSubscriptionValueException;
 use Ramsey\Uuid\Uuid;
 use Symfony\Component\DependencyInjection\ServiceLocator;
 use Symfony\Component\HttpFoundation\Request;
@@ -52,9 +53,15 @@ class DonationRequestUtils
     /**
      * @throws InvalidDonationPayloadException
      */
-    public function createFromRequest(Request $request, float $amount, int $duration, ?Adherent $currentUser): DonationRequest
+    public function createFromRequest(Request $request, ?Adherent $currentUser): DonationRequest
     {
         $clientIp = $request->getClientIp();
+        $amount = (float) $request->query->get('montant');
+        $duration = $request->query->getInt('abonnement', PayboxPaymentSubscription::NONE);
+
+        if (!PayboxPaymentSubscription::isValid($duration)) {
+            throw new InvalidPayboxPaymentSubscriptionValueException($duration);
+        }
 
         if ($currentUser) {
             $donation = DonationRequest::createFromAdherent($currentUser, $clientIp, $amount, $duration);
--- a/src/Exception/InvalidPayboxPaymentSubscriptionValueException.php
+++ b/src/Exception/InvalidPayboxPaymentSubscriptionValueException.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace AppBundle\Exception;
+
+class InvalidPayboxPaymentSubscriptionValueException extends \InvalidArgumentException
+{
+    public function __construct(int $duration, \Throwable $previous = null)
+    {
+        parent::__construct(sprintf('Duration value "%d" not supported.', $duration), 0, $previous);
+    }
+}
--- a/src/Form/AdherentRegistrationType.php
+++ b/src/Form/AdherentRegistrationType.php
@@ -40,13 +40,16 @@ class AdherentRegistrationType extends AbstractType
         $builder->addEventListener(FormEvents::POST_SET_DATA, function (FormEvent $formEvent) {
             /** @var MembershipRequest $membershipRequest */
             $membershipRequest = $formEvent->getData();
-            $membershipRequest->comMobile = $membershipRequest->comEmail;
+            if ($membershipRequest) {
+                $membershipRequest->comMobile = $membershipRequest->comEmail;
+            }
         });
     }
 
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
+            'data_class' => MembershipRequest::class,
             'validation_groups' => ['Update', 'Conditions', 'Registration'],
         ]);
     }
--- a/src/Form/DonationRequestType.php
+++ b/src/Form/DonationRequestType.php
@@ -3,6 +3,8 @@
 namespace AppBundle\Form;
 
 use AppBundle\Donation\DonationRequest;
+use AppBundle\Donation\DonationRequestUtils;
+use AppBundle\Entity\Adherent;
 use AppBundle\Form\DataTransformer\FloatToStringTransformer;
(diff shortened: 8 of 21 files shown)


src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+class FormController extends Controller
+{
+    /**
+     * @Route("/validate/{formType}", name="api_form_validate")
+     * @Method("POST")
+     */
+    public function validate(Request $request, Serializer $serializer, string $formType): Response
+    {
+        $form = $this->createForm($formType, null, ['csrf_protection' => false]);
```
B-Galati: I am not sure it's a good idea to instantiate something provided by the client.. 
I would prefer a whitelisting approach to avoid any problem.
Also, we could create a mapping of form to avoid passing the whole form type class in the path.
E.g. `/validate/become-adherent` would map to `BecomeAdherentType::class`

ottaviano: maybe do it with custom `ParamConvertor` [doc](https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/converters.html#creating-a-converter)

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,52 @@
...
+
+        $form->submit($params = $request->request->all()[$form->getConfig()->getName()], false)->isValid();
+        $errors = $form->getErrors(true);
+
+        return new Response($serializer->serialize($errors, 'json'));
```
ottaviano: what about use `JsonResponse` class?

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * @Route("/validate/{formType}/fields", name="api_form_validate_fields")
+     * @Method("POST")
+     */
+    public function validateField(Request $request, ValidatorInterface $validator, Serializer $serializer, string $formType): Response
```
ottaviano: where is used this action?

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * @Route("/validate/{formType}", name="api_form_validate")
+     * @Method("POST")
+     */
+    public function validate(Request $request, Serializer $serializer, string $formType): Response
+    {
+        $form = $this->createForm($formType, null, ['csrf_protection' => false]);
+
+        $form->submit($params = $request->request->all()[$form->getConfig()->getName()], false)->isValid();
```
ottaviano: why do you need `$params`

templates/donation/informations.html.twig
```diff
@@ -1,5 +1,14 @@
 {% extends 'donation/layout.html.twig' %}
 
+{% block javascripts %}
+    {% import 'javascript.js.twig' as js %}
```
ottaviano: why this import?

templates/membership/register.html.twig
```diff
@@ -12,7 +12,7 @@
     {% import 'javascript.js.twig' as js %}
     <script type="text/javascript">
         Kernel.onLoad(function() {
-            App.runRegistration();
+            App.runRegistration('{{ form_type|replace({'\\': '\\\\'}) }}');
```
ottaviano: `form_type|js`?

src/Controller/Api/FormController.php
cacahouwete: Yes I was thinking about that

tests/Controller/Api/FormControllerTest.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Tests\AppBundle\Controller\Api;
+
+use AppBundle\Form\AdherentRegistrationType;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Tests\AppBundle\Controller\ApiControllerTestTrait;
+use Tests\AppBundle\Controller\ControllerTestTrait;
+use Tests\AppBundle\MysqlWebTestCase;
+
+class FormControllerTest extends MysqlWebTestCase
```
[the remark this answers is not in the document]
cacahouwete: I was thinking phpunit was much easier to do it. I've never done behat api test.

tests/Controller/Api/FormControllerTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Tests\AppBundle\Controller\Api;
+
+use AppBundle\Form\AdherentRegistrationType;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Tests\AppBundle\Controller\ApiControllerTestTrait;
+use Tests\AppBundle\Controller\ControllerTestTrait;
+use Tests\AppBundle\MysqlWebTestCase;
+
```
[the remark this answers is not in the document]
cacahouwete: done

front/pages/join.js
```diff
@@ -1,6 +1,8 @@
 import validateEmail from '../validator/emailValidator';
+import formValidator from '../validator/formValidator';
 
-export default () => {
+export default (formType) => {
+    const form = dom('[name="adherent_registration"]') || dom('[name="become_adherent"]');
```
ottaviano: 1) `dom('form[name="adherent_registration"]')`
2) why `|| dom(...)`? check in this JS file you work with `#adherent_registration_` fields

front/pages/registration.js
```diff
@@ -1,6 +1,8 @@
 import validateEmail from '../validator/emailValidator';
+import formValidator from '../validator/formValidator';
 
-export default () => {
+export default (formType) => {
+    const form = dom('[name="user_registration"]');
```
ottaviano: `dom('form[name="user_registration"]')`

src/Controller/Api/FormController.php
B-Galati: Not a big fan. It adds an overhead for all requests of all the website for only 2 routes. IMO it's not a good use case for the moment.

front/validator/formValidator.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export default (formType, form) => {
+    formType = encodeURI(formType);
+    form.addEventListener('change', () => {
+        const url = `/api/form/validate/${formType}`;
+
+        const formData = new FormData(form);
+        formData.toRemove = [];
+        formData.forEach((value, key, $this) => {
```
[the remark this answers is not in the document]
cacahouwete: I can't remove in a loop, I have to do it in a second loop. If I do it in same loop I check there are a value he will escape some fields :s

src/Controller/Api/FormController.php
cacahouwete: to check each params send with validator

front/pages/join.js
cacahouwete: there are 2 possible form in this view :s

src/Controller/Api/FormController.php
cacahouwete: remove

templates/membership/register.html.twig
cacahouwete: this filter does not exist , are you sure ?

src/Controller/Api/FormController.php
cacahouwete: Not work :s

src/Controller/Api/FormController.php
cacahouwete: for now I will make a simple map with key value

cacahouwete: And if I have the time, create an interface to map all formtype exposed with a static method to get his key

templates/membership/register.html.twig
B-Galati: He meant `e('js')`

src/Controller/Api/FormController.php
B-Galati: There is no need for the formtype to be aware of this thing so the mapping is more than fine.

B-Galati:
@cacahouwete What about my review comment ?

salahm:
Be careful, data (and validation) should be consistent between front and backend. Also, i don't think it's a good practice to implement two different validation logic.

B-Galati:
@salahm Yeap :+1: 

I think we gonna install something like API Platform on this project because we are going to have some pages entirely rendered with React. So this system will be soon deprecated.

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,46 @@
...
+    // TODO: Replace this by a compiler pass who crawl all FormTypeApiExposeInterface to add key and class in a FormTypeApiExposeManager
+    private const FORM_MAP = [
+        'become-adherent' => BecomeAdherentType::class,
+        'adherent-registration' => AdherentRegistrationType::class,
+        'user-registration' => UserRegistrationType::class,
+        'donation-subscription' => DonationSubscriptionRequestType::class,
+    ];
+
+    /**
+     * @Route("/validate/{formType}", name="api_form_validate")
```
HeahDude: Missing `condition="request.isXmlHttpRequest"`

src/Form/AdherentRegistrationType.php
```diff
@@ -55,4 +58,9 @@ public function getParent()
     {
         return AdherentType::class;
     }
+
+    public static function getExposeKey(): string
```
HeahDude: Should be removed. Used `form.vars.name` value instead.

templates/membership/join.html.twig
```diff
@@ -17,7 +17,7 @@
     {% import 'javascript.js.twig' as js %}
     <script type="text/javascript">
         Kernel.onLoad(function() {
-            App.runJoin();
+            App.runJoin('{{ form_type }}');
```
HeahDude: `form.vars.name`

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @Route("/form")
+ */
+class FormController extends Controller
+{
+    // TODO: Replace this by a compiler pass who crawl all FormTypeApiExposeInterface to add key and class in a FormTypeApiExposeManager
+    private const FORM_MAP = [
```
HeahDude: After all you should not need this map. "internal" urls do not have to be pretty, and you should better use the FQCN of the types as a POST parameter.

HeahDude: And no need for a compiler pass either ;).

src/Controller/EnMarche/MembershipController.php
```diff
@@ -59,6 +59,7 @@ public function registerAction(Request $request, GeoCoder $geoCoder, Authorizati
             'membership' => $membership,
             'form' => $form->createView(),
             'countries' => UnitedNationsBundle::getCountries($request->getLocale()),
+            'form_type' => UserRegistrationType::getExposeKey(),
```
HeahDude: To remove.

src/Form/FormTypeApiExposeInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace AppBundle\Form;
+
+interface FormTypeApiExposeInterface
```
HeahDude: To remove.

front/validator/formValidator.js
```diff
@@ -0,0 +1,58 @@
...
+        node.errors.forEach((value) => {
+            htmlErrors.insertAdjacentHTML('beforeend', `<li class="form__error">${value}</li>`);
+        });
+    } else if (htmlErrors) {
+        htmlErrors.parentNode.removeChild(htmlErrors);
+    }
+}
+
+export default (formType, form) => {
+    console.log(formType);
```
HeahDude: :trollface:

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function validate(Request $request, Serializer $serializer, string $formType): Response
+    {
+        if (!class_exists($formType)) {
+            throw $this->createNotFoundException('Form not found');
+        }
+
+        $form = $this->createForm($formType, null, ['csrf_protection' => false]);
+
+        $form->submit($params = $request->request->all()[$form->getConfig()->getName()], false)->isValid();
```
HeahDude:
```php
if (!$request->request->has($form->getName()) {
    // throw
}
```

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+class FormController extends Controller
+{
+    /**
+     * @Route("/validate/{formType}", name="api_form_validate", condition="request.isXmlHttpRequest()")
+     * @Method("POST")
+     */
+    public function validate(Request $request, Serializer $serializer, string $formType): Response
+    {
+        if (!class_exists($formType)) {
```
HeahDude: Please add a check for the `FormTypeInterface`

src/Donation/DonationRequestUtils.php
```diff
@@ -52,9 +53,15 @@ public function __construct(ServiceLocator $locator)
...
      * @throws InvalidDonationPayloadException
      */
-    public function createFromRequest(Request $request, float $amount, int $duration, ?Adherent $currentUser): DonationRequest
+    public function createFromRequest(Request $request, ?Adherent $currentUser): DonationRequest
     {
         $clientIp = $request->getClientIp();
+        $amount = (float) $request->query->get('montant');
+        $duration = $request->query->getInt('abonnement', PayboxPaymentSubscription::NONE);
+
+        if (!PayboxPaymentSubscription::isValid($duration)) {
```
ottaviano: this is form validation and it already works with `PayboxSubscriptionValidator` and `@Assert` in `DonationRequest`
```
/**
     * @AssertPayboxSubscription
     */
    private $duration;
```

src/Exception/InvalidPayboxPaymentSubscriptionValueException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace AppBundle\Exception;
+
+use Throwable;
```
ottaviano: useless, use `\Throwable` instead pls

src/Form/DonationRequestType.php
```diff
@@ -68,6 +87,19 @@ public function buildView(FormView $view, FormInterface $form, array $options)
...
+    public function preSetData(FormEvent $formEvent): void
+    {
+        if (!$formEvent->getData()) {
+            $user = $this->tokenStorage->getToken()->getUser();
+
+            if (!$user instanceof Adherent) {
+                $user = null;
+            }
+
+            $formEvent->setData($this->donationRequestUtils->createFromRequest($this->requestStack->getCurrentRequest(), $user));
```
ottaviano: why do you this?

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,42 @@
...
+
+        $form = $this->createForm($formType, null, ['csrf_protection' => false]);
+
+        if (!$request->request->has($form->getName())) {
+            throw new NoParametersExpected(sprintf('No parameters with key "%s" founded.', $form->getName()));
+        }
+
+        $form->submit($params = $request->request->get($form->getName()), false)->isValid();
+
+        return new Response($serializer->serialize(
```
ottaviano: can you optimize the response to reduce the size of response? 

I have this big payload juste for validate one field 
image

src/Form/DonationRequestType.php
cacahouwete: Because it's in preset data we have to init default value if we need to use a factory. If not the form type will try to instantiate 'data_class' and will throw an exception because of missing parameters.

src/Donation/DonationRequestUtils.php
cacahouwete: I just move code from controller to this function.

And I think the goal was to throw an exception if the duration is not valid and not render the form.

src/Controller/Api/FormController.php
cacahouwete: I moved to true the flatten get errors

ottaviano: what is your payload now?

cacahouwete: I have the same payload :s

src/Controller/Api/FormController.php
```diff
@@ -0,0 +1,42 @@
...
+    public function validate(Request $request, Serializer $serializer, string $formType): Response
+    {
+        if (!is_subclass_of($formType, FormTypeInterface::class)) {
+            throw $this->createNotFoundException('Form not found');
+        }
+
+        $form = $this->createForm($formType, null, ['csrf_protection' => false]);
+
+        if (!$request->request->has($form->getName())) {
+            throw new InvalidArgumentException(sprintf('No parameters with key "%s" founded.', $form->getName()));
```
HeahDude: BadRequestHttpException

src/Exception/InvalidPayboxPaymentSubscriptionValueException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace AppBundle\Exception;
+
+class InvalidPayboxPaymentSubscriptionValueException extends \InvalidArgumentException
+{
+    public function __construct(int $duration, int $code = 0, \Throwable $previous = null)
```
HeahDude: No need for `$code` IMO

src/Form/DonationRequestType.php
```diff
@@ -68,6 +87,19 @@ public function buildView(FormView $view, FormInterface $form, array $options)
         }
     }
 
+    public function preSetData(FormEvent $formEvent): void
```
HeahDude: `createInitialDonationRequest()`

src/Controller/Api/FormController.php
```diff
@@ -35,7 +36,11 @@ public function validate(Request $request, Serializer $serializer, string $formT
         $form->submit($request->request->get($form->getName()), false)->isValid();
 
         return new Response($serializer->serialize(
-            $form->getErrors(true),
+            [
+                'children' => array_filter($form->all(), function (FormInterface $form) {
```
HeahDude: Does `'children' => \array_filter($form->getErrors(true))` work? I would cost much less

src/Controller/Api/FormController.php
```diff
@@ -38,8 +38,8 @@ public function validate(Request $request, Serializer $serializer, string $formT
         return new Response($serializer->serialize(
             [
                 'children' => array_filter($form->all(), function (FormInterface $form) {
-                    return 0 < $form->getErrors(true)->count();
```
ottaviano: do you test this? I tried this and it

cacahouwete: Ok I just add a check if the form is submitted ^^,

PR #2353, opened 2018-04-14
[docker] Remove deprecated parameter

jeremychauvet:
Remove a deprecated parameter : pulling multiple images in parallel is now enabled by default.
See https://docs.docker.com/compose/reference/pull/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ deps: vendor web/built                                                          
 # Internal rules
 
 build:
-	$(DOCKER_COMPOSE) pull --parallel --ignore-pull-failures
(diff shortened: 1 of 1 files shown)


B-Galati:
Thank you @jeremychauvet 

I think it's a bit early to remove the parameter as CircleCI does not support the very last version of docker-compose yet.

B-Galati:
See https://circleci.com/docs/2.0/configuration-reference/#machine

jeremychauvet:
Thanks for the feedback @B-Galati :) ! U can close the PR then

PR #2361, opened 2018-04-17
[ch3212] add roles for admin security for organizational chart

cacahouwete:


app/config/security.yml
```diff
@@ -160,6 +160,14 @@ security:
             - ROLE_APP_ADMIN_EMAIL_LIST
             - ROLE_APP_ADMIN_EMAIL_VIEW
 
+        ROLE_ADMIN_ORGANIGRAMM:
+            - ROLE_APP_ADMIN_ORGANIZATIONAL_CHART_ITEM_ADMIN_EXPORT
```
B-Galati: I guess you can replace them all by `- ROLE_APP_ADMIN_ORGANIZATIONAL_CHART_ITEM_ADMIN_ALL`
If this granularity of roles is not needed.

thomasbisignani: I agree with @B-Galati, if you want to grant `LIST, VIEW, CREATE, EDIT, DELETE and EXPORT`, you can use `ALL`.
See https://sonata-project.org/bundles/admin/master/doc/reference/security.html

PR #2363, opened 2018-04-18
Donation : Removed subscription durations to keep only the unlimited

thomasbisignani:


front/components/AmountChooser.js
```diff
@@ -93,6 +93,9 @@ export default class AmountChooser extends React.Component {
                     <input type="checkbox" name="abonnement" id="donation-monthly"/>
                     <label htmlFor="donation-monthly">Je veux donner ce montant chaque mois</label>
                 </div>
+                <div className="amount-chooser__help">
```
B-Galati: `class` instead of `className` no ?

ottaviano: not in [JSX](https://reactjs.org/docs/jsx-in-depth.html)

B-Galati: XD thanks!

PR #2367, opened 2018-04-19
Create endpoint for managed adherent count + by gender

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/3634/1%C3%A8re-partie-section-adh%C3%A9rents-box-kpis-adh%C3%A9rents-managed-area-name

### DIFF
--- a/src/Controller/Api/AdherentsController.php
+++ b/src/Controller/Api/AdherentsController.php
@@ -17,30 +17,35 @@ use Symfony\Component\HttpFoundation\Response;
 class AdherentsController extends Controller
 {
     /**
-     * @Route("/count", name="app_referent_dashboard_users")
+     * @Route("/count", name="app_adherents_count")
      * @Method("GET")
      */
-    public function usersAction(AdherentRepository $adherentRepository): Response
+    public function adherentsCountAction(AdherentRepository $adherentRepository): Response
     {
         $count = $adherentRepository->countByGender();
 
+        return new JsonResponse($this->aggregateCount($count));
+    }
+
+    /**
+     * @Route("/count-by-referent-area", name="app_adherents_count_for_referent_managed_area")
+     * @Method("GET")
+     */
+    public function adherentsCountForReferentManagedAreaAction(AdherentRepository $adherentRepository): Response
+    {
+        $count = $adherentRepository->countByGenderManagedBy($this->getUser());
+
+        return new JsonResponse($this->aggregateCount($count));
+    }
+
+    private function aggregateCount(array $count): array
+    {
         array_walk($count, function (&$item) {
             $item = (int) $item['count'];
         });
 
-        $total = array_sum($count);
+        $count['total'] = array_sum($count);
 
-        array_walk($count, function (&$item) use ($total) {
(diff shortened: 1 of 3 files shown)


src/Repository/AdherentRepository.php
```diff
@@ -499,6 +499,28 @@ public function countByGender(): array
...
+            ->groupBy('a.gender')
+            ->getQuery()
+            ->getArrayResult()
+        ;
+    }
+
+    public function countByGenderManagedBy(Adherent $referent): array
+    {
+        if (!$referent->isReferent()) {
+            return [];
```
B-Galati: Perhaps it would be better to throw an exception as it looks like a dev mistake to call this method with non referent.

src/Repository/AdherentRepository.php
```diff
@@ -499,6 +499,28 @@ public function countByGender(): array
...
+        }
+
+        return $this->createQueryBuilder('a', 'a.gender')
+            ->select('a.gender, COUNT(a.id) AS count')
```
B-Galati: Nothing important, but `COUNT(a)` should be enough instead of `COUNT(a.id)` no ?

src/Controller/Api/AdherentsController.php
```diff
@@ -39,6 +39,29 @@ public function usersAction(AdherentRepository $adherentRepository): Response
...
 
+    /**
+     * @Route("/count_for_referent_managed_area", name="app_adherents_count_for_referent_managed_area")
+     * @Method("GET")
+     */
+    public function adherentsCountForReferentManagedAreaAction(AdherentRepository $adherentRepository): Response
+    {
+        $count = $adherentRepository->countByGenderManagedBy($this->getUser());
+
+        array_walk($count, function (&$item) {
```
B-Galati: The repository could simply answer with a value object.
It would isolate the casting/sum logic into it.

i.e : 

```
$count = new GenderCount($male, $female);
$count->getTotal();
$count->getMaleCount();
$count->getFemaleCount();
$count->getMalePercentage();
$count->getFemalePercentage();
```

Regarding the percentage, it would be more flexible to let the front handle this calculation. They could change the rounding or use raw data without asking us.

illusionOfParadise: This is done in a generic way, if one day we add another gender we will have no need to modify code.
I agree with front calculation of percentage.

B-Galati: Generic but cumbersome.
At least do a private method to avoid copy/pasting.

src/Controller/Api/AdherentsController.php
```diff
@@ -17,30 +17,35 @@
...
-
-        array_walk($count, function (&$item) use ($total) {
-            $item = $this->calculatePercentage($item, $total);
-        });
+        return new JsonResponse($this->aggregateCount($count));
+    }
 
-        $count['total'] = $total;
+    /**
+     * @Route("/count_for_referent_managed_area", name="app_adherents_count_for_referent_managed_area")
```
B-Galati: cs: hyphen case and I do thing the route should be called `count-by-referent-area`

Sorry I missed that on first review.

PR #2372, opened 2018-04-23
[ch3751] Don t allow user to donate multiple monthly donations

cacahouwete:
https://app.clubhouse.io/enmarche/story/3751/emp%C3%AAcher-les-personnes-qui-font-d%C3%A9j%C3%A0-un-don-mensuel-d-en-faire-un-second

### DIFF
--- a/features/donation.feature
+++ b/features/donation.feature
@@ -1,0 +1,198 @@
+Feature: The goal is to donate one time or multiple time with a subscription
+  In order to donate
+  As an anonymous user or connected user
+  I should be able to donate punctually or subscribe foreach month
+
+  Scenario Outline: The user have to be able to go to donation page every where
+    Given the following fixtures are loaded:
+      | LoadHomeBlockData |
+      | LoadArticleData   |
+      | LoadPageData      |
+    And I am on "<url>"
+    And the response status code should be 200
+    Then I should see "Donner"
+
+    When I follow "Donner"
+    Then I should be on "/don"
+
+    Examples:
+      | url           |
+      | /             |
+      | /le-mouvement |
+      | /evenements   |
+      | /comites      |
+      | /campus       |
+      | /articles     |
+
+  @javascript
+  Scenario: An anonymous user can donate successfully
+    Given I am on "/don"
+    And I press "OK"
+    And wait 1 second until I see "Continuer"
+    When I press "Continuer"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=0"
+
+    When I fill in the following:
+      | app_donation_gender | male                     |
+      | Nom                 | Jean                     |
+      | Prénom              | Dupont                   |
+      | Adresse email       | jean.dupont@en-marche.fr |
+      | Code postal         | 75001                    |
+      | Ville               | Paris                    |
+      | Adresse postale     | 1 allée vivaldie         |
+      | Numéro de téléphone | 0123456789               |
+    And I click the "donation_check_label" element
+    And I click the "donation_check_nationality_label" element
+    And I press "Je donne"
+    Then I wait 5 second until I see "Numéro de carte"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYpagepaiement.cgi"
+
+    When I fill in the following:
+      | NUMERO_CARTE | 1111 2222 3333 4444 |
+      | CVVX         | 123                 |
+    And I select "12" from "MOIS_VALIDITE"
+    And I select "18" from "AN_VALIDITE"
+    And I press "VALIDER"
+    Then I wait 5 second until I see "Paiement accepté"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYtraitetransaction.cgi"
+
+    When I click on the "1" "img" element
+    Then I should see "Votre soutien compte beaucoup pour moi."
+
+  @javascript
+  Scenario: The user can subscribe to donate each month successfully but can't have a second subscription
+    Given I am on "/don"
+    And I press "OK"
+    And wait 1 second until I see "Continuer"
+    When I click the "donation-monthly_label" element
+    And I press "Continuer"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
+
+    When I fill in the following:
+      | app_donation_gender | male                     |
+      | Nom                 | Jean                     |
+      | Prénom              | Dupont                   |
+      | Adresse email       | jean.dupont@en-marche.fr |
+      | Code postal         | 75001                    |
+      | Ville               | Paris                    |
+      | Adresse postale     | 1 allée vivaldie         |
+      | Numéro de téléphone | 0123456789               |
+    And I click the "donation_check_label" element
+    And I click the "donation_check_nationality_label" element
+    And I press "Je donne"
+    Then I wait 5 second until I see "Numéro de carte"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYpagepaiement.cgi"
+
+    When I fill in the following:
+      | NUMERO_CARTE | 1111 2222 3333 4444 |
+      | CVVX         | 123                 |
+    And I select "12" from "MOIS_VALIDITE"
+    And I select "18" from "AN_VALIDITE"
+    And I press "VALIDER"
+    Then I wait 5 second until I see "Paiement accepté"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYtraitetransaction.cgi"
+
+    When I click on the "1" "img" element
+    Then I should see "Votre soutien compte beaucoup pour moi."
+
+    Given I am on "/don"
+    And wait 1 second until I see "Continuer"
+    And I click the "donation-monthly_label" element
+    When I press "Continuer"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
+
+    When I fill in the following:
+      | app_donation_gender | male                     |
+      | Nom                 | Jean                     |
+      | Prénom              | Dupont                   |
+      | Adresse email       | jean.dupont@en-marche.fr |
+      | Code postal         | 75001                    |
+      | Ville               | Paris                    |
+      | Adresse postale     | 1 allée vivaldie         |
+      | Numéro de téléphone | 0123456789               |
+    And I click the "donation_check_label" element
+    And I click the "donation_check_nationality_label" element
+    And I press "Je donne"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
+    And I should see "Vous faites déjà un don mensuel à La République En Marche ! Vous pouvez vous rendre sur votre profil pour en modifier le montant ou faire un nouveau don unique."
+
+    When I follow "faire un nouveau don unique"
+    Then I should be on "/don/coordonnees?montant=50"
+    And the "app_donation_gender" field should contain "male"
+    And the "Nom" field should contain "Jean"
+    And the "Prénom" field should contain "Dupont"
+    And the "Adresse email" field should contain "jean.dupont@en-marche.fr"
+
+  @javascript
+  Scenario: The logged user can subscribe to donate each month successfully but can't have a second subscription without unsubscribe before
+    Given the following fixtures are loaded:
+      | LoadAdherentData |
+    And I am logged as "jacques.picard@en-marche.fr"
+    And I am on "/don"
+    And I press "OK"
+    And wait 1 second until I see "Continuer"
+    When I click the "donation-monthly_label" element
+    And I press "Continuer"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
+
+    When I click the "donation_check_label" element
+    And I click the "donation_check_nationality_label" element
+    And I press "Je donne"
+    Then I wait 5 second until I see "Numéro de carte"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYpagepaiement.cgi"
+
+    When I fill in the following:
+      | NUMERO_CARTE | 1111 2222 3333 4444 |
+      | CVVX         | 123                 |
+    And I select "12" from "MOIS_VALIDITE"
+    And I select "18" from "AN_VALIDITE"
+    And I press "VALIDER"
+    Then I wait 5 second until I see "Paiement accepté"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYtraitetransaction.cgi"
+
+    When I click on the "1" "img" element
+    Then I should see "Votre soutien compte beaucoup pour moi."
+
+    Given I am on "/don"
+    And wait 1 second until I see "Continuer"
+    And I click the "donation-monthly_label" element
+    When I press "Continuer"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
+
+    When I click the "donation_check_label" element
+    And I click the "donation_check_nationality_label" element
+    And I press "Je donne"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
+    And I should see "Vous faites déjà un don mensuel à La République En Marche ! Vous pouvez vous rendre sur votre profil pour en modifier le montant ou faire un nouveau don unique."
+
+    When I follow "vous rendre sur votre profil"
+    Then I should be on "/parametres/mon-compte"
+
+    When I follow "Mes dons"
+    And I follow "Arrêter mon don mensuel"
+    And I press "Oui"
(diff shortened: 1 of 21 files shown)


behat.yml.dist
```diff
@@ -6,6 +6,8 @@ default:
                 - Behatch\Context\BrowserContext:
                     timeout: 3
                 - Behatch\Context\XmlContext
+                - Behatch\Context\DebugContext:
+                    screenshotDir: "var/cache/selenium/"
```
B-Galati: It's more `var/logs` than `var/cache` no ?
Not sure it should be enabled all the time. I would prefer to activate it on demand per developer thanks to `behat.yml` file instead of having it all the time. What do you think ?

features/donation.feature
```diff
@@ -0,0 +1,150 @@
...
+  Scenario: The logged user can subscribe to donate each month successfully but can't have a second subscription
+    Given the following fixtures are loaded:
+      | LoadAdherentData |
+    And I am logged as "jacques.picard@en-marche.fr"
+    And I am on "/don"
+    And I press "OK"
+    And wait 1 second until I see "Continuer"
+    When I click the "donation-monthly_label" element
+    And I press "Continuer"
+    Then I should be on "/don/coordonnees?montant=50&abonnement=1"
```
B-Galati: Please add a line break after each `Given` / `When` /  `Then` block.

cacahouwete: done

src/Controller/EnMarche/DonationController.php
```diff
@@ -53,20 +53,26 @@ public function informationsAction(Request $request)
...
+            (float) $amount,
+            $subscription,
+            $this->getUser()
+        );
 
         $form = $this->createForm(DonationRequestType::class, $donationRequest, ['locale' => $request->getLocale()]);
 
         if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
             $this->get('app.donation_request.handler')->handle($donationRequest);
+            $donationRequestUtils->resetDonationRequest();
```
B-Galati: From this context it would make more sense to call the method `terminateDonationRequest()`

src/Controller/EnMarche/DonationController.php
```diff
@@ -53,20 +53,26 @@ public function informationsAction(Request $request)
...
 
             return $this->redirectToRoute('donation_pay', [
                 'uuid' => $donationRequest->getUuid()->toString(),
             ]);
         }
 
+        $donationRequestUtils->saveDonationRequest($donationRequest);
```
B-Galati: Same but with `startDonationRequest()`.
What do you think ?

src/Donation/DonationRequest.php
```diff
@@ -10,11 +10,23 @@
 use Misd\PhoneNumberBundle\Validator\Constraints\PhoneNumber as AssertPhoneNumber;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Component\Validator\Constraints as Assert;
 
+/**
+ * TODO: add endSubscriptionAt in the list when end subscription will be merged.
```
B-Galati: Do we have a card for that ?
If yes, it would be better to remove this TODO and add the information in the card IMO.

src/Donation/DonationRequestUtils.php
```diff
@@ -56,6 +59,10 @@ public function createFromRequest(Request $request, float $amount, int $duration
     {
         $clientIp = $request->getClientIp();
 
+        if ($donation = $this->session->get('donation_request')) {
```
B-Galati: It can be moved up to the top as $clientIp is useless if it goes in the condition.

src/Form/TypeExtension/FormTypeExtension.php
```diff
@@ -30,9 +31,11 @@ public function configureOptions(OptionsResolver $resolver)
             ->setDefaults([
                 'cast_null_to_string' => false,
                 'form_full' => false,
+                'error_raw' => false,
```
B-Galati: I do think it's really and has nothing to do in form types. This is purely twig logic. IMO it should go directly in twig template like this for example :

```twig
{{ form_errors(form, {output_raw: true}) }}
```

Perhaps there is an even better solution. What do you think @illusionOfParadise @Remg @ottaviano @thomasbisignani?

src/Repository/DonationRepository.php
```diff
@@ -26,4 +33,13 @@ public function findByEmailAddressOrderedByDonatedAt(string $email, string $orde
             ['donatedAt' => $order]
         );
     }
+
+    public function findSubscriptions(array $criteria): array
+    {
+        if (PayboxPaymentSubscription::NONE === $criteria['duration']) {
```
B-Galati: I don't really understand this wrapper around `findBy` method. Could you explain why please?

I think this logic belongs to the validator and that's all.

src/Donation/DonationRequest.php
ottaviano: I think it'll be better to add `status` property and ckeck it value instead of check if `subscription_ended_at is not null`

src/Donation/DonationRequest.php
```diff
@@ -10,11 +10,23 @@
...
+ *
+ * @UniqueEntity(
+ *     service="AppBundle\Validator\UniqueDonationSubscriptionValidator",
+ *     message="donation.subscription.not_unique",
+ *     fields={
+ *     "emailAddress",
+ *     "duration"
```
ottaviano: why duration? I think you need juste email to validate this donation request

src/Donation/DonationRequest.php
cacahouwete: Yes there are an other card to unsubscribe, the idea is to rebase with master when it will be merged^^,

src/Form/TypeExtension/FormTypeExtension.php
ottaviano: I like this solution, keep this logic only in templates templates

src/Donation/DonationRequest.php
cacahouwete: Because we need want unique subscription

src/Donation/DonationRequest.php
cacahouwete: Good idea, but I didn't want to do a bigger PR and maybe have an other PR for this refactor ?

src/Form/TypeExtension/FormTypeExtension.php
cacahouwete: In same logic, all translation_domain, attr, label, placeholder, disable_fields, with_character_count, choice_attr, ... have to be removed ?

src/Donation/DonationRequest.php
cacahouwete: It could be done in this card maybe ?
https://app.clubhouse.io/enmarche/story/3967/etq-enmarche-je-souhaite-avoir-l-historique-de-toutes-les-transactions-paybox

src/Form/TypeExtension/FormTypeExtension.php
B-Galati: It is recommended to manage them in twig directly, like button etc. You can check Symfony best practices.

src/Donation/DonationRequest.php
B-Galati: Just rebase one of the 2 PR on the other one if they are dependent.

cacahouwete: I prefer to wait, because I think it's dangerous to be based on an other PR :s

src/Form/TypeExtension/FormTypeExtension.php
cacahouwete: I moved set value in twig, but we need the declaration in formtypeextension

B-Galati: No we don't, just test if the variable is set or not. https://twig.symfony.com/doc/2.x/tests/defined.html

cacahouwete: then we should have the same thing for attr label ... then ? Why it's not like that ?

src/Donation/DonationRequest.php
cacahouwete: some commits will be added and after that it will have a squash, if it's squash , I will have to remove all commits from original rebase and after that rebase on master. And the other thing is I will refactor Donation entity to have all transactions, and add a status with some different values like: confirmation_pending, finished, subscription_in_progress, error.

src/Form/TypeExtension/FormTypeExtension.php
B-Galati: I guess it's because it is considered a real option even if it's entirely related to the view. It should be ok when you want to develop in RAD mode.

Anyway the framework is trying to solve the problem of 80% of ppl. Here we just want to solve our problem :D

src/Donation/DonationRequest.php
B-Galati: Ok so it's just a matter of organization that you and only you is responsible for. It's not really dangerous. You just need to remember which commits belongs to which branch. Easy to do with a correct naming.

src/Form/TypeExtension/FormTypeExtension.php
HeahDude: A form type is also responsible for configuring the form view, I d'ont see no harm to let an option set it, especially when it helps keeping templating simple.

B-Galati: Do as you wish @cacahouwete :+1: not a big deal.

src/Validator/UniqueDonationSubscriptionValidator.php
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * @param DonationRequest $value
+     * @param UniqueEntity    $constraint
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        if (!$constraint instanceof UniqueEntity) {
+            throw new UnexpectedTypeException($constraint, UniqueCommittee::class);
+        }
+
```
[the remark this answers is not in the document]
cacahouwete: done

src/Donation/DonationRequest.php
cacahouwete: Yes, I said dangerous because I could have a miss during manipulation :'( , but yes it's only organization ^^,

ottaviano: yes it was a suggestion ;)

src/Donation/DonationRequest.php
ottaviano: yes, right :)

features/donation.feature
```diff
@@ -0,0 +1,198 @@
...
+    And I press "Je donne"
+    Then I wait 5 second until I see "Numéro de carte"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYpagepaiement.cgi"
+
+    When I fill in the following:
+      | NUMERO_CARTE | 1111 2222 3333 4444 |
+      | CVVX         | 123                 |
+    And I select "12" from "MOIS_VALIDITE"
+    And I select "18" from "AN_VALIDITE"
+    And I press "VALIDER"
```
thomasbisignani: `Valider`

features/donation.feature
```diff
@@ -0,0 +1,198 @@
...
+    Then I wait 5 second until I see "PAIEMENT ACCEPTé"
```
thomasbisignani: `Paiement accepté"

features/donation.feature
```diff
@@ -0,0 +1,198 @@
...
```
thomasbisignani: `Paiement accepté"

features/donation.feature
```diff
@@ -0,0 +1,198 @@
...
+    And I press "Je donne"
+    Then I wait 5 second until I see "Numéro de carte"
+    And I should be on "https://preprod-tpeweb.paybox.com/cgi/MYpagepaiement.cgi"
+
+    When I fill in the following:
+      | NUMERO_CARTE | 1111 2222 3333 4444 |
+      | CVVX         | 123                 |
+    And I select "12" from "MOIS_VALIDITE"
+    And I select "18" from "AN_VALIDITE"
+    And I press "VALIDER"
```
thomasbisignani: `Valider`

features/donation.feature
```diff
@@ -0,0 +1,198 @@
...
+    Then I wait 5 second until I see "PAIEMENT ACCEPTé"
```
thomasbisignani: `Paiement accepté`

features/donation.feature
```diff
@@ -0,0 +1,198 @@
...
```
thomasbisignani: `Paiement accepté`

features/donation.feature
cacahouwete: Not possible , it's the name of the button because the label is an image

features/donation.feature
cacahouwete: Not possible , it's the name of the button because the label is an image

src/Validator/UniqueDonationSubscriptionValidator.php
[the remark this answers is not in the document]
ottaviano: the if `if (!$value instanceof DonationRequest)` already handles this case :)

src/Donation/DonationRequest.php
```diff
@@ -327,4 +337,9 @@ private static function createPhoneNumber(int $countryCode = 33, string $number
 
         return $phone;
     }
+
+    public function hasSubscription(): bool
```
ottaviano: `isSubscription` no?

src/Donation/DonationRequestUtils.php
```diff
@@ -43,6 +44,7 @@ class DonationRequestUtils
         // Other
         self::PAYBOX_UNKNOWN => 'error',
     ];
+    private const SESSION_DONATION_REQUEST = 'donation_request';
```
ottaviano: what about just `SESSION_KEY`?

src/Validator/UniqueDonationSubscriptionValidator.php
```diff
@@ -0,0 +1,63 @@
...
+        if (null === $value) {
+            return;
+        }
+
+        if (!$value instanceof DonationRequest) {
+            throw new UnexpectedTypeException($value, DonationRequest::class);
+        }
+
+        if (PayboxPaymentSubscription::NONE === $value->getDuration()) {
```
ottaviano: why you check the same things twice ? 
`if (PayboxPaymentSubscription::NONE === $value->getDuration()` L43
`if ($value->hasSubscription()` L47

src/Validator/UniqueDonationSubscriptionValidator.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * @param DonationRequest $value
+     * @param UniqueEntity    $constraint
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        if (!$constraint instanceof UniqueEntity) {
+            throw new UnexpectedTypeException($constraint, UniqueCommittee::class);
```
HeahDude: `UniqueCommittee`??

src/Donation/DonationRequest.php
```diff
@@ -10,11 +10,21 @@
 use Misd\PhoneNumberBundle\Validator\Constraints\PhoneNumber as AssertPhoneNumber;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Component\Validator\Constraints as Assert;
 
+/**
+ * @UniqueEntity(
+ *     service="AppBundle\Validator\UniqueDonationSubscriptionValidator",
+ *     message="donation.subscription.not_unique",
```
HeahDude: You should create a constraint to set the message and use it as:
```php
use App\Validator\Constraints\UniqueSubscribedDonation as AssertUniqueSubscribedDonation;

/**
 * @AssertUniqueSubscribedDonation
 */
```

cacahouwete: That's what I did, but @ottaviano  told me to use UniqueEntity. I didn't see any objections. I can recreate it if he is ok too ^^,

src/Validator/UniqueDonationSubscriptionValidator.php
cacahouwete: Ooooo, my bad :s , I will wait @ottaviano response for the next com before correct it. But thx ;)

src/Donation/DonationRequest.php
HeahDude: I've an objection, this is not an entity, and you're not using its validator :).

PR #2373, opened 2018-04-24
Small fix on the donation tab page

marineb:


templates/user/my_donation.html.twig
```diff
@@ -24,28 +24,30 @@
...
+            <tbody>
+                {% for donation in donations|reverse %}
+                    <tr>
+                        <td>
+                            {{ donation.donatedAt|date('d M Y') }}
+                        </td>
+                        <td>
+                            {{ donation.amountInEuros }} €
+                        </td>
+                    </tr>
```
ottaviano: you can do this with `{% else %}` and add one message:
```
{% for donation in donations|reverse %}
...
{% else %}
    <tr><td colspan="2">Vous n'avez encore fait de don</td></tr>
{% endfor %}
```

marineb: Good point. Thank you! Will change.

marineb: Actually, that's not the only area that needs to disappear, I made another tweak instead.

PR #2374, opened 2018-04-24
Create user email subscription history

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/3660/stocker-l-historique-des-choix-de-notification-des-utilisateurs

### DIFF
--- a/app/config/services/services.xml
+++ b/app/config/services/services.xml
@@ -98,6 +98,7 @@
             </argument>
         </service>
         <service id="AppBundle\Form\CommitteeAroundAdherentType"/>
+        <service id="AppBundle\Form\AdherentEmailSubscriptionType"/>
 
         <!-- GeoCoder -->
         <service id="GeoIp2\WebService\Client">
@@ -149,6 +150,7 @@
         <!--Command-->
         <service id="AppBundle\Command\ApiScheduleCommitteeCreationCommand" />
         <service id="AppBundle\Command\ApiScheduleEventCreationCommand" />
+        <service id="AppBundle\Command\InitializeEmailSubscriptionHistoryCommand"/>
 
         <!-- UserDocument -->
         <service id="AppBundle\UserDocument\UserDocumentManager" public="true" />
@@ -166,5 +168,8 @@
 
         <!-- Validators -->
         <service id="AppBundle\Validator\DateRangeValidator" />
+
+        <!-- History -->
+        <service id="AppBundle\History\EmailSubscriptionHistoryHandler" />
     </services>
 </container>
--- a/app/migrations/Version20180425145459.php
+++ b/app/migrations/Version20180425145459.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20180425145459 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('
+            CREATE TABLE adherent_email_subscription_histories (
+                uuid CHAR(36) NOT NULL COMMENT \'(DC2Type:uuid)\', 
+                adherent_uuid CHAR(36) NOT NULL COMMENT \'(DC2Type:uuid)\', 
+                referent_tag_id INT UNSIGNED DEFAULT NULL, 
+                subscribed_email_type VARCHAR(50) NOT NULL,
+                action VARCHAR(32) NOT NULL, 
+                date DATETIME NOT NULL COMMENT \'(DC2Type:datetime_immutable)\', 
+                INDEX adherent_email_subscription_histories_adherent_uuid_idx (adherent_uuid),
+                INDEX adherent_email_subscription_histories_adherent_action_idx (action),
+                INDEX adherent_email_subscription_histories_adherent_date_idx (date),
+                INDEX adherent_email_subscription_histories_adherent_email_type_idx (subscribed_email_type), 
+                INDEX IDX_51AD83549C262DB3 (referent_tag_id),
+                PRIMARY KEY(uuid)
+            ) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB
+        ');
+        $this->addSql('ALTER TABLE adherent_email_subscription_histories ADD CONSTRAINT FK_272CB3E99C262DB3 FOREIGN KEY (referent_tag_id) REFERENCES referent_tags (id)');
+    }
+
+    public function down(Schema $schema)
+    {
+        $this->addSql('DROP TABLE adherent_email_subscription_histories');
+    }
+}
--- a/src/Admin/AdherentAdmin.php
+++ b/src/Admin/AdherentAdmin.php
@@ -3,6 +3,7 @@
 namespace AppBundle\Admin;
 
 use AppBundle\Entity\Adherent;
+use AppBundle\History\EmailSubscriptionHistoryHandler;
 use AppBundle\Entity\BoardMember\BoardMember;
 use AppBundle\Entity\BoardMember\Role;
 use AppBundle\Entity\CommitteeMembership;
@@ -44,12 +45,24 @@ class AdherentAdmin extends AbstractAdmin
     ];
 
     private $dispatcher;
+    private $emailSubscriptionHistoryManager;
 
-    public function __construct($code, $class, $baseControllerName, EventDispatcherInterface $dispatcher)
-    {
+    /**
+     * @var array Useful on update to know state before update
+     */
+    private $oldEmailsSubscriptions;
+
+    public function __construct(
+        $code,
+        $class,
+        $baseControllerName,
+        EventDispatcherInterface $dispatcher,
+        EmailSubscriptionHistoryHandler $emailSubscriptionHistoryManager
+    ) {
         parent::__construct($code, $class, $baseControllerName);
 
         $this->dispatcher = $dispatcher;
+        $this->emailSubscriptionHistoryManager = $emailSubscriptionHistoryManager;
     }
 
     public function getTemplate($name)
@@ -416,10 +429,24 @@ class AdherentAdmin extends AbstractAdmin
     }
 
     /**
+     * @param Adherent $subject
+     */
+    public function setSubject($subject)
+    {
+        if (null === $this->oldEmailsSubscriptions) {
+            $this->oldEmailsSubscriptions = $subject->getEmailsSubscriptions();
+        }
+        parent::setSubject($subject);
+    }
+
+    /**
      * @param Adherent $object
      */
     public function postUpdate($object)
     {
+        // No need to handle referent tags update as they are not update-able from admin
+        $this->emailSubscriptionHistoryManager->handleSubscriptionsUpdate($object, $this->oldEmailsSubscriptions);
+
         $this->dispatcher->dispatch(UserEvents::USER_UPDATED, new UserEvent($object));
     }
 
--- a/src/Command/InitializeEmailSubscriptionHistoryCommand.php
+++ b/src/Command/InitializeEmailSubscriptionHistoryCommand.php
@@ -1,0 +1,141 @@
+<?php
+
+namespace AppBundle\Command;
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\Reporting\EmailSubscriptionHistory;
+use AppBundle\Entity\Reporting\EmailSubscriptionHistoryAction;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Doctrine\ORM\QueryBuilder;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class InitializeEmailSubscriptionHistoryCommand extends Command
+{
+    private const BATCH_SIZE = 50;
+
+    protected static $defaultName = 'app:adherent:initialize-email-subscriptions-history';
+
+    /**
+     * @var SymfonyStyle
+     */
+    private $io;
+    private $em;
+
+    public function __construct(EntityManagerInterface $em)
+    {
+        $this->em = $em;
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Create email subscriptions history. The history will be created for both users and adherents even inactives.')
+        ;
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $this->io = new SymfonyStyle($input, $output);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if ($this->isAlreadyInitialize()) {
+            $this->io->error('Cannot initialize email subscription history. It already exists.');
+
+            return 1;
+        }
+
+        $this->io->title('Starting email subscription history initialization.');
+
+        $progressBar = new ProgressBar($output, $this->getAdherentCount());
+
+        $this->em->beginTransaction();
+
+        $nb = 0;
+        foreach ($this->getAdherents() as $result) {
+            $adherent = $result[0];
+
+            /** @var Adherent $adherent */
+            foreach ($adherent->getReferentTags() as $referentTag) {
+                foreach ($adherent->getEmailsSubscriptions() as $subscription) {
+                    $subscriptionHistory = new EmailSubscriptionHistory(
+                        $adherent,
+                        $subscription,
+                        $referentTag,
+                        EmailSubscriptionHistoryAction::SUBSCRIBE(),
+                        $adherent->getActivatedAt() ?: $adherent->getRegisteredAt()
+                    );
+                    $this->em->persist($subscriptionHistory);
+                }
+            }
+
+            $progressBar->advance();
+
+            ++$nb;
+
+            if (0 === ($nb % self::BATCH_SIZE)) {
+                $this->em->flush();
+                $this->em->clear();
+                $nb = 0;
+            }
+        }
+
+        $this->em->flush();
+        $this->em->commit();
+
+        $progressBar->finish();
+
+        $this->io->newLine(2);
+        $this->io->success('Email subscription history initialized successfully!');
+    }
+
+    private function isAlreadyInitialize(): bool
+    {
+        $nbHistories = $this
+            ->em
+            ->getRepository(EmailSubscriptionHistory::class)
+            ->createQueryBuilder('history')
+            ->select('COUNT(history)')
+            ->getQuery()
+            ->getSingleScalarResult()
+        ;
+
+        return $nbHistories > 0;
+    }
+
+    private function getAdherents(): IterableResult
+    {
+        return $this
+            ->createQueryBuilder()
+            ->getQuery()
+            ->iterate()
+        ;
+    }
+
+    private function getAdherentCount(): int
+    {
+        return $this
+            ->createQueryBuilder()
+            ->select('COUNT(adherent)')
+            ->getQuery()
+            ->getSingleScalarResult()
+        ;
+    }
+
+    private function createQueryBuilder(): QueryBuilder
+    {
+        return $this
+            ->em
+            ->getRepository(Adherent::class)
+            ->createQueryBuilder('adherent')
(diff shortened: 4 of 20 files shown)


app/config/services/repositories.xml
```diff
@@ -202,5 +202,10 @@
             <factory service="doctrine" method="getRepository" />
             <argument>AppBundle\Entity\ReferentTag</argument>
         </service>
+
+        <service id="AppBundle\Repository\AdherentEmailSubscriptionHistoryRepository">
```
B-Galati: You can now use `ServiceEntityRepository` instead now

app/config/services/services.xml
```diff
@@ -127,5 +128,8 @@
 
         <!-- UserDocument -->
         <service id="AppBundle\UserDocument\UserDocumentManager" public="true" />
+
+        <!-- History -->
+        <service id="AppBundle\History\AdherentEmailSubscriptionHistoryManager" public="true" />
```
B-Galati: It looks like public property is useless no ?

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        $this->em->flush();
+        $this->em->commit();
+
+        $progressBar->finish();
+
+        $this->io->newLine(2);
+        $this->io->text('Email subscription history initialized successfully!');
```
B-Galati: As an alternative `success` method is amazing :star2:

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,133 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class InitializeAdherentEmailSubscriptionHistoryCommand extends ContainerAwareCommand
+{
+    private const COMMAND_NAME_ALIAS = 'app:adherent:iesh';
+    private const BATCH_SIZE = 50;
+
+    protected static $defaultName = 'app:adherent:initialize_email_subscriptions_history';
```
B-Galati: cs: hyphen instead of underscore please.

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->em = $this->getContainer()->get('doctrine.orm.entity_manager');
+        $this->io = new SymfonyStyle($input, $output);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if ($this->checkHistoryExists()) {
+            $this->io->error('Impossible to initialize email subscription history. It already exists.');
+
+            return;
```
B-Galati: Return something different than 0 if it's an error.

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        $this->io->title('Starting email subscription history initialization.');
+
+        $progressBar = new ProgressBar($output, $this->getCount());
+
+        $this->em->beginTransaction();
+
+        $nb = 0;
+        foreach ($this->getIterator() as $result) {
```
B-Galati: Let's call it `getAdherents`. iterator is a detail.

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,133 @@
...
+    private function getIterator(): IterableResult
+    {
+        return $this
+            ->getQueryBuilder()
+            ->getQuery()
+            ->iterate()
+        ;
+    }
+
+    private function getCount(): int
```
B-Galati: `getAdherentCount` is a more explicit name IMO.

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @ORM\Column(type="bigint")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Adherent")
+     * @ORM\JoinColumn(onDelete="CASCADE")
```
B-Galati: We should keep all information we have at a given time so it should be set adherent to `null` when one adherent is deleted.

B-Galati: And IMO it should be explicit so we should not rely on that cascade stuff. The goal is to avoid silly mistakes.

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    private $adherent;
+
+    /**
+     * @ORM\Column(length=50)
+     * @Assert\Choice(callback={"AppBundle\Membership\AdherentEmailSubscription", "getMergedSubscriptions"}, strict=true)
```
B-Galati: I don't think we need this Constraint no ?

tests/Controller/EnMarche/AdherentControllerTest.php
```diff
@@ -150,94 +151,101 @@ public function testEditAdherentProfile(): void
...
 
-        $errors = $crawler->filter('.form__errors > li');
+        $crawler = $this->client->request(Request::METHOD_GET, '/parametres/mon-compte/modifier');
 
-        $this->assertStatusCode(Response::HTTP_OK, $this->client);
-        $this->assertSame(3, $errors->count());
-        $this->assertSame('Le code postal doit contenir moins de 15 caractères.', $errors->eq(0)->text());
-        $this->assertSame('L\'adresse ne peut pas dépasser 150 caractères.', $errors->eq(1)->text());
-        $this->assertSame('Cette adresse e-mail existe déjà.', $errors->eq(2)->text());
+//        $inputPattern = 'input[name="adherent[%s]"]';
```
B-Galati: What are all these comments ?

src/Admin/AdherentAdmin.php
```diff
@@ -44,12 +45,19 @@ class AdherentAdmin extends AbstractAdmin
...
+        $code,
+        $class,
+        $baseControllerName,
+        EventDispatcherInterface $dispatcher,
+        AdherentEmailSubscriptionHistoryManager $emailSubscriptionHistoryManager
+    ) {
         parent::__construct($code, $class, $baseControllerName);
 
         $this->dispatcher = $dispatcher;
+        $this->emailSubscriptionHistoryManager = $emailSubscriptionHistoryManager;
```
Remg: this property is not declared

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,115 @@
...
+     */
+    private $unsubscribedAt;
+
+    /**
+     * @param Adherent    $adherent
+     * @param string      $subscribedEmailsType
+     * @param ReferentTag $referentTag
+     * @param \DateTime   $subscribedAt
+     * @param \DateTime   $unsubscribedAt
+     */
```
Remg: is this docblock useful ?

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,133 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class InitializeAdherentEmailSubscriptionHistoryCommand extends ContainerAwareCommand
+{
+    private const COMMAND_NAME_ALIAS = 'app:adherent:iesh';
+    private const BATCH_SIZE = 50;
+
+    protected static $defaultName = 'app:adherent:initialize-email-subscriptions-history';
```
Remg: is this property useful ?

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getAdherent(): Adherent
+    {
+        return $this->adherent;
+    }
+
+    public function setAdherent(Adherent $adherent)
```
Remg: missing return type

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
B-Galati: Yes it's one way to enable lazy commands in Symfony

Remg: oh yes it is \o/

B-Galati: https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading

B-Galati:
@illusionOfParadise Please don't squash your commits on big PR like this so that we don't need to review everything every time.

src/Admin/AdherentAdmin.php
illusionOfParadise: Done.

src/Repository/AdherentEmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        parent::__construct($registry, AdherentEmailSubscriptionHistory::class);
+    }
+
+    /**
+     * @param Adherent $adherent
+     *
+     * @return AdherentEmailSubscriptionHistory[]
+     */
+    public function findByAdherent(Adherent $adherent, $withoutInactives = true): array
```
ottaviano: where do you use the second argument (outside tests)? if not used, IMO not need to add this only for test.

src/Repository/AdherentEmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,75 @@
...
+            ->getResult()
+        ;
+    }
+
+    /**
+     * @param Adherent $adherent
+     *
+     * @return AdherentEmailSubscriptionHistory[]
+     */
+    public function findAllByAdherentAndReferentTag(Adherent $adherent, ReferentTag $tag): array
```
ottaviano: where is it used?

src/Repository/AdherentEmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,75 @@
...
+
+        return $qb->getQuery()->getResult();
+    }
+
+    /**
+     * @param Adherent $adherent
+     *
+     * @return AdherentEmailSubscriptionHistory[]
+     */
+    public function findAllByAdherentAndType(Adherent $adherent, string $subscriptionType): array
```
ottaviano: it's only used in test? really need keep this method?

app/config/services/repositories.xml
```diff
@@ -202,5 +202,7 @@
             <factory service="doctrine" method="getRepository" />
             <argument>AppBundle\Entity\ReferentTag</argument>
         </service>
+
+        <service id="AppBundle\Repository\AdherentEmailSubscriptionHistoryRepository" />
```
ottaviano: maybe `EmailSubscriptionHistory(Repository|Type|Command|Manager)` (without `Adherent`) will it be enough?

app/migrations/Version20180425145459.php
```diff
@@ -0,0 +1,33 @@
...
+
+class Version20180425145459 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('
+            CREATE TABLE adherent_email_subscription_history (
+                id BIGINT AUTO_INCREMENT NOT NULL, 
+                adherent_id INT UNSIGNED DEFAULT NULL, 
+                referent_tag_id INT UNSIGNED DEFAULT NULL,
```
ottaviano: why do we need to know the referent tag of adherent on subscription?

src/Admin/AdherentAdmin.php
```diff
@@ -415,11 +424,26 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
     }
 
+    /**
+     * @param Adherent $object
+     */
+    public function preUpdate($object)
+    {
+        $em = $this->getModelManager()->getEntityManager($this->getClass());
+        $original = $em->getUnitOfWork()->getOriginalEntityData($object);
+        $this->emailSubscriptionChanged = count(array_diff($original['emailsSubscriptions'], $object->getEmailsSubscriptions())) > 0
```
ottaviano: 1) what about use `setSubject` method for storing the oldSubject (before update)?
check CRUDController of Sonata Admin, lines 308 and 322, the controller calls `admin::setSubject` twice with old and new object, you can create a new property of admin class `oldSubject` and
```

public function setSubject(subject) {
  if (null === $this->oldSubject) {
    $this->oldSubject = clone $subject;
  }
  parent::setSubject($subject);
}
```

2) you can use the oldSubject property in `postUpdate` (not preUpdate) method
3) you check twice the diff between before and after states of subscriptions: first in this class, and second in `AdherentEmailSubscriptionListener`. I think we can do this better with one listener `AdherentUpdateListener`. It will listen `ADHERENT_UPDATE` event and will make the check of before and after states of adherent subscriptions. you need to dispatch this event.
4) I think to know if subscription is changed, you need to check:
 - count(beforeEmailSubscriptions)
 - count(afterEmailSubscriptions)
 - count(array_intersect(beforeEmailSubscriptions, afterEmailSubscriptions))
if the three count values is different, so it's changed

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $this->em->beginTransaction();
+
+        $nb = 0;
+        foreach ($this->getAdherents() as $result) {
+            $adherent = reset($result);
+
+            foreach ($adherent->getReferentTags() as $referentTag) {
+                foreach ($adherent->getEmailsSubscriptions() as $subscription) {
+                    $subscriptionHistory = new AdherentEmailSubscriptionHistory($adherent, $subscription, $referentTag, $adherent->activatedAt ?? $adherent->registeredAt);
```
ottaviano: CS: use multiline pls

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $this->io->title('Starting email subscription history initialization.');
+
+        $progressBar = new ProgressBar($output, $this->getAdherentCount());
+
+        $this->em->beginTransaction();
+
+        $nb = 0;
+        foreach ($this->getAdherents() as $result) {
+            $adherent = reset($result);
```
ottaviano: why not just read the first element `$result[0]`?

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+namespace AppBundle\Entity;
+
+use Algolia\AlgoliaSearchBundle\Mapping\Annotation as Algolia;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Table(name="adherent_email_subscription_history")
```
ottaviano: it's the default table name

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
...
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+
+    /**
+     * @var Adherent|null
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Adherent")
+     * @ORM\JoinColumn(onDelete="SET NULL")
```
ottaviano: do we really need to store one subscription history without adherent (with NULL value for adherent_id)?

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * @var \DateTime
+     *
+     * @ORM\Column(type="datetime")
+     */
+    private $subscribedAt;
+
+    /**
+     * @var \DateTime
```
ottaviano: `|null`

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->subscribedAt = $subscribedAt;
+        $this->unsubscribedAt = $unsubscribedAt;
+    }
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    public function getAdherent(): ?Adherent
```
ottaviano: without `?`, you can't have null for this property

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        return $this->adherent;
+    }
+
+    public function setAdherent(Adherent $adherent): void
+    {
+        $this->adherent = $adherent;
+    }
+
+    public function getSubscribedEmailsType(): ?string
```
ottaviano: without `?`, you can't have null for this property

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        return $this->subscribedAt;
+    }
+
+    public function getUnsubscribedAt(): ?\DateTime
+    {
+        return $this->unsubscribedAt;
+    }
+
+    public function setUnsubscribedAt(\DateTime $unsubscribedAt): void
```
ottaviano: `?\DateTime`

src/Entity/AdherentEmailSubscriptionHistory.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getAdherent(): ?Adherent
+    {
+        return $this->adherent;
+    }
+
+    public function setAdherent(Adherent $adherent): void
```
ottaviano: why do you add the setter for this property and not for referentTag or subscribedEmailsType? maybe remove it?

app/migrations/Version20180425145459.php
B-Galati: We need that to filter by tag (which refers to area and more)

app/config/services/repositories.xml
B-Galati: Yes ;-)

B-Galati:
I am taking over this PR while @illusionOfParadise is out.

src/Entity/AdherentEmailSubscriptionHistory.php
B-Galati: Yes we can if the user delete his account.

src/Command/InitializeAdherentEmailSubscriptionHistoryCommand.php
B-Galati: I guess it is more accurate to use $result[0] in that case because $result does not have random keys.

src/Entity/AdherentEmailSubscriptionHistory.php
B-Galati: Yes we want to keep them at all time to have consistent stats.

src/Command/InitializeEmailSubscriptionHistoryCommand.php
```diff
@@ -0,0 +1,146 @@
...
+
+    protected static $defaultName = 'app:adherent:initialize-email-subscriptions-history';
+
+    /**
+     * @var SymfonyStyle
+     */
+    private $io;
+
+    /**
+     * @var EntityManager
```
cacahouwete: This docbloc is useless, you already declare it in your construct.

B-Galati:
@illusionOfParadise @ottaviano @Remg @cacahouwete Ready for review

src/Entity/Reporting/EmailSubscriptionHistory.php
```diff
@@ -0,0 +1,119 @@
...
+     * @ORM\Column(length=32)
+     */
+    private $action;
+
+    /**
+     * @var \DateTime
+     *
+     * @ORM\Column(type="datetime")
+     */
+    private $date;
```
cacahouwete: This date should be immutable no ? he will not change in time ? Why using carbon here ?

PR #2383, opened 2018-04-26
Add an endpoint for emails subscriptions graph by month in referent managed area

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/3637/1%C3%A8re-partie-section-adh%C3%A9rents-graph-double-histogramme-emails-subscription

Need to change tests to be not time depending.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -1,11 +1,27 @@
 <?php
 
 use Behat\MinkExtension\Context\RawMinkContext;
-
-require_once __DIR__.'/../../vendor/phpunit/phpunit/src/Framework/Assert/Functions.php';
+use Cake\Chronos\Chronos;
+use Webmozart\Assert\Assert;
 
 class FeatureContext extends RawMinkContext
 {
+    /**
+     * @Given (I )freeze the clock to :dateTime
+     */
+    public function freezeClock(string $dateTime): void
+    {
+        Chronos::setTestNow($dateTime);
+    }
+
+    /**
+     * @AfterScenario
+     */
+    public function defreezeClock(): void
+    {
+        Chronos::setTestNow();
+    }
+
     /**
      * @Given I resolved the captcha
      */
@@ -29,7 +45,7 @@ class FeatureContext extends RawMinkContext
     {
         $field = $this->getSession()->getPage()->findById($elementId);
 
-        assertNotNull($field, 'Cannot find "'.$elementId.'"');
+        Assert::notNull($field, 'Cannot find "'.$elementId.'"');
 
         $field->click();
     }
--- a/src/Entity/Reporting/EmailSubscriptionHistory.php
+++ b/src/Entity/Reporting/EmailSubscriptionHistory.php
@@ -20,7 +20,7 @@ use Ramsey\Uuid\UuidInterface;
  *          @ORM\Index(name="adherent_email_subscription_histories_adherent_email_type_idx", columns="subscribed_email_type"),
  *      }
  * )
- * @ORM\Entity
+ * @ORM\Entity(repositoryClass="AppBundle\Repository\EmailSubscriptionHistoryRepository")
  *
  * @Algolia\Index(autoIndex=false)
  */
--- a/src/History/EmailSubscriptionHistoryHandler.php
+++ b/src/History/EmailSubscriptionHistoryHandler.php
@@ -5,15 +5,20 @@ namespace AppBundle\History;
 use AppBundle\Entity\Adherent;
 use AppBundle\Entity\Reporting\EmailSubscriptionHistory;
 use AppBundle\Entity\Reporting\EmailSubscriptionHistoryAction;
+use AppBundle\Membership\AdherentEmailSubscription;
+use AppBundle\Repository\EmailSubscriptionHistoryRepository;
+use Cake\Chronos\Chronos;
 use Doctrine\ORM\EntityManagerInterface;
 
 class EmailSubscriptionHistoryHandler
 {
     private $em;
+    private $historyRepository;
 
-    public function __construct(EntityManagerInterface $em)
+    public function __construct(EntityManagerInterface $em, EmailSubscriptionHistoryRepository $historyRepository)
     {
         $this->em = $em;
+        $this->historyRepository = $historyRepository;
     }
 
     /**
@@ -87,5 +92,25 @@ class EmailSubscriptionHistoryHandler
                 );
             }
         }
+    }
+
+    public function queryCountByMonth(Adherent $referent, int $months = 6): array
+    {
+        foreach (range(0, $months - 1) as $month) {
+            $until = (new Chronos("last day of -$month month"))->setTime(23, 59, 59, 999);
+
+            $subscriptions = $this->historyRepository->countAllByTypeForReferentManagedArea(
+                $referent,
+                [
+                    AdherentEmailSubscription::SUBSCRIBED_EMAILS_LOCAL_HOST,
+                    AdherentEmailSubscription::SUBSCRIBED_EMAILS_REFERENTS,
+                ],
+                $until
+            );
+
+            $countByMonth[$until->format('Y-m')] = $subscriptions;
+        }
+
+        return $countByMonth;
     }
 }
--- a/src/Repository/EmailSubscriptionHistoryRepository.php
+++ b/src/Repository/EmailSubscriptionHistoryRepository.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace AppBundle\Repository;
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\Reporting\EmailSubscriptionHistory;
+use AppBundle\Entity\Reporting\EmailSubscriptionHistoryAction;
+use Cake\Chronos\Chronos;
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Common\Persistence\ManagerRegistry;
+
+class EmailSubscriptionHistoryRepository extends ServiceEntityRepository
+{
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, EmailSubscriptionHistory::class);
+    }
+
+    /**
+     * Count active adherents (not users) history for each specified subscription types in the referent managed area before the specified date.
+     */
+    public function countAllByTypeForReferentManagedArea(Adherent $referent, array $subscriptionsTypes, \DateTimeInterface $until): array
+    {
+        if (!$referent->isReferent()) {
+            throw new \InvalidArgumentException('Adherent must be a referent.');
+        }
+
+        $qb = $this->createQueryBuilder('history')
+            ->select('history.adherentUuid, history.subscribedEmailType, history.action, COUNT(history) AS count')
+            ->where('history.referentTag IN (:tags)')
+            ->andWhere('history.subscribedEmailType IN (:subscriptions)')
+            ->andWhere('history.date <= :until')
+            ->groupBy('history.adherentUuid, history.subscribedEmailType, history.action')
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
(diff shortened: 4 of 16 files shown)


src/Repository/AdherentEmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        parent::__construct($registry, AdherentEmailSubscriptionHistory::class);
+    }
+
+    /**
+     * @param Adherent $adherent
+     *
+     * @return AdherentEmailSubscriptionHistory[]
+     */
+    public function findByAdherent(Adherent $adherent, $withoutInactives = true): array
```
cacahouwete: Why no explicit type hint for `$withoutInactives` ?

src/Repository/AdherentEmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Count active adherents (not users) history for each specified subscription types in the referent managed area before the specified date.
+     *
+     * @param Adherent  $referent
+     * @param array     $subscriptionsTypes
+     * @param \DateTime $from
+     *
+     * @return array
+     */
+    public function countAllByTypeForReferentManagedArea(Adherent $referent, array $subscriptionsTypes, \DateTime $beforeDate, $cache = false): array
```
cacahouwete: same for `$cache`

src/Entity/Reporting/EmailSubscriptionHistory.php
```diff
@@ -26,19 +35,21 @@ class EmailSubscriptionHistory
     private $uuid;
 
     /**
-     * @var Adherent|null
+     * Only UUID is used instead of the Entity because we cannot lose this data if one unsubscribes.
+     * Otherwise stats would be fucked up.
```
illusionOfParadise: Could you please change this phrase.

app/migrations/Version20180425145459.php
```diff
@@ -10,24 +10,26 @@ class Version20180425145459 extends AbstractMigration
     public function up(Schema $schema)
     {
         $this->addSql('
-            CREATE TABLE adherent_email_subscription_history (
```
illusionOfParadise: This PR is based on this one https://github.com/EnMarche/en-marche.fr/pull/2374/files#diff-646623fd47639d7f596336914941814cR8. 
It seems to be more correct to change migration in the base PR, because if it will be merged first the migration will be executed and these modifications will not be taken into account.

features/bootstrap/FeatureContext.php
```diff
@@ -1,11 +1,27 @@
...
 use Behat\MinkExtension\Context\RawMinkContext;
-
-require_once __DIR__.'/../../vendor/phpunit/phpunit/src/Framework/Assert/Functions.php';
+use Carbon\Carbon;
+use Webmozart\Assert\Assert;
 
 class FeatureContext extends RawMinkContext
 {
+    /**
+     * @Given (I )freeze the clock to :dateTime
```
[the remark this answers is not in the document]
B-Galati: Yes but it is more flexible like that and it does not cost anything

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,62 @@
...
+            throw new \InvalidArgumentException('Adherent must be a referent.');
+        }
+
+        $qb = $this->createQueryBuilder('history')
+            ->select('history.subscribedEmailType, history.action, COUNT(history) AS count')
+            ->andWhere('history.referentTag IN (:tags)')
+            ->andWhere('history.subscribedEmailType IN (:subscriptions)')
+            ->andWhere('history.date <= :until')
+            ->groupBy('history.subscribedEmailType, history.action')
+
```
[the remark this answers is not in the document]
B-Galati: The goal was to isolate the query itself from everything else to make it more readable. Should I keep like that or remove the empty line then?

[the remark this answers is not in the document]
thomasbisignani: IMHO you can remove it, we don't isolate that in our query builder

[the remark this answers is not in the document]
B-Galati: Alright :+1:

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,62 @@
...
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
+            ->setParameter('subscriptions', $subscriptionsTypes)
+            ->setParameter('until', $until)
+            ->getQuery()
+        ;
+
+        // Let's cache past data as they are never going to change
+        $firstDayOfMonth = (new Carbon('first day of this month'))->setTime(0, 0);
+        if ($firstDayOfMonth > $until) {
+            $qb->useResultCache(true, 25920000); // 30 days
```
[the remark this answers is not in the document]
B-Galati: OMG! Thank you so much for that :+1: :medal_military: :100: 
I changed it to 60 days

features/bootstrap/FeatureContext.php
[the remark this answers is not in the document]
thomasbisignani: Alright, ok :+1:

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,61 @@
...
+            ->andWhere('history.date <= :until')
+            ->groupBy('history.subscribedEmailType, history.action')
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
+            ->setParameter('subscriptions', $subscriptionsTypes)
+            ->setParameter('until', $until)
+            ->getQuery()
+        ;
+
+        // Let's cache past data as they are never going to change
+        $firstDayOfMonth = (new Carbon('first day of this month'))->setTime(0, 0);
```
ottaviano: same question for Carbon

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, EmailSubscriptionHistory::class);
+    }
+
+    /**
+     * Count active adherents (not users) history for each specified subscription types in the referent managed area before the specified date.
+     */
+    public function countAllByTypeForReferentManagedArea(Adherent $referent, array $subscriptionsTypes, \DateTime $until): array
```
ottaviano: what about rename this method on `...ForReferentTags(array $referentTags...` and remove the `!$referent->isReferent()` check?
because it's what you do, filter by tag (it can be the tags of one referent or of one simple adherent)?

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,61 @@
...
+        $results = $qb->getArrayResult();
+        $countBySubscriptionType = array_fill_keys($subscriptionsTypes, 0);
+
+        foreach ($results as ['action' => $action, 'subscribedEmailType' => $type, 'count' => $count]) {
+            if (EmailSubscriptionHistoryAction::SUBSCRIBE === $action) {
+                $countBySubscriptionType[$type] += $count;
+            } elseif (EmailSubscriptionHistoryAction::UNSUBSCRIBE === $action) {
+                $countBySubscriptionType[$type] -= $count;
+            } else {
+                throw new \RuntimeException("'$action' is not handled");
```
ottaviano: it's really useful in this context?

src/History/EmailSubscriptionHistoryHandler.php
```diff
@@ -90,4 +93,24 @@ private function createEmailSubscriptionHistory(
             }
         }
     }
+
+    public function queryCountByMonth(Adherent $referent, int $months = 6): array
+    {
+        foreach (range(0, $months - 1) as $month) {
+            $until = (new Carbon("last day of -$month month"))->setTime(23, 59, 59, 999);
```
ottaviano: what is the advantage of Carbon here instead of \DateTime class?

src/Repository/EmailSubscriptionHistoryRepository.php
B-Galati: It's future proof and does not cost anything ;-)

src/History/EmailSubscriptionHistoryHandler.php
B-Galati: It allows us to freeze/mock the clock/time in our test. We cannot do that if we use \DateTime directly.

For example in Behat or Phpunit we can set the clock date to 01/01/2018 so that when we call this method we would have always the same last 6 months:

- 01/2018
- 12/2017
- 11/2017
- etc.

src/Repository/EmailSubscriptionHistoryRepository.php
B-Galati: It's needed if we want our test to go into this part of the code. Otherwise it would never go in.

src/Repository/EmailSubscriptionHistoryRepository.php
B-Galati: I do agree :+1:

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,61 @@
...
+        if (!$referent->isReferent()) {
+            throw new \InvalidArgumentException('Adherent must be a referent.');
+        }
+
+        $qb = $this->createQueryBuilder('history')
+            ->select('history.subscribedEmailType, history.action, COUNT(history) AS count')
+            ->andWhere('history.referentTag IN (:tags)')
+            ->andWhere('history.subscribedEmailType IN (:subscriptions)')
+            ->andWhere('history.date <= :until')
+            ->groupBy('history.subscribedEmailType, history.action')
```
B-Galati: As pointed out by @ottaviano on Slack, the result will be wrong if adherent has multiple tags.

src/History/EmailSubscriptionHistoryHandler.php
ottaviano: I see, thanks 👍

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        if (!$referent->isReferent()) {
+            throw new \InvalidArgumentException('Adherent must be a referent.');
+        }
+
+        $qb = $this->createQueryBuilder('history')
+            ->select('history.adherentUuid, history.subscribedEmailType, history.action, COUNT(history) AS count')
+            ->where('history.referentTag IN (:tags)')
+            ->andWhere('history.subscribedEmailType IN (:subscriptions)')
+            ->andWhere('history.date <= :until')
```
ottaviano: do you have an index for this column?

B-Galati: Yes -> https://github.com/EnMarche/en-marche.fr/pull/2383/files#diff-82b04cbbba96b1c98b716e7a91d4f82eR19

src/Repository/EmailSubscriptionHistoryRepository.php
```diff
@@ -19,43 +19,56 @@ public function __construct(ManagerRegistry $registry)
...
 
         $qb = $this->createQueryBuilder('history')
-            ->select('history.subscribedEmailType, history.action, COUNT(history) AS count')
-            ->andWhere('history.referentTag IN (:tags)')
+            ->select('history.adherentUuid, history.subscribedEmailType, history.action, COUNT(history) AS count')
+            ->where('history.referentTag IN (:tags)')
             ->andWhere('history.subscribedEmailType IN (:subscriptions)')
             ->andWhere('history.date <= :until')
-            ->groupBy('history.subscribedEmailType, history.action')
+            ->groupBy('history.adherentUuid, history.subscribedEmailType, history.action')
```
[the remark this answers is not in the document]
B-Galati: It's not nullable for the record.

PR #2392, opened 2018-05-03
Specify log format for Nginx

Remg:


docker/prod/nginx.conf
```diff
@@ -11,7 +11,11 @@ events {
 }
 
 http {
-    access_log /dev/stdout;
+    log_format main '$http_x_forwarded_for - $remote_user [$time_local] '
```
atamano: could you also add the remote_addr info please ? something like this
```
log_format main '$http_x_forwarded_for ($remote_addr) - $remote_user [$time_local] '
'"$request" $status $body_bytes_sent "$http_referer" '
'"$http_user_agent"' ;
```

PR #2396, opened 2018-05-07
[ch3967] Refactor donations to register transactions.

cacahouwete:
https://app.clubhouse.io/enmarche/story/3967/etq-enmarche-je-souhaite-avoir-l-historique-de-toutes-les-transactions-paybox

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ PHPCSFIXER?=$(EXEC) php -d memory_limit=1024m vendor/bin/php-cs-fixer
 BEHAT_ARGS?=-vvv
 
 .DEFAULT_GOAL := help
-.PHONY: help start stop reset db db-diff db-migrate db-rollback db-load watch clear clean test tu tf tj lint ls ly lt
+.PHONY: help start stop reset db db-diff db-diff-dump db-migrate db-rollback db-load watch clear clean test tu tf tj lint ls ly lt
 .PHONY: lj build up perm deps cc phpcs phpcsfix tty tfp tfp-rabbitmq tfp-db test-behat test-phpunit-functional
 .PHONY: wait-for-rabbitmq wait-for-db security-check
 
@@ -72,6 +72,9 @@ db: vendor wait-for-db                                                          
 
 db-diff: vendor wait-for-db                                                                            ## Generate a migration by comparing your current database to your mapping information
 	$(EXEC) $(CONSOLE) doctrine:migration:diff
+
+db-diff-dump: vendor wait-for-db                                                                       ## Generate a migration by comparing your current database to your mapping information and display it in console
+	$(EXEC) $(CONSOLE) doctrine:schema:update --dump-sql
 
 db-migrate: vendor wait-for-db                                                                         ## Migrate database schema to the latest available version
 	$(EXEC) $(CONSOLE) doctrine:migration:migrate -n
--- a/app/config/services/services.xml
+++ b/app/config/services/services.xml
@@ -4,7 +4,10 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <defaults autowire="true" autoconfigure="true" public="false" />
+        <defaults autowire="true" autoconfigure="true" public="false">
+            <bind key="$donationRequestUtilsLocator" type="service" id="app.donation_request.utils.locator" />
+            <bind key="$environment">%kernel.environment%</bind>
+        </defaults>
 
         <!-- Adherents -->
         <service id="AppBundle\Membership\AdherentManager" />
@@ -56,20 +59,35 @@
         <service id="AppBundle\Validator\UniqueDonationSubscriptionValidator" />
         <service id="AppBundle\Donation\DonationFactory" />
         <service id="AppBundle\Donation\PayboxPaymentUnsubscription">
-            <argument type="service" id="app.mailer.transactional" key="$mailer"/>
+            <argument type="service" key="$mailer" id="app.mailer.transactional" />
+        </service>
+        <service id="AppBundle\Donation\TransactionSubscriber">
+            <argument type="service" key="$mailer" id="app.mailer.transactional" />
+        </service>
+        <service id="AppBundle\Donation\TransactionCallbackHandler">
+            <argument type="service" key="$mailer" id="app.mailer.transactional" />
+        </service>
+        <service id="AppBundle\Donation\DonationRequestHandler"/>
+        <service id="AppBundle\Donation\DonationRequestUtils" />
+
+        <service id="app.donation_request.utils.locator" class="Symfony\Component\DependencyInjection\ServiceLocator">
+            <argument type="collection">
+                <argument key="validator" type="service" id="validator" />
+                <argument key="session" type="service" id="session" />
+                <argument key="security.csrf.token_manager" type="service" id="security.csrf.token_manager" />
+            </argument>
+            <tag name="container.service_locator" />
         </service>
 
-        <service id="AppBundle\Donation\DonationRequestUtils">
-            <argument type="service" id="app.donation_request.utils.locator" key="$locator" />
-        </service>
+        <service id="AppBundle\Donation\PayboxFormFactory" />
 
         <!-- Membership -->
         <service id="AppBundle\Membership\MembershipRegistrationProcess"/>
 
         <!-- Event -->
-        <service id="AppBundle\Event\EventFactory" public="true"/>
+        <service id="AppBundle\Event\EventFactory" public="true" />
 
-        <service id="AppBundle\Event\EventRegistrationSubscriber" >
+        <service id="AppBundle\Event\EventRegistrationSubscriber">
             <argument type="service" id="app.mailer.transactional" />
         </service>
 
--- a/app/migrations/Version20180502142950.php
+++ b/app/migrations/Version20180502142950.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace Migrations;
+
+use AppBundle\Donation\DonationRequestUtils;
+use AppBundle\Entity\Donation;
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+class Version20180502142950 extends AbstractMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    public function up(Schema $schema): void
+    {
+        $donationRequestUtils = $this->container->get(DonationRequestUtils::class);
+
+        $this->addSql('CREATE INDEX donation_uuid_idx ON donations (uuid)');
+        $this->addSql('CREATE INDEX donation_email_idx ON donations (email_address)');
+        $this->addSql('CREATE INDEX donation_duration_idx ON donations (duration)');
+        $this->addSql('ALTER TABLE donations ADD status VARCHAR(25) DEFAULT NULL, MODIFY created_at DATETIME NOT NULL COMMENT \'(DC2Type:datetime_immutable)\', ADD paybox_order_ref VARCHAR(255) DEFAULT NULL');
+        $this->addSql('CREATE INDEX donation_status_idx ON donations (status)');
+        $results = $this->connection->executeQuery('SELECT uuid, CONCAT(first_name, \' \', last_name) AS full_name FROM donations')->fetchAll();
+
+        foreach ($results as ['uuid' => $uuid, 'full_name' => $fullName]) {
+            $slug = $donationRequestUtils->buildDonationReference(Uuid::fromString($uuid), $fullName);
+            $this->addSql(
+                'UPDATE donations SET paybox_order_ref = :slug WHERE uuid = :uuid',
+                ['slug' => $slug, 'uuid' => $uuid]
+            );
+        }
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.finished = 1 AND donations.donated_at IS NOT NULL',
+            ['status' => Donation::STATUS_FINISHED]
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.finished = 1 AND donations.donated_at IS NULL',
+            ['status' => Donation::STATUS_ERROR]
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.finished = 0',
+            ['status' => Donation::STATUS_WAITING_CONFIRMATION]
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.duration != 0 AND donations.donated_at IS NOT NULL',
+            ['status' => Donation::STATUS_SUBSCRIPTION_IN_PROGRESS]
+        );
+        $this->addSql('ALTER TABLE donations CHANGE donated_at updated_at DATETIME DEFAULT NULL');
+        $this->addSql('UPDATE donations SET donations.updated_at = donations.created_at WHERE donations.updated_at IS NULL');
+
+        $this->addSql('CREATE TABLE transaction (id INT UNSIGNED AUTO_INCREMENT NOT NULL, donation_id INT UNSIGNED DEFAULT NULL, paybox_result_code VARCHAR(100) DEFAULT NULL, paybox_authorization_code VARCHAR(100) DEFAULT NULL, paybox_payload JSON DEFAULT NULL COMMENT \'(DC2Type:json_array)\', paybox_date_time DATETIME DEFAULT NULL COMMENT \'(DC2Type:datetime_immutable)\', paybox_transaction_id VARCHAR(255) DEFAULT NULL, paybox_subscription_id VARCHAR(255) DEFAULT NULL, created_at DATETIME NOT NULL COMMENT \'(DC2Type:datetime_immutable)\', UNIQUE INDEX UNIQ_723705D15A4036C7 (paybox_transaction_id), INDEX IDX_723705D14DC1279C (donation_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
+        $this->addSql('CREATE INDEX transaction_result_idx ON transaction (paybox_result_code)');
+        $this->addSql('ALTER TABLE transaction ADD CONSTRAINT FK_723705D14DC1279C FOREIGN KEY (donation_id) REFERENCES donations (id)');
+        $this->addSql(
+            'INSERT INTO transaction (donation_id, paybox_result_code, paybox_authorization_code, paybox_payload, paybox_date_time, paybox_transaction_id, created_at)
+                    (SELECT id, paybox_result_code, paybox_authorization_code, paybox_payload,
+                      IF (paybox_payload->"$.date" IS NOT NULL AND paybox_payload->"$.time" IS NOT NULL, STR_TO_DATE(
+                        CONCAT(
+                          SUBSTR(paybox_payload->"$.date", 2, 2),
+                          \',\',
+                          SUBSTR(paybox_payload->"$.date", 4, 2),
+                          \',\',
+                          SUBSTR(paybox_payload->"$.date", 6, 4),
+                          \' \',
+                          SUBSTR(paybox_payload->"$.time", 2, 8)
+                        ),
+                        \'%d,%m,%Y %H:%i:%s\'
+                      ), NULL),
+                      IF (paybox_payload->"$.transaction" != \'0\', paybox_payload->"$.transaction", NULL),
+                      created_at
+                      FROM donations)'
+        );
+        $this->addSql('ALTER TABLE donations MODIFY status VARCHAR(25) NOT NULL, MODIFY updated_at DATETIME NOT NULL');
+        $this->addSql('ALTER TABLE donations DROP paybox_result_code, DROP paybox_authorization_code, DROP paybox_payload, DROP finished');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE donations ADD paybox_result_code VARCHAR(100) DEFAULT NULL, ADD paybox_authorization_code VARCHAR(100) DEFAULT NULL, ADD paybox_payload JSON DEFAULT NULL COMMENT \'(DC2Type:json_array)\', ADD finished TINYINT(1) NOT NULL, CHANGE updated_at donated_at DATETIME DEFAULT NULL');
+        $this->addSql(
+            'UPDATE transaction T, donations D
+            SET
+                D.paybox_payload = T.paybox_payload,
+                D.paybox_result_code = T.paybox_result_code,
+                D.paybox_authorization_code = T.paybox_authorization_code,
+                D.finished = IF (D.status like \'%waiting_confirmation%\', 0, 1),
+                D.donated_at = IF (D.status LIKE \'%waiting_confirmation%\' OR D.status LIKE \'%error%\', NULL, D.donated_at)
+            WHERE D.id = T.donation_id'
+        );
+        $this->addSql('DROP INDEX donation_uuid_idx ON donations');
+        $this->addSql('DROP INDEX donation_email_idx ON donations');
+        $this->addSql('DROP INDEX donation_duration_idx ON donations');
+        $this->addSql('DROP INDEX donation_status_idx ON donations');
+        $this->addSql('ALTER TABLE donations DROP status, DROP paybox_order_ref');
+        $this->addSql('DROP TABLE transaction');
+    }
+}
--- a/src/Controller/EnMarche/DonationController.php
+++ b/src/Controller/EnMarche/DonationController.php
@@ -3,9 +3,12 @@
 namespace AppBundle\Controller\EnMarche;
 
 use AppBundle\Donation\DonationRequest;
+use AppBundle\Donation\DonationRequestHandler;
 use AppBundle\Donation\DonationRequestUtils;
+use AppBundle\Donation\PayboxFormFactory;
 use AppBundle\Donation\PayboxPaymentSubscription;
 use AppBundle\Donation\PayboxPaymentUnsubscription;
+use AppBundle\Donation\TransactionCallbackHandler;
 use AppBundle\Entity\Donation;
 use AppBundle\Exception\PayboxPaymentUnsubscriptionException;
 use AppBundle\Exception\InvalidPayboxPaymentSubscriptionValueException;
@@ -19,6 +22,7 @@ use AppBundle\Repository\NewsletterSubscriptionRepository;
 use Psr\Log\LoggerInterface;
 use Ramsey\Uuid\Uuid;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
@@ -29,6 +33,9 @@ use Symfony\Component\HttpFoundation\Response;
  */
 class DonationController extends Controller
 {
+    public const RESULT_STATUS_EFFECTUE = 'effectue';
+    public const RESULT_STATUS_ERREUR = 'erreur';
+
     /**
      * @Route(defaults={"_enable_campaign_silence"=true}, name="donation_index")
      * @Method("GET")
@@ -52,7 +59,7 @@ class DonationController extends Controller
      * @Route("/coordonnees", defaults={"_enable_campaign_silence"=true}, name="donation_informations")
      * @Method({"GET", "POST"})
      */
-    public function informationsAction(Request $request, DonationRequestUtils $donationRequestUtils)
+    public function informationsAction(Request $request, DonationRequestUtils $donationRequestUtils, DonationRequestHandler $donationRequestHandler)
     {
         if (!$amount = $request->query->get('montant')) {
             return $this->redirectToRoute('donation_index');
@@ -67,7 +74,7 @@ class DonationController extends Controller
         }
 
         if ($form->isSubmitted() && $form->isValid()) {
-            $this->get('app.donation_request.handler')->handle($donationRequest);
+            $donationRequestHandler->handle($donationRequest);
             $donationRequestUtils->terminateDonationRequest();
 
             return $this->redirectToRoute('donation_pay', [
@@ -92,9 +99,9 @@ class DonationController extends Controller
      * )
      * @Method("GET")
      */
-    public function payboxAction(Donation $donation)
+    public function payboxAction(PayboxFormFactory $payboxFormFactory, Donation $donation)
     {
-        $paybox = $this->get('app.donation.form_factory')->createPayboxFormForDonation($donation);
+        $paybox = $payboxFormFactory->createPayboxFormForDonation($donation);
 
         return $this->render('donation/paybox.html.twig', [
             'url' => $paybox->getUrl(),
@@ -106,7 +113,7 @@ class DonationController extends Controller
      * @Route("/callback/{_callback_token}", defaults={"_enable_campaign_silence"=true}, name="donation_callback")
      * @Method("GET")
      */
-    public function callbackAction(Request $request, $_callback_token)
+    public function callbackAction(Request $request, TransactionCallbackHandler $transactionCallbackHandler, string $_callback_token)
     {
         $id = explode('_', $request->query->get('id'))[0];
 
@@ -114,7 +121,7 @@ class DonationController extends Controller
             return $this->redirectToRoute('donation_index');
         }
 
-        return $this->get('app.donation.transaction_callback_handler')->handle($id, $request, $_callback_token);
+        return $transactionCallbackHandler->handle($id, $request, $_callback_token);
     }
 
     /**
@@ -124,17 +131,19 @@ class DonationController extends Controller
      *     defaults={"_enable_campaign_silence"=true},
      *     name="donation_result"
      * )
+     * @ParamConverter("donation", options={"mapping": {"uuid": "uuid"}})
      * @Method("GET")
      */
     public function resultAction(
         Request $request,
-        Donation $donation,
         MembershipRegistrationProcess $membershipRegistrationProcess,
         AdherentRepository $adherentRepository,
-        NewsletterSubscriptionRepository $newsletterSubscriptionRepository
+        NewsletterSubscriptionRepository $newsletterSubscriptionRepository,
+        Donation $donation,
+        string $status
     ) {
         $retryUrl = null;
-        if (!$donation->isSuccessful()) {
+        if ($donation->hasError()) {
             $retryUrl = $this->generateUrl(
                 'donation_informations',
                 $this->get(DonationRequestUtils::class)->createRetryPayload($donation, $request)
@@ -144,7 +153,7 @@ class DonationController extends Controller
         $membershipRegistrationProcess->terminate();
 
         return $this->render('donation/result.html.twig', [
-            'successful' => $donation->isSuccessful(),
+            'successful' => self::RESULT_STATUS_EFFECTUE === $status,
             'error_code' => $request->query->get('code'),
             'donation' => $donation,
             'retry_url' => $retryUrl,
--- a/src/Controller/EnMarche/UserController.php
+++ b/src/Controller/EnMarche/UserController.php
@@ -13,6 +13,7 @@ use AppBundle\Membership\MembershipRequest;
 use AppBundle\Membership\MembershipRequestHandler;
 use AppBundle\Membership\UnregistrationCommand;
 use AppBundle\Repository\DonationRepository;
+use AppBundle\Repository\TransactionRepository;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
@@ -41,12 +42,12 @@ class UserController extends Controller
      * @Route("/mes-dons", name="app_user_profile_donation")
      * @Method("GET")
      */
-    public function profileDonationAction(DonationRepository $donationRepository): Response
+    public function profileDonationAction(DonationRepository $donationRepository, TransactionRepository $transactionRepository): Response
     {
         $userEmail = $this->getUser()->getEmailAddress();
 
         return $this->render('user/my_donation.html.twig', [
-            'donations' => $donationRepository->findByEmailAddressOrderedByDonatedAt($userEmail),
+            'successful_transactions' => $transactionRepository->findAllSuccessfulTransactionByEmail($userEmail),
             'subscribed_donations' => $donationRepository->findAllSubscribedDonationByEmail($userEmail),
             'last_subscription_ended' => $donationRepository->findLastSubscriptionEndedDonationByEmail($userEmail),
         ]);
--- a/src/DataFixtures/ORM/LoadDonationData.php
+++ b/src/DataFixtures/ORM/LoadDonationData.php
@@ -5,53 +5,76 @@ namespace AppBundle\DataFixtures\ORM;
 use AppBundle\Donation\PayboxPaymentSubscription;
 use AppBundle\Entity\Adherent;
 use AppBundle\Entity\Donation;
+use AppBundle\Entity\Transaction;
+use Cocur\Slugify\Slugify;
 use Doctrine\Bundle\FixturesBundle\Fixture;
 use Doctrine\Common\Persistence\ObjectManager;
 use Ramsey\Uuid\Uuid;
 
 class LoadDonationData extends Fixture
 {
+    private $slugify;
+
+    public function __construct()
+    {
+        $this->slugify = Slugify::create();
+    }
+
     public function load(ObjectManager $manager)
     {
+        /** @var Adherent $adherent0 */
+        $adherent0 = $this->getReference('adherent-1');
         /** @var Adherent $adherent1 */
         $adherent1 = $this->getReference('adherent-4');
         /** @var Adherent $adherent2 */
         $adherent2 = $this->getReference('adherent-3');
 
-        $donation0 = $this->create($adherent1, 50.);
-        $donation1 = $this->create($adherent2, 50.);
-        $donation2 = $this->create($adherent2, 40.);
-        $donation3 = $this->create($adherent2, 60., PayboxPaymentSubscription::UNLIMITED);
-        $donation4 = $this->create($adherent2, 100., PayboxPaymentSubscription::UNLIMITED);
+        $donationNormal = $this->createDonation($adherent0);
+        $donationMonthly = $this->createDonation($adherent0, 42., PayboxPaymentSubscription::UNLIMITED);
+        $donation0 = $this->createDonation($adherent1, 50.);
+        $donation1 = $this->createDonation($adherent2, 50.);
+        $donation2 = $this->createDonation($adherent2, 40.);
+        $donation3 = $this->createDonation($adherent2, 60., PayboxPaymentSubscription::UNLIMITED);
+        $donation4 = $this->createDonation($adherent2, 100., PayboxPaymentSubscription::UNLIMITED);
+
+        $transactionNormal = $this->createTransaction($donationNormal);
+        $transactionMonthly = $this->createTransaction($donationMonthly);
+        $transaction0 = $this->createTransaction($donation0);
+        $transaction1 = $this->createTransaction($donation1);
+        $transaction2 = $this->createTransaction($donation2);
+        $transaction3 = $this->createTransaction($donation3);
+        $transaction4 = $this->createTransaction($donation4);
 
         $donation3->stopSubscription();
 
-        $this->setDonateAt($donation2, '-1 day');
-        $this->setDonateAt($donation3, '-100 day');
-        $this->setDonateAt($donation4, '-50 day');
+        $this->setDonateAt($transaction2, '-1 day');
+        $this->setDonateAt($transaction3, '-100 day');
+        $this->setDonateAt($transaction4, '-50 day');
 
+        $manager->persist($donationNormal);
+        $manager->persist($donationMonthly);
         $manager->persist($donation0);
         $manager->persist($donation1);
         $manager->persist($donation2);
         $manager->persist($donation3);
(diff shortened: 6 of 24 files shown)


src/Repository/TransactionRepository.php
```diff
@@ -0,0 +1,46 @@
...
+     * @return Transaction[]
+     */
+    public function findAllTransactionSuccess(string $emailAddress, string $ordered = 'ASC'): array
+    {
+        return $this->createQueryBuilder('transaction')
+            ->innerJoin('transaction.donation', 'donation')
+            ->andWhere('donation.emailAddress = :email')
+            ->andWhere('transaction.payboxResultCode = :resultCode')
+            ->setParameters([
+                'resultCode' => '00000',
```
thomasbisignani: You should use a const (in the `Transaction` entity) for this success code.

app/config/services/services.xml
```diff
@@ -55,14 +57,27 @@
...
+            <argument type="collection">
+                <argument key="validator" type="service" id="validator" />
+                <argument key="session" type="service" id="session" />
+                <argument key="security.csrf.token_manager" type="service" id="security.csrf.token_manager" />
+            </argument>
+            <tag name="container.service_locator" />
+        </service>
+
+        <service id="AppBundle\Donation\PayboxFormFactory">
+            <argument>%kernel.environment%</argument>
```
HeahDude: Could be bound to `$environment`?

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private $donationRequestUtils;
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('CREATE INDEX donation_uuid_idx ON donations (uuid)');
+        $this->addSql('ALTER TABLE donations ADD status VARCHAR(25) DEFAULT NULL, CHANGE donated_at updated_at DATETIME DEFAULT NULL, ADD paybox_order_ref VARCHAR(255) DEFAULT NULL');
+        $results = $this->connection->executeQuery('SELECT uuid, CONCAT(first_name, \' \', last_name) AS full_name FROM donations')->fetchAll();
+
+        foreach ($results as $result) {
```
HeahDude: Wouldn't it be more readable to list args in the `foreach` directly, like:
```php
foreach ($results as ['uuid' => $uuid, 'full_name' => $fullName]) {
```
?

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+            ['uuid' => $uuid, 'full_name' => $fullName] = $result;
+            $slug = $this->donationRequestUtils->buildDonationReference(Uuid::fromString($uuid), $fullName);
+            $this->addSql("UPDATE donations SET paybox_order_ref = '$slug' WHERE uuid = '$uuid'");
```
HeahDude: Please use the second arg to pass parameters.

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+        }
+        $this->addSql('UPDATE donations SET donations.status = \'finished\' WHERE donations.finished = 1');
```
HeahDude: Couldn't you use constants here?

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->addSql('UPDATE donations SET donations.status = \'waiting_confirmation\' WHERE donations.finished = 0');
+        $this->addSql('UPDATE donations SET donations.status = \'subscription_in_progress\' WHERE donations.duration != 0');
+        $this->addSql('UPDATE donations SET donations.updated_at = donations.created_at WHERE donations.updated_at IS NULL');
+
+        $this->addSql('CREATE TABLE transaction (id INT UNSIGNED AUTO_INCREMENT NOT NULL, donation_id INT UNSIGNED DEFAULT NULL, paybox_result_code VARCHAR(100) DEFAULT NULL, paybox_authorization_code VARCHAR(100) DEFAULT NULL, paybox_payload JSON DEFAULT NULL COMMENT \'(DC2Type:json_array)\', paybox_date_time DATETIME DEFAULT NULL, paybox_transaction_id VARCHAR(255) DEFAULT NULL, paybox_subscription_id VARCHAR(255) DEFAULT NULL, created_at DATETIME NOT NULL, UNIQUE INDEX UNIQ_723705D15A4036C7 (paybox_transaction_id), INDEX IDX_723705D14DC1279C (donation_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
```
HeahDude: Unsigned ids?

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->addSql('ALTER TABLE transaction ADD CONSTRAINT FK_723705D14DC1279C FOREIGN KEY (donation_id) REFERENCES donations (id)');
+        $this->addSql(
+            'INSERT INTO transaction (donation_id, paybox_result_code, paybox_authorization_code, paybox_payload, paybox_date_time, paybox_transaction_id, created_at)
+                    (SELECT id, paybox_result_code, paybox_authorization_code, paybox_payload,
+                      IF (paybox_payload->"$.date" IS NOT NULL AND paybox_payload->"$.time" IS NOT NULL, STR_TO_DATE(
+                        CONCAT(
+                          SUBSTR(paybox_payload->"$.date", 2, 2),
```
HeahDude: Please use second arg to pass parameters to the query

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+        );
+        $this->addSql('ALTER TABLE donations DROP status, DROP paybox_order_ref');
+        $this->addSql('DROP TABLE transaction');
+        $this->addSql('DROP INDEX donation_uuid_idx ON donations');
+    }
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        if (!$container) {
+            throw new \LogicException('We need container to get DonationRequestUtils service');
```
HeahDude: "... the container ..."

src/Controller/EnMarche/DonationController.php
```diff
@@ -63,7 +67,7 @@ public function informationsAction(Request $request, DonationRequestUtils $donat
         }
 
         if ($form->isSubmitted() && $form->isValid()) {
-            $this->get('app.donation_request.handler')->handle($donationRequest);
+            $donationRequestHandler->handle($donationRequest);
```
HeahDude: You should keep it public to initialize the service only when the form is submitted, or implement`ServiceSubscriberInterface`

src/Donation/DonationStatusEnum.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace AppBundle\Donation;
+
+use MyCLabs\Enum\Enum;
```
HeahDude: :puke:, why adding this lib to EM?

src/Donation/DonationStatusEnum.php
```diff
@@ -0,0 +1,15 @@
...
+
+class DonationStatusEnum extends Enum
+{
+    public const WAITING_CONFIRMATION = 'waiting_confirmation';
```
HeahDude: Could be declared in `Donation` class.

src/Entity/Donation.php
```diff
@@ -23,78 +27,93 @@ class Donation implements GeoPointInterface
     use EntityPersonNameTrait;
 
     /**
+     * @var int
```
HeahDude: No need when the constructor has type hints already

src/Entity/Donation.php
```diff
@@ -279,10 +259,55 @@ public function nextDonationAt(\DateTime $fromDay = null): \DateTime
...
 
-        $donationDate = clone $this->donatedAt;
+        $donationDate = clone $this->createdAt;
 
         return $donationDate->modify(
             sprintf('+%d months', $donationDate->diff($fromDay)->m + 1)
         );
     }
+
+    public function getTransactions(): Collection
```
HeahDude: Please add:
```php
/**
 * @var Collection|Transaction[]
 */
```

src/Entity/Donation.php
```diff
@@ -279,10 +259,55 @@ public function nextDonationAt(\DateTime $fromDay = null): \DateTime
...
+    {
+        return $this->transactions;
+    }
+
+    public function getStatus(): string
+    {
+        return $this->status;
+    }
+
+    public function isError(): bool
```
HeahDude: `hasError`?

src/Entity/Donation.php
```diff
@@ -279,10 +259,55 @@ public function nextDonationAt(\DateTime $fromDay = null): \DateTime
...
+    {
+        return $this->payboxOrderRef;
+    }
+
+    public function getPayboxOrderRefWithSuffix(): string
+    {
+        return $this->payboxOrderRef.PayboxPaymentSubscription::getCommandSuffix($this->amount, $this->duration);
+    }
+
+    public function getUpdatedAt(): \DateTime
```
HeahDude: Please use `\DateTimeInterface`

src/Entity/Transaction.php
```diff
@@ -0,0 +1,164 @@
...
+     * @var Donation
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Donation", inversedBy="transactions")
+     */
+    private $donation;
+
+    public function __construct(Donation $donation, array $payboxPayload)
+    {
+        $this->donation = $donation;
+        $this->createdAt = new \DateTime();
```
HeahDude: `DateTimeImmutable`

src/Entity/Transaction.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->payboxPayload = $payboxPayload;
+        [
+            'result' => $this->payboxResultCode,
+        ] = $payboxPayload;
+
+        if (isset($payboxPayload['transaction']) && $transaction = $payboxPayload['transaction']) {
```
HeahDude:
```php
if ($transaction = $payboxPayload['transaction'] ??  false) {
```

src/Entity/Transaction.php
```diff
@@ -0,0 +1,164 @@
...
+            $this->payboxTransactionId = $transaction;
+        }
+
+        if (isset($payboxPayload['date'], $payboxPayload['time'])) {
+            [
+                'date' => $date,
+                'time' => $time,
+            ] = $payboxPayload;
+            $this->payboxDateTime = \DateTime::createFromFormat(
```
HeahDude: Can you use immutable date?

src/Entity/Transaction.php
```diff
@@ -0,0 +1,164 @@
...
+                'Y/m/d H:i:s',
+                substr($date, 4, 4).'/'.substr($date, 2, 2).'/'.substr($date, 0, 2).' '.$time
+            );
+        }
+
+        if (isset($payboxPayload['authorization']) && $authorization = $payboxPayload['authorization']) {
```
HeahDude: Same as above?

src/Entity/Transaction.php
```diff
@@ -0,0 +1,164 @@
...
+            $this->payboxAuthorizationCode = $authorization;
+        }
+
+        if (isset($payboxPayload['subscription']) && $subscription = $payboxPayload['subscription']) {
```
HeahDude: here too

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,85 @@
...
+                D.finished = IF (D.status like \'%waiting_confirmation%\', 0, 1),
+                D.donated_at = IF (D.status like \'%waiting_confirmation%\', NULL, D.donated_at)
+            WHERE D.id = T.donation_id'
+        );
+        $this->addSql('ALTER TABLE donations DROP status, DROP paybox_order_ref');
+        $this->addSql('DROP TABLE transaction');
+        $this->addSql('DROP INDEX donation_uuid_idx ON donations');
+    }
+
+    public function setContainer(ContainerInterface $container = null)
```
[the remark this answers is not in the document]
HeahDude: No migrations are NOT autoloaded, so not configured but the container, so do not use autowiring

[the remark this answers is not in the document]
thomasbisignani: It's possible since Symfony 2.4 : https://symfony.com/doc/master/bundles/DoctrineMigrationsBundle/index.html#container-aware-migrations

It has already been used here https://github.com/EnMarche/en-marche.fr/blob/master/app/migrations/Version20170904153003.php#L14

src/Donation/TransactionSubscriber.php
```diff
@@ -6,36 +6,50 @@
...
     {
         $this->manager = $manager;
-        $this->mailer = $mailer;
+        $this->mailerTransactional = $mailerTransactional;
         $this->requestStack = $requestStack;
     }
 
+    /**
+     * {@inheritdoc}
+     */
```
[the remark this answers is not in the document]
cacahouwete: @HeahDude told me it's useful when you read a PR by adding `{@inheritdoc}` you explicitly say it's an implementation or an override. I will remove it if it's useless for you ^^,

src/Entity/Transaction.php
```diff
@@ -0,0 +1,164 @@
...
+     */
+    private $donation;
+
+    public function __construct(Donation $donation, array $payboxPayload)
+    {
+        $this->donation = $donation;
+        $this->createdAt = new \DateTime();
+        $this->payboxPayload = $payboxPayload;
+        [
+            'result' => $this->payboxResultCode,
```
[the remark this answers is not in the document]
cacahouwete: I don t want `$payboxPayload['result'] = $this->payboxResultCode;`.

I want `$this->payboxResultCode = $payboxPayload['result'];` :D

app/migrations/Version20180502142950.php
cacahouwete: Have you already seen negative IDs ? ^^, In witch case signed ids could be useful when it's in auto_increment ?

app/migrations/Version20180502142950.php
cacahouwete: It's not a parameter ! It's mysql way to read and use json data

app/migrations/Version20180502142950.php
ottaviano: https://dev.mysql.com/doc/refman/5.7/en/integer-types.html ;)

src/Entity/Transaction.php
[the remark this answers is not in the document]
thomasbisignani: Yes so I prefer this short syntax `$this->payboxResultCode = $payboxPayload['result'];` :D

src/Donation/DonationStatusEnum.php
cacahouwete: Not me ^^, last time I did it in class, somebody told me to use Enum. I agree with you this can be done in Donation class directly.

B-Galati: I did, this is a great lib. What would you do instead ?

B-Galati: Note that you are not supposed to use it every single. Use it when it makes sense to you. I did not check for that case.

src/Controller/EnMarche/UserController.php
```diff
@@ -39,12 +40,12 @@ public function profileOverviewAction(): Response
...
      * @Method("GET")
      */
-    public function profileDonationAction(DonationRepository $donationRepository): Response
+    public function profileDonationAction(DonationRepository $donationRepository, TransactionRepository $transactionRepository): Response
     {
         $userEmail = $this->getUser()->getEmailAddress();
 
         return $this->render('user/my_donation.html.twig', [
-            'donations' => $donationRepository->findByEmailAddressOrderedByDonatedAt($userEmail, 'DESC'),
+            'transactions_success' => $transactionRepository->findAllSuccessfulTransactionByEmail($userEmail),
```
thomasbisignani: For more consistent finally I think we can rename that to `findAllSuccessTransactionByEmail`

Makefile
```diff
@@ -73,6 +73,9 @@ db: vendor wait-for-db
 db-diff: vendor wait-for-db                                                                            ## Generate a migration by comparing your current database to your mapping information
 	$(EXEC) $(CONSOLE) doctrine:migration:diff
 
+db-dump: vendor wait-for-db                                                                            ## Generate a migration by comparing your current database to your mapping information
```
B-Galati: The name is misleading, I would go for `db-diff-dump`.

Please update the comment.

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,91 @@
...
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+class Version20180502142950 extends AbstractMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    public function up(Schema $schema): void
+    {
+        if (!$this->container) {
```
thomasbisignani: IMO this test is useless no ?

src/Entity/Donation.php
```diff
@@ -279,10 +246,58 @@ public function nextDonationAt(\DateTime $fromDay = null): \DateTime
...
+    }
+
+    public function isSubscriptionInProgress(): bool
+    {
+        return self::STATUS_SUBSCRIPTION_IN_PROGRESS === $this->getStatus();
+    }
+
+    public function isSuccess(): bool
+    {
+        return self::STATUS_WAITING_CONFIRMATION !== $this->getStatus() && !$this->hasError();
```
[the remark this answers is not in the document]
cacahouwete: Than cancel monthly donation, refunded donation are not in success scope ?

src/Entity/Donation.php
```diff
@@ -114,43 +125,39 @@ public function __construct(
...
+        $this->createdAt = new \DateTimeImmutable();
         $this->duration = $duration;
+        $this->status = self::STATUS_WAITING_CONFIRMATION;
+        $this->payboxOrderRef = $payboxOrderRef;
+        $this->transactions = new ArrayCollection();
     }
 
     public function __toString(): string
     {
         return $this->lastName.' '.$this->firstName.' ('.($this->amount / 100).' €)';
```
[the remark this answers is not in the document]
cacahouwete: Why not^^, I didn't do this code but I can refacto ;)

app/migrations/Version20180502142950.php
cacahouwete: Don't know It was to be sure I have a container set. But I can remove it

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
+<?php
+
+namespace AppBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity(repositoryClass="AppBundle\Repository\TransactionRepository")
```
B-Galati: Disable indexation for Algolia please

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+     * @ORM\Column(type="datetime_immutable")
+     */
+    private $createdAt;
+
+    /**
+     * @var Donation
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Donation", inversedBy="transactions")
+     */
+    private $donation;
```
B-Galati: We want to avoid  bidirectional association whenever it's possible so please make sure it's needed in both places.

https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/best-practices.html#constrain-relationships-as-much-as-possible

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getPayboxPayloadAsJson(): string
+    {
+        return json_encode($this->payboxPayload, JSON_PRETTY_PRINT);
```
B-Galati: Why `JSON_PRETTY_PRINT` ?

src/Entity/Donation.php
```diff
@@ -22,6 +25,13 @@ class Donation implements GeoPointInterface
     use EntityPostAddressTrait;
     use EntityPersonNameTrait;
 
+    public const STATUS_WAITING_CONFIRMATION = 'waiting_confirmation';
```
B-Galati: Just my thought, no need to change but it was ok with the enum class.

src/Repository/TransactionRepository.php
```diff
@@ -0,0 +1,46 @@
...
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, Transaction::class);
+    }
+
+    public function findByPayboxTransactionId(string $transactionId): ?Transaction
+    {
+        return $this->createQueryBuilder('t')
+            ->where('t.payboxTransactionId = :transactionId')
+            ->setParameters(['transactionId' => $transactionId])
```
B-Galati: You can use `setParameter` if you only have one.

src/Repository/TransactionRepository.php
```diff
@@ -0,0 +1,46 @@
...
+            ->getOneOrNullResult()
+        ;
+    }
+
+    /**
+     * @param string $emailAddress
+     *
+     * @return Transaction[]
+     */
+    public function findAllSuccessTransactionByEmail(string $emailAddress): array
```
B-Galati: ~~`SuccessTransaction`~~ -> `SuccessfulTransaction` 
because we say a successful transaction not a success transaction.

src/Controller/EnMarche/UserController.php
```diff
@@ -39,12 +40,12 @@ public function profileOverviewAction(): Response
...
      * @Method("GET")
      */
-    public function profileDonationAction(DonationRepository $donationRepository): Response
+    public function profileDonationAction(DonationRepository $donationRepository, TransactionRepository $transactionRepository): Response
     {
         $userEmail = $this->getUser()->getEmailAddress();
 
         return $this->render('user/my_donation.html.twig', [
-            'donations' => $donationRepository->findByEmailAddressOrderedByDonatedAt($userEmail, 'DESC'),
+            'transactions_success' => $transactionRepository->findAllSuccessTransactionByEmail($userEmail),
```
B-Galati: ~~transactions_success~~  -> successful_transactions

src/Entity/Transaction.php
cacahouwete: Don't know,  I just move this method from Donation.

src/Entity/Donation.php
```diff
@@ -43,58 +53,59 @@ class Donation implements GeoPointInterface
...
-    private $payboxResultCode;
-
-    /**
-     * @ORM\Column(length=100, nullable=true)
+     * @var string|null
+     *
+     * @ORM\Column(length=50, nullable=true)
      */
-    private $payboxAuthorizationCode;
+    private $clientIp;
```
B-Galati: This client ip should be moved to transaction no ? I don't really understand its purpose with the new entity `Transaction`

src/Entity/Donation.php
```diff
@@ -279,10 +246,58 @@ public function nextDonationAt(\DateTime $fromDay = null): \DateTime
...
+    {
+        return self::STATUS_FINISHED === $this->getStatus();
+    }
+
+    public function isSubscriptionInProgress(): bool
+    {
+        return self::STATUS_SUBSCRIPTION_IN_PROGRESS === $this->getStatus();
+    }
+
+    public function isSuccess(): bool
```
B-Galati: Why do we keep this method ? I don't see how a donation can be successful when it's finished or in progress. 

IMHO, it does not make sense anymore.

src/Mailer/Message/DonationMessage.php
```diff
@@ -2,12 +2,14 @@
 
 namespace AppBundle\Mailer\Message;
 
-use AppBundle\Entity\Donation;
+use AppBundle\Entity\Transaction;
 
 final class DonationMessage extends Message
 {
-    public static function createFromDonation(Donation $donation): self
+    public static function createFromDonation(Transaction $transaction): self
```
B-Galati: Rename the method to `createFromTransaction` please.

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+    private $createdAt;
+
+    /**
+     * @var Donation
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Donation", inversedBy="transactions")
+     */
+    private $donation;
+
+    public function __construct(Donation $donation, array $payboxPayload)
```
B-Galati: Please unit-test this

tests/Controller/EnMarche/DonationControllerTest.php
```diff
@@ -182,10 +183,22 @@ public function testSuccessFulProcess(int $duration)
...
+            $this->assertTrue($donation->isSubscriptionInProgress());
+            $donation->nextDonationAt();
+        } else {
+            $this->assertTrue($donation->isFinished());
+
+            $this->expectException(\LogicException::class);
+            $this->expectExceptionMessage('Donation without subscription can\'t have next donation date.');
+            $donation->nextDonationAt();
+        }
+        $this->assertNotEmpty($donation->getTransactions());
```
B-Galati: It would be better to assert the number of elements you expect.

tests/Controller/EnMarche/DonationControllerTest.php
```diff
@@ -182,10 +183,22 @@ public function testSuccessFulProcess(int $duration)
...
+        /** @var Transaction $transaction */
+        $transaction = $donation->getTransactions()->first();
+        $this->assertSame('00000', $transaction->getPayboxResultCode());
+        $this->assertSame('XXXXXX', $transaction->getPayboxAuthorizationCode());
```
B-Galati: Perhaps you would need to add more tests regarding this object here no ?

src/Donation/DonationRequestUtils.php
```diff
@@ -28,32 +30,32 @@ class DonationRequestUtils
...
-    private $locator;
+    private $donationRequestUtilsLocator;
     private $slugify;
     private $membershipRegistrationProcess;
 
-    public function __construct(ServiceLocator $locator, Slugify $slugify, MembershipRegistrationProcess $membershipRegistrationProcess)
+    public function __construct(ServiceLocator $donationRequestUtilsLocator, Slugify $slugify, MembershipRegistrationProcess $membershipRegistrationProcess)
     {
-        $this->locator = $locator;
+        $this->donationRequestUtilsLocator = $donationRequestUtilsLocator;
```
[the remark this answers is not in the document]
cacahouwete: I bound the argument name to a service

```xml
        <defaults autowire="true" autoconfigure="true" public="false" >
            <bind key="$mailerTransactional" type="service" id="app.mailer.transactional" />
            <bind key="$donationRequestUtilsLocator" type="service" id="app.donation_request.utils.locator" />
            <bind key="$environment" >%kernel.environment%</bind>
        </defaults>
```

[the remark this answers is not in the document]
B-Galati: Not a good reason. The name of dependencies should not be influenced by your DI tool.

src/Entity/Donation.php
cacahouwete: It's used in result page when the user come back to the website to display success page or error page

B-Galati: How is this accurate ? What's the workflow of these statuses ?

src/Entity/Donation.php
cacahouwete: Client IP is set only one time when the order(donation) is created after that all transactions received are a responses from paybox to say if it's ok or not and why. No ?

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,91 @@
...
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.finished = 0',
+            ['status' => Donation::STATUS_WAITING_CONFIRMATION]
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.duration != 0',
+            ['status' => Donation::STATUS_SUBSCRIPTION_IN_PROGRESS]
+        );
+        $this->addSql('UPDATE donations SET donations.updated_at = donations.created_at WHERE donations.updated_at IS NULL');
```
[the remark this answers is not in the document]
cacahouwete: When the status or any other things change I think it's good to know when ?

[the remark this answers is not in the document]
B-Galati: Sorry it's not my question.

I meant, why do you need to set this field as required while it was not required before. It feels like an unneeded change.

src/Entity/Donation.php
B-Galati: Ok got it now. Thanks ;-) 
Could you add a comment about this please?

src/Controller/EnMarche/DonationController.php
```diff
@@ -136,7 +141,7 @@ public function resultAction(Request $request, Donation $donation, MembershipReg
         $membershipRegistrationProcess->terminate();
 
         return $this->render('donation/result.html.twig', [
-            'successful' => $donation->isSuccessful(),
+            'successful' => $donation->isSuccess(),
```
ottaviano: it was correct with `isSuccessful`, why change this?

src/Controller/EnMarche/UserController.php
ottaviano: or maybe just `'transactions' => ...`?

src/DataFixtures/ORM/LoadDonationData.php
```diff
@@ -63,20 +72,24 @@ public function create(Adherent $adherent, float $amount = 50.0, int $duration =
...
         ]);
 
         return $donation;
     }
 
     public function setDonateAt(Donation $donation, string $modifier): void
     {
-        $reflectDonation = new \ReflectionObject($donation);
-        $reflectDonationAt = $reflectDonation->getProperty('donatedAt');
+        $reflectDonation = new \ReflectionObject($transaction = $donation->getTransactions()->first());
```
ottaviano: wrong variable name

src/Donation/TransactionCallbackHandler.php
```diff
@@ -15,15 +16,17 @@ class TransactionCallbackHandler
...
     private $donationRequestUtils;
+    private $transactionRepository;
 
-    public function __construct(UrlGeneratorInterface $router, ObjectManager $entityManager, MailerService $mailer, DonationRequestUtils $donationRequestUtils)
+    public function __construct(UrlGeneratorInterface $router, ObjectManager $entityManager, MailerService $mailerTransactional, DonationRequestUtils $donationRequestUtils, TransactionRepository $transactionRepository)
     {
         $this->router = $router;
         $this->entityManager = $entityManager;
-        $this->mailer = $mailer;
+        $this->mailerTransactional = $mailerTransactional;
```
ottaviano: agree with @B-Galati, this change is useless

src/Donation/TransactionCallbackHandler.php
```diff
@@ -15,15 +16,17 @@ class TransactionCallbackHandler
 {
     private $router;
     private $entityManager;
-    private $mailer;
+    private $mailerTransactional;
     private $donationRequestUtils;
+    private $transactionRepository;
 
-    public function __construct(UrlGeneratorInterface $router, ObjectManager $entityManager, MailerService $mailer, DonationRequestUtils $donationRequestUtils)
+    public function __construct(UrlGeneratorInterface $router, ObjectManager $entityManager, MailerService $mailerTransactional, DonationRequestUtils $donationRequestUtils, TransactionRepository $transactionRepository)
```
ottaviano: CS: multiligne pls

src/Donation/TransactionCallbackHandler.php
```diff
@@ -34,20 +37,23 @@ public function handle(string $uuid, Request $request, string $callbackToken): R
             return new RedirectResponse($this->router->generate('donation_index'));
         }
 
-        if (!$donation->isFinished()) {
-            $donation->finish($this->donationRequestUtils->extractPayboxResultFromCallBack($request, $callbackToken));
+        $payload = $this->donationRequestUtils->extractPayboxResultFromCallBack($request, $callbackToken);
```
ottaviano: `CallBack` -> `Callback`

src/Donation/TransactionSubscriber.php
```diff
@@ -6,36 +6,45 @@
...
-    private $mailer;
+    private $mailerTransactional;
     private $requestStack;
 
-    public function __construct(ObjectManager $manager, MailerService $mailer, RequestStack $requestStack)
+    public function __construct(ObjectManager $manager, MailerService $mailerTransactional, RequestStack $requestStack)
     {
         $this->manager = $manager;
-        $this->mailer = $mailer;
+        $this->mailerTransactional = $mailerTransactional;
```
ottaviano: same for name change

src/Donation/TransactionSubscriber.php
```diff
@@ -6,36 +6,45 @@
...
         $this->requestStack = $requestStack;
     }
 
+    public static function getSubscribedEvents()
+    {
+        return [
+            PayboxEvents::PAYBOX_IPN_RESPONSE => [
+                'onPayboxIpnResponse',
```
ottaviano: CS: IMO not need to be multiline here

src/Entity/Donation.php
```diff
@@ -43,58 +53,59 @@ class Donation implements GeoPointInterface
...
+     * @var string|null
+     *
+     * @ORM\Column(length=50, nullable=true)
      */
-    private $payboxAuthorizationCode;
+    private $clientIp;
 
     /**
-     * @ORM\Column(type="json_array", nullable=true)
+     * @ORM\Column(type="datetime_immutable")
```
ottaviano: `@var` missing

src/Entity/Donation.php
```diff
@@ -114,43 +125,39 @@ public function __construct(
...
         $this->duration = $duration;
+        $this->status = self::STATUS_WAITING_CONFIRMATION;
+        $this->payboxOrderRef = $payboxOrderRef;
+        $this->transactions = new ArrayCollection();
     }
 
     public function __toString(): string
     {
-        return $this->lastName.' '.$this->firstName.' ('.($this->amount / 100).' €)';
+        return sprintf('%s %s (%s €)', $this->lastName, $this->firstName, (string) ($this->amount / 100));
```
ottaviano: what about `(%.2f €)` for the amount and remove cast string?

src/Entity/Donation.php
```diff
@@ -5,12 +5,15 @@
...
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Mapping\Annotation as Gedmo;
 use libphonenumber\PhoneNumber;
 use Ramsey\Uuid\UuidInterface;
 
 /**
- * @ORM\Table(name="donations")
+ * @ORM\Table(name="donations", indexes={@ORM\Index(name="donation_uuid_idx", columns={"uuid"})})
```
ottaviano: can you add the index for `email`, `duration` and `status` too pls?

app/migrations/Version20180502142950.php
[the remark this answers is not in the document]
cacahouwete: It's to have the last update, donated_at transformed to updated_at, donated_at can be null, updated_at will be not null because we want to know when was the last change. At the entity creation updated_at have the same value to created_at

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+    {
+        $this->donation = $donation;
+        $this->createdAt = new \DateTimeImmutable();
+        $this->payboxPayload = $payboxPayload;
+        $this->payboxResultCode = $payboxPayload['result'];
+
+        if ($transaction = $payboxPayload['transaction'] ?: false) {
```
ottaviano: double if for one action?
what about `$this->payboxTransactionId = $payboxPayload['transaction'] ?? false;`

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+            $this->payboxTransactionId = $transaction;
+        }
+
+        if (isset($payboxPayload['date'], $payboxPayload['time'])) {
+            [
+                'date' => $date,
+                'time' => $time,
+            ] = $payboxPayload;
+            $this->payboxDateTime = \DateTimeImmutable::createFromFormat(
```
ottaviano: what is the value of $date and $time? do you have example?

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+                'Y/m/d H:i:s',
+                substr($date, 4, 4).'/'.substr($date, 2, 2).'/'.substr($date, 0, 2).' '.$time
+            );
+        }
+
+        if ($authorization = $payboxPayload['authorization'] ?: false) {
```
ottaviano: why do you double if for this?

src/Entity/Transaction.php
```diff
@@ -0,0 +1,170 @@
...
+            $this->payboxAuthorizationCode = $authorization;
+        }
+
+        if ($subscription = $payboxPayload['subscription'] ?: false) {
```
ottaviano: same

src/Entity/Transaction.php
ottaviano: why do you move this if it is not used?
can you check and remove unused code pls?

src/Repository/TransactionRepository.php
```diff
@@ -0,0 +1,46 @@
...
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, Transaction::class);
+    }
+
+    public function findByPayboxTransactionId(string $transactionId): ?Transaction
+    {
+        return $this->createQueryBuilder('transaction')
+            ->where('transaction.payboxTransactionId = :transactionId')
+            ->setParameters(['transactionId' => $transactionId])
```
ottaviano: `setParameter` without `s`

src/Repository/TransactionRepository.php
```diff
@@ -0,0 +1,46 @@
...
+     * @param string $emailAddress
+     *
+     * @return Transaction[]
+     */
+    public function findAllSuccessTransactionByEmail(string $emailAddress): array
+    {
+        return $this->createQueryBuilder('transaction')
+            ->innerJoin('transaction.donation', 'donation')
+            ->andWhere('donation.emailAddress = :email')
+            ->andWhere('transaction.payboxResultCode = :resultCode')
```
ottaviano: do you have an index for this column?

templates/user/my_donation.html.twig
```diff
@@ -54,10 +54,11 @@
                 </tr>
             </thead>
             <tbody>
-                {% for donation in donations %}
+                {% for transaction in transactions_success %}
+                    {% set donation = transaction.donation %}
```
ottaviano: can you check with profiler of SF how many db queries you execute when you are in this view?
I think you need improve your dql for select donation too.

app/migrations/Version20180502142950.php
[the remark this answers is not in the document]
B-Galati: Ok fair enough, thank you! 
I missed some stuff :D

src/Entity/Donation.php
B-Galati: I am sorry but the semantic is wrong somewhere.

From a human point of view: neither `finished` nor  `subscription_in_progress`  implies success.

Any thoughts @ottaviano @Remg @thomasbisignani @illusionOfParadise ?

src/Entity/Transaction.php
B-Galati: I think he does not want to set `$this->payboxTransactionId` with `false`.

What about `$this->payboxTransactionId = $payboxPayload['transaction'] ?: null;`

Don't use `??` if `$payboxPayload['transaction']` is a mandatory field.

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.finished = 1',
+            ['status' => Donation::STATUS_FINISHED]
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.finished = 0',
+            ['status' => Donation::STATUS_WAITING_CONFIRMATION]
+        );
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.duration != 0',
```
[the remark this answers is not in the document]
cacahouwete: I don't think it's possible, it handle 2 different status :s , but if you have a better ;)

app/migrations/Version20180502142950.php
```diff
@@ -0,0 +1,88 @@
...
+                          \' \',
+                          SUBSTR(paybox_payload->"$.time", 2, 8)
+                        ),
+                        \'%d,%m,%Y %H:%i:%s\'
+                      ), NULL),
+                      IF (paybox_payload->"$.transaction" != \'0\', paybox_payload->"$.transaction", NULL),
+                      created_at
+                      FROM donations)'
+        );
+        $this->addSql('ALTER TABLE donations MODIFY status VARCHAR(25) NOT NULL, MODIFY updated_at DATETIME NOT NULL');
```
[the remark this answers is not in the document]
cacahouwete: MODIFY change the column without renaming, I just want modify to put column NOT NULL.

src/Entity/Donation.php
cacahouwete: already declared in construct

app/migrations/Version20180502142950.php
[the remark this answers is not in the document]
ottaviano: oh sorry, my bad, I did not see the difference `:)

app/migrations/Version20180502142950.php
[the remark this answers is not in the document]
ottaviano: 👍

src/Entity/Transaction.php
cacahouwete: It is the date time when the transaction have been done in paybox. It could be done yesterday but we receive the notification only now.

ottaviano: @cacahouwete p48 array N2 
http://www1.paybox.com/wp-content/uploads/2017/08/ManuelIntegrationVerifone_PayboxSystem_V8.0_EN.pdf 

you can do directly: 

```
\DateTimeImmutable::createFromFormat('DDMMYYYYH:i:s', $date.$time) // without substr & co
```

templates/user/my_donation.html.twig
cacahouwete: forgot addSelect :D

src/Entity/Transaction.php
cacahouwete: When I do it like that I have :
```php
Return value of AppBundle\Entity\Transaction::getPayboxDateTime() must be an instance of DateTimeImmutable or null, boolean returned
```

Caused by the fact createFromFormat return false :s

cacahouwete: create format need a separation between each element :s

cacahouwete: But i can do dmYH:i:s

src/Entity/Transaction.php
```diff
@@ -0,0 +1,159 @@
...
+        $this->payboxResultCode = $payboxPayload['result'];
+        $this->payboxAuthorizationCode = $payboxPayload['authorization'] ?: null;
+        $this->payboxSubscriptionId = $payboxPayload['subscription'] ?: null;
+        $this->payboxTransactionId = $payboxPayload['transaction'] ?: null;
+
+        if (isset($payboxPayload['date'], $payboxPayload['time'])) {
+            [
+                'date' => $date,
+                'time' => $time,
+            ] = $payboxPayload;
```
ottaviano: IMO this step is useless, you can read date and time indexes from array directly

src/Controller/EnMarche/DonationController.php
cacahouwete: I did it but the team make the decision it was overkill

src/Entity/Donation.php
cacahouwete: Enum class was overkill, no methods in enum was used.

src/Controller/EnMarche/DonationController.php
B-Galati: Let me clarify.

This optimization is overkill by itself. Set the service public just for that does not worth the shot IMHO. Instantiating one more service on one route every single time is a tiny tiny tiny thing.

`ServiceSubscriberInterface` was used by @cacahouwete but it's and overhead for all the controller while it's only needed here and it makes the code less readable for no good reason.

src/Controller/EnMarche/DonationController.php
```diff
@@ -136,7 +145,7 @@ public function resultAction(Request $request, Donation $donation, MembershipReg
         $membershipRegistrationProcess->terminate();
 
         return $this->render('donation/result.html.twig', [
-            'successful' => $donation->isSuccessful(),
+            'successful' => 'effectue' === $status,
```
thomasbisignani: Can you use a const for this status ?

src/Donation/TransactionSubscriber.php
```diff
@@ -6,36 +6,43 @@
...
     private $manager;
     private $mailer;
     private $requestStack;
 
-    public function __construct(ObjectManager $manager, MailerService $mailer, RequestStack $requestStack)
+    public function __construct(ObjectManager $manager, MailerService $mailerTransactional, RequestStack $requestStack)
     {
         $this->manager = $manager;
-        $this->mailer = $mailer;
+        $this->mailer = $mailerTransactional;
```
B-Galati: ~~`$mailerTransactional`~~ -> `$mailer`

tests/Controller/EnMarche/DonationControllerTest.php
```diff
@@ -302,10 +318,16 @@ public function testRetryProcess(int $duration)
...
+        $this->assertTrue($donation->hasError());
+
+        /** @var Transaction[] $transactions */
+        $transactions = $this->transactionRepository->findBy(['donation' => $donation]);
+        $this->assertCount(1, $transactions);
+        $transaction = $transactions[0];
+        $this->assertSame('00001', $transaction->getPayboxResultCode());
+        $this->assertNull($transaction->getPayboxAuthorizationCode());
+        $this->assertNull($transaction->getPayboxTransactionId());
+        $this->assertNull($transaction->getPayboxTransactionId());
```
B-Galati: Same assertion twice

src/Donation/TransactionCallbackHandler.php
B-Galati: @cacahouwete you forgot this one

src/Donation/TransactionSubscriber.php
cacahouwete: I bound `$mailerTransactional` to the service `app.mailer.transactional`, he is used in many other services, if in long term we want use auto config/wire , I think it's a good idea to use bind arguments no ?

app/config/services/services.xml
```diff
@@ -4,7 +4,10 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <defaults autowire="true" autoconfigure="true" public="false" />
+        <defaults autowire="true" autoconfigure="true" public="false" >
+            <bind key="$donationRequestUtilsLocator" type="service" id="app.donation_request.utils.locator" />
+            <bind key="$environment" >%kernel.environment%</bind>
```
illusionOfParadise: An extra space. Check other extra spaces.

PR #2399, opened 2018-05-14
Add new DateRange validator and use it for Event & Action forms

ottaviano:
https://app.clubhouse.io/enmarche/story/4097/etq-cr%C3%A9ateur-d-%C3%A9v%C3%A9nement-je-dois-avoir-mon-calendrier-qui-ajouter-la-date-de-fin-%C3%A0-2h-du-m%C3%AAme-jour-et-je-dois-%C3%AAtre-limit%C3%A9

image

### DIFF
--- a/src/Form/BaseEventCommandType.php
+++ b/src/Form/BaseEventCommandType.php
@@ -14,6 +14,8 @@ class BaseEventCommandType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $now = $this->createBeginDate($options['minutes']);
+
         $builder
             ->add('name', TextType::class, [
                 'filter_emojis' => true,
@@ -28,10 +30,12 @@ class BaseEventCommandType extends AbstractType
             ])
             ->add('address', AddressType::class)
             ->add('beginAt', DateTimeType::class, [
+                'data' => $now,
                 'years' => $options['years'],
                 'minutes' => $options['minutes'],
             ])
             ->add('finishAt', DateTimeType::class, [
+                'data' => (clone $now)->modify('+2 hours'),
                 'years' => $options['years'],
                 'minutes' => $options['minutes'],
             ])
@@ -57,5 +61,36 @@ class BaseEventCommandType extends AbstractType
                 'event_category_class' => EventCategory::class,
             ])
         ;
+    }
+
+    /**
+     * Creates the begin date in the future for the next minute step.
+     * ex 1:
+     *     if current time is 16h27m, the returned time will be 16h30m with these minutes steps: [0, 15, 30, 45]
+     * ex 2:
+     *     with time 16h52m it returns 17h00
+     */
+    private function createBeginDate(array $minutes): \DateTime
+    {
+        $now = new \DateTime();
+        $nowMinute = $now->format('i');
+        $step = null;
+
+        foreach ($minutes as $step) {
+            if ($nowMinute <= $step) {
+                break;
+            }
+        }
+
+        if ($step < $nowMinute) {
+            $now
+                ->modify('next hour')
+                ->modify(sprintf('-%d minutes', $nowMinute))
+            ;
+        } else {
+            $now->modify(sprintf('+%d minutes', $step - $nowMinute));
+        }
+
+        return $now;
     }
 }
--- a/src/Validator/DateRange.php
+++ b/src/Validator/DateRange.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace AppBundle\Validator;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ * @Target({"CLASS", "ANNOTATION"})
+ */
+class DateRange extends Constraint
+{
+    public $startDateField;
+    public $endDateField;
+    public $interval;
+    public $message = 'common.date_range.invalid_interval';
+
+    public function getRequiredOptions()
+    {
+        return ['startDateField', 'endDateField', 'interval'];
+    }
+
+    public function getTargets()
+    {
+        return self::CLASS_CONSTRAINT;
+    }
+}
--- a/src/Validator/DateRangeValidator.php
+++ b/src/Validator/DateRangeValidator.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace AppBundle\Validator;
+
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+class DateRangeValidator extends ConstraintValidator
+{
+    private $propertyAccessor;
+
+    public function __construct(PropertyAccessorInterface $propertyAccessor)
+    {
+        $this->propertyAccessor = $propertyAccessor;
+    }
+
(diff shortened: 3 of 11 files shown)


src/Validator/DateRangeValidator.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace AppBundle\Validator;
+
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+class DateRangeValidator extends ConstraintValidator
```
B-Galati: This class would deserve some tests no ?

translations/validators.fr.yml
```diff
@@ -78,6 +78,7 @@ committee.must_accept_contacting_terms: Vous devez accepter d'être contacté(e)
 
 committee.event.invalid_category: Le type d'événement sélectionné n'est pas valide.
 committee.event.invalid_date_range: La date de fin de l'événement doit être postérieure à la date de début.
+committee.event.invalid_finish_date: La date de fin de votre événement ne peut pas dépasser le {{ limit }}.
```
B-Galati: I guess we could have the same message ID for committees and CP. Not a big deal.

src/Validator/DateRangeValidator.php
```diff
@@ -0,0 +1,59 @@
...
+        $interval = \DateInterval::createFromDateString($constraint->interval);
+
+        $startDate = clone $this->propertyAccessor->getValue($value, $constraint->startDateField);
+        $endDate = $this->propertyAccessor->getValue($value, $constraint->endDateField);
+
+        if (!$startDate instanceof \DateTime) {
+            throw new ConstraintDefinitionException('The start date field should be of type DateTime');
+        }
+
+        if (!$endDate instanceof \DateTime) {
```
B-Galati: Testing against `\DateTimeInterface` would be more flexible.

translations/validators.fr.yml
```diff
@@ -97,6 +98,8 @@ citizen_project.description.max_length: Votre texte de description est trop long
 citizen_project.message.min_length: Le message doit contenir au moins {{ limit }} caractères.
 citizen_project.emoji.max_length: Vous devez coller un emoji.
 
+citizen_project.action.invalid_finish_date: La date de fin de votre action ne peut pas dépasser le {{ limit }}.
```
illusionOfParadise: To be consistent with the wording in the project: 
```
#
# CitizenAction
#
citizen_action.invalid_finish_date: La date de fin de votre action citoyenne ne peut pas dépasser le {{ limit }}.
```

templates/citizen_action_manager/_layout.html.twig
```diff
@@ -54,6 +54,7 @@
                     {{ form_errors(citizen_action_form.finishAt) }}
                     {{ form_label(citizen_action_form.finishAt, 'Fin', { label_attr: { class: 'text--bold' } }) }}
                     {{ form_widget(citizen_action_form.finishAt, { attr: { class: 'citizen_action__finish_at l__row' } }) }}
+                    <div class="form__info">Votre action ne peut pas durer plus de 3 jours.</div>
```
illusionOfParadise: To be consistent with the wording in the project: `action` -> `action citoyenne`.

src/Validator/DateRange.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @Annotation
+ * @Target({"CLASS", "ANNOTATION"})
+ */
+class DateRange extends Constraint
+{
+    public $startDateField;
+    public $endDateField;
+    public $interval;
+    public $message = 'The end date can not exceed {{ limit }}';
```
thomasbisignani: Can you use translation here ?

thomasbisignani: Yes for the built-in validators, but IMHO it's more consistent and easly to use translations for our validators messages, for example : https://github.com/EnMarche/en-marche.fr/blob/master/src/Validator/GeocodableAddress.php#L17

B-Galati: I check them all and it's not all of them which use translation string so I let you deal with @ottaviano :p
But yea you made a point, it would be easier to use with a default message which matches a translation.

src/Form/BaseEventCommandType.php
```diff
@@ -58,4 +62,31 @@ public function configureOptions(OptionsResolver $resolver)
...
     }
+
+    /**
+     * Creates the begin date in the future for the next minute step.
+     */
+    private function createBeginDate(array $minutes): \DateTime
+    {
+        $now = new \DateTime();
+        $nowMinute = $now->format('i');
+        $step = reset($minutes);
```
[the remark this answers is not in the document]
ottaviano: to initialise $step variable if $minutes array is empty

src/Validator/DateRangeValidator.php
ottaviano: Test added 👍

translations/validators.fr.yml
ottaviano: the message changes `votre événement` and `votre action citoyenne`

src/Validator/DateRangeValidator.php
ottaviano: ok for the `$endDate` but not for `$startDate` because I would call `->add()` method and DateTimeInterface don't have this method.

translations/validators.fr.yml
ottaviano: @illusionOfParadise check `committee.event.*` keys

src/Form/BaseEventCommandType.php
```diff
@@ -58,4 +62,31 @@ public function configureOptions(OptionsResolver $resolver)
             ])
         ;
     }
+
+    /**
+     * Creates the begin date in the future for the next minute step.
```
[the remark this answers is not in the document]
B-Galati: Thank you !

templates/citizen_action_manager/_layout.html.twig
ottaviano: done 👍 tnx

src/Validator/DateRange.php
ottaviano: not big deal, replaced by key 😗 @thomasbisignani

src/Form/BaseEventCommandType.php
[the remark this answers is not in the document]
B-Galati: In that case, `$step` should be initialized to `null` and conditions below should be modified with something like :

```php
      if ($step === null) {
           return $now;
      }

      if ($step < $nowMinute) {
            return $now
                ->modify('next hour')
                ->modify(sprintf('-%d minutes', $nowMinute))
            ;
     }

    return $now->modify(sprintf('+%d minutes', $step - $nowMinute));
```

WDYT ?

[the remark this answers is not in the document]
ottaviano: I initialise `$step` with `null` and if `$minutes` is empty then the returned time will be the next hour because it valides `$step < $nowMinute` condition.

ex: if current time is 16h05 and `$minutes` is empty, and `$step` is null, then returned time will be 17h00.
ok for you?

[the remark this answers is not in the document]
B-Galati: Yes yes yes :1st_place_medal:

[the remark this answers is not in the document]
ottaviano: in this case it will be the same time, no pb :)

PR #2400, opened 2018-05-14
[ch3092] Instant validation: fix some js validation

cacahouwete:
https://app.clubhouse.io/enmarche/story/3092/int%C3%A9grer-l-instant-field-validation-sur-les-formulaires-de-don-et-d-adh%C3%A9sion


front/validator/formValidator.js
```diff
@@ -35,9 +38,19 @@ export default (formType, form) => {
 
         const formData = new FormData(form);
         const emptyFieldNames = [];
+        let id;
```
ottaviano: you can do this in line N 47 directly

PR #2406, opened 2018-05-15
Fix slow Test with time using

ottaviano:
```
1) Tests\AppBundle\OAuth\PersistentTokenFactoryTest::testCreateRefreshToken
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'1526423610'
+'1526423609'

/app/tests/OAuth/PersistentTokenFactoryTest.php:249

```


B-Galati:
It does not fix anything IMO, as phpunit bridge is mocking `time`, `microtime`, `sleep` and `usleep`. We should switch to Carbon instead, like we did on https://github.com/EnMarche/en-marche.fr/pull/2383

PR #2416, opened 2018-05-17
Fix display donation, we want display only donations finish in success.

cacahouwete:


src/Repository/DonationRepository.php
```diff
@@ -26,12 +26,19 @@ public function findOneByUuid(string $uuid): ?Donation
...
+     * @return Donation[]
+     */
     public function findByEmailAddressOrderedByDonatedAt(string $email): array
     {
-        return $this->findBy(
-            ['emailAddress' => $email],
-            ['donatedAt' => 'DESC']
-        );
+        return $this->createQueryBuilder('donation')
+            ->andWhere('donation.emailAddress = :email')
```
ottaviano: little optim: first where should be `->where(...)`

PR #2420, opened 2018-05-18
Added adhesion CTA on donation result page

thomasbisignani:


src/Repository/AdherentRepository.php
```diff
@@ -64,6 +64,11 @@ public function findOneByEmail(string $email): ?Adherent
         return $this->findOneBy(['emailAddress' => $email]);
     }
 
+    public function isAdherent(string $email): bool
+    {
+        return (bool) $this->findOneBy(['emailAddress' => $email]);
```
ottaviano: 1) what about use the COUNT with QueryBuilder to avoid hydrate the Adherent object just for cast in boolean value?

or maybe do `return 0 < count($this->findAdherentsUuidByEmailAddress($email);` ?

2) do you need to check `adherents.adherent` value too?

thomasbisignani: Nice catch, thx ! Alright with the COUNT :+1:

src/Controller/EnMarche/DonationController.php
```diff
@@ -140,7 +140,8 @@ public function resultAction(Request $request, Donation $donation, MembershipReg
             'error_code' => $request->query->get('code'),
             'donation' => $donation,
             'retry_url' => $retryUrl,
-            'is_new_adherent' => $request->query->get('is_new_adherent'),
+            'is_registration' => $request->query->get('is_registration'),
+            'is_adherent' => $request->query->get('is_adherent'),
```
B-Galati: Why do we need to get the information from the query string while we could have it directly with `$donation` object and a call to the repository? Like https://github.com/EnMarche/en-marche.fr/pull/2420/files#diff-ff29cb5887bec45ec82e598374291962R142

PR #2421, opened 2018-05-18
Add referent tags to unregistration process

B-Galati:


src/Entity/Unregistration.php
```diff
@@ -96,13 +98,21 @@ class Unregistration
...
+
     public function __construct(
         UuidInterface $uuid,
         array $reasons,
         ?string $comment,
         \DateTime $registeredAt,
         string $postalCode = null,
-        bool $isAdherent = false
+        bool $isAdherent = false,
+        array $referentTags
```
ottaviano: required arg should not be after the optionals

PR #2422, opened 2018-05-18
Add partial organizer name to Event serialization

Remg:


B-Galati:
What about using the UUID of the organizer ?

Remg:
I thought about this, but we are exposing events & committees and these data are public.
The user endpoint is private and a User uuid wouldn't mean anything for these endpoint.
Let's say the functional logic that anonymise the user is (already) done by EM and not API.

PR #2424, opened 2018-05-18
fix wording with link

cacahouwete:


front/components/AmountChooser.js
```diff
@@ -96,7 +96,8 @@ export default class AmountChooser extends React.Component {
                     </label>
                 </div>
                 <div className="amount-chooser__help">
-                    Vous aurez la possibilité d'annuler votre don mensuel à tout moment depuis votre espace adhérent.
+                    Vous aurez la possibilité d'annuler votre don mensuel à tout moment depuis votre espace adhérent
+                    ou en <a href="/contact">nous contactant directement</a>.
```
Remg: Url should not be hardcoded.

It should be given as a property in the dom:
https://github.com/EnMarche/en-marche.fr/blob/master/templates/donation/index.html.twig#L7

PR #2430, opened 2018-05-21
Add count events for current month to endpoint

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/3646/3e-partie-section-evenements-box-kpi-nb-events

### DIFF
--- a/src/Repository/EventRepository.php
+++ b/src/Repository/EventRepository.php
@@ -450,7 +450,7 @@ SQL;
     public function queryCountByMonth(Adherent $referent, int $months = 5): QueryBuilder
     {
         return $this->createQueryBuilder('event')
-            ->select('COUNT(event) AS count, YEAR_MONTH(event.beginAt) AS yearmonth')
+            ->select('COUNT(DISTINCT event.id) AS count, YEAR_MONTH(event.beginAt) AS yearmonth')
             ->innerJoin('event.referentTags', 'tag')
             ->where('tag IN (:tags)')
             ->setParameter('tags', $referent->getManagedArea()->getTags())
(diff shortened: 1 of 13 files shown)


src/Repository/EventRepository.php
```diff
@@ -512,6 +512,20 @@ public function countReferentEventsInReferentManagedArea(Adherent $referent): ar
         return $this->aggregateCountByMonth($result, BaseEvent::REFERENT_EVENT_TYPE.'s');
     }
 
+    public function countTotalEventsInReferentManagedAreaForCurrentMonth(Adherent $referent): int
+    {
+        $this->checkReferent($referent);
+
+        $query = $this->queryCountByMonth($referent, 1);
```
ottaviano: with this it will be `count` of the previous month + current month, and not only of current month, no?

src/Repository/EventRepository.php
```diff
@@ -512,6 +512,20 @@ public function countReferentEventsInReferentManagedArea(Adherent $referent): ar
...
+
+        $result = $query
+            ->getQuery()
+            ->getSingleResult()
+        ;
+
+        return (int) $result['count'];
```
ottaviano: `return (int) $....->getSingleScalarResult();` ?

illusionOfParadise: It will not work, because there are more than one column `count`.

src/Repository/EventRepository.php
illusionOfParadise: Surely! Thanks !
Fixed.

ottaviano: Sure :)?
https://3v4l.org/Hmi82

PR #2432, opened 2018-05-21
[ch3145] Delete all important in amp css

cacahouwete:
https://app.clubhouse.io/enmarche/story/3145/supprimer-l-attribut-important-des-pages-amp

### DIFF
--- a/front/style/layout/_buttons.scss
+++ b/front/style/layout/_buttons.scss
@@ -204,9 +204,9 @@
 }
 
 @mixin btn--as-link {
-    background: none!important;
+    background: none;
     border: none;
-    padding: 0!important;
+    padding: 0;
     border-bottom: 1px solid transparent;
 
     &:hover {
@@ -233,10 +233,10 @@ input {
 
 .btn--as-link {
     &--error {
-        color: $red!important;
+        color: $red;
         &:hover {
-            border-color: $red!important;;
-            color: $red!important;;
+            border-color: $red;
+            color: $red;
         }
     }
 }
@@ -259,7 +259,7 @@ button.btn {
         padding: 20px 25px;
 
         @media(max-width: $mobile) {
-          @include text--default;
+            @include text--default;
         }
     }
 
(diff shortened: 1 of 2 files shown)


front/style/app.scss
```diff
@@ -10,14 +10,17 @@
 
 @import "layout/base";
 @import "layout/banners";
-@import "layout/buttons";
+@import "layout/buttons-site";
```
marineb: I would keep it `buttons` here.

front/style/layout/_buttons-site.scss
```diff
@@ -0,0 +1,23 @@
+@mixin btn--as-link {
+    background: none!important;
```
marineb: Did you try just removing the `important!` tags in this file instead of creating a new file just for amp buttons? What happened when you did that?

cacahouwete: I was thinking that buttons I wouldn't break something because there are used every where :s

marineb: Cool, let's just remove the `important` tags then :)

front/style/layout/_buttons.scss
```diff
@@ -233,10 +233,10 @@ input {
 
 .btn--as-link {
     &--error {
-        color: $red!important;
+        color: $red;
         &:hover {
-            border-color: $red!important;;
-            color: $red!important;;
+            border-color: $red;;
```
thomasbisignani: Why double `;` ? I think that was a mistake, can you remove this ?

front/style/layout/_buttons.scss
```diff
@@ -233,10 +233,10 @@ input {
...
+            color: $red;;
```
thomasbisignani: Same here

front/style/layout/_buttons.scss
cacahouwete: oO, I just revert ^^, it was like that before start this issue, I will fix it anyway

PR #2435, opened 2018-05-21
Add filter by adherent role

ottaviano:
https://app.clubhouse.io/enmarche/story/2473/admin-adh%C3%A9rents-je-veux-pouvoir-filtrer-tous-les-r%C3%B4les

### DIFF
--- a/src/Admin/AdherentAdmin.php
+++ b/src/Admin/AdherentAdmin.php
@@ -2,9 +2,11 @@
 
 namespace AppBundle\Admin;
 
+use AppBundle\Adherent\AdherentRoleEnum;
 use AppBundle\Entity\Adherent;
 use AppBundle\Entity\BoardMember\BoardMember;
 use AppBundle\Entity\BoardMember\Role;
+use AppBundle\Entity\CitizenProject;
 use AppBundle\Entity\CommitteeMembership;
 use AppBundle\Form\ActivityPositionType;
 use AppBundle\Form\Admin\CoordinatorManagedAreaType;
@@ -17,6 +19,8 @@ use AppBundle\Intl\UnitedNationsBundle;
 use AppBundle\Membership\AdherentEmailSubscription;
 use AppBundle\Membership\UserEvent;
 use AppBundle\Membership\UserEvents;
+use Doctrine\ORM\Query\Expr;
+use Doctrine\ORM\QueryBuilder;
 use Misd\PhoneNumberBundle\Form\Type\PhoneNumberType;
 use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\ListMapper;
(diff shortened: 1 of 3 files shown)


ottaviano:
\o/ refacto forever!

src/Admin/AdherentAdmin.php
```diff
@@ -409,6 +370,83 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                        return false;
+                    }
+
+                    $where = new Expr\Orx();
+
+                    /** @var QueryBuilder $qb */
+
+                    // Referent
+                    if (\in_array(AdherentRoleEnum::REFERENT, $value['value'], true)) {
+                        $where->add(sprintf('%s.managedArea IS NOT NULL', $alias));
```
illusionOfParadise: To correspond to role definition in the `Adherent` class https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/Adherent.php#L774 we should check that tags exist.
Check other role definitions in this class.

translations/messages.fr.yml
```diff
@@ -27,6 +28,14 @@ adherent.activity_position.retired: Retraité
 adherent.activity_position.employed: En activité
 adherent.activity_position.unemployed: En recherche d'emploi
 
+referent: référents
```
illusionOfParadise: Should be in the singular `référent`.

src/Admin/AdherentAdmin.php
Remg: yes. in order to be a Referent,  `managedArea` should not be null (as stated here) AND `mangedArea.tags` should not be empty

ottaviano: not need with this https://github.com/EnMarche/en-marche.fr/blame/8eee031dd79af478fabeeea4bdddeaf367976d89/src/Form/EventListener/ReferentManagedAreaListener.php#L29

managed_area_id become NULL if tags is empty

translations/messages.fr.yml
ottaviano: done 👍 thanks

src/Admin/AdherentAdmin.php
```diff
@@ -409,6 +370,83 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                    if ($committeeRoles = array_intersect([AdherentRoleEnum::COMMITTEE_SUPERVISOR, AdherentRoleEnum::COMMITTEE_HOST], $value['value'])) {
+                        $qb->leftJoin(sprintf('%s.memberships', $alias), 'ms');
+                        $where->add('ms.privilege IN (:privileges)');
+                        if (\in_array(AdherentRoleEnum::COMMITTEE_SUPERVISOR, $committeeRoles, true)) {
+                            $privileges[] = CommitteeMembership::COMMITTEE_SUPERVISOR;
+                        }
+                        if (\in_array(AdherentRoleEnum::COMMITTEE_HOST, $committeeRoles, true)) {
+                            $privileges[] = CommitteeMembership::COMMITTEE_HOST;
+                        }
+                        $qb->setParameter('privileges', $privileges);
```
B-Galati: It looks like `$committeeRoles` can be used directly instead of building `$privileges`.

Perhaps I am missing something.

ottaviano: @B-Galati, not really https://github.com/EnMarche/en-marche.fr/blob/c2d028522fe39699b9bca6001064a732fe9a7c18/src/Entity/CommitteeMembership.php#L34

I prefix by `committee_` for what if we need to add `citizen_project_host` for example...

B-Galati: Yes I saw what I missed now :+1: 
The constant was not the same, thank you.

ottaviano:
@B-Galati, for `Adherent` role, Guillaume confirms to me that no, not need it.

PR #2443, opened 2018-05-22
Add Committee relation to CommitteeMembership

ottaviano:


src/RepublicanSilence/AdherentZone/CommitteeHostZoneRetriever.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace AppBundle\RepublicanSilence\AdherentZone;
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\CommitteeMembership;
+use Symfony\Component\HttpFoundation\Request;
+
+class CommitteeHostZoneRetriever implements AdherentZoneRetrieverInterface
```
B-Galati: Retriever is a type of dog in English like in French.
IMO this is, either a `Manager` or a `Handler` or a `QueryHandler`.

src/RepublicanSilence/AdherentZone/CommitteeHostZoneRetriever.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function getAdherentZone(Adherent $adherent, Request $request): array
```
B-Galati: It would be more flexible and easier to test if we only pass the slug value directly. instead of the whole Request object.

src/Entity/ReferentManagedArea.php
```diff
@@ -107,4 +107,9 @@ public function setMarkerLongitude(?string $markerLongitude): void
 
         $this->markerLongitude = $markerLongitude;
     }
+
+    public function getCodes(): array
```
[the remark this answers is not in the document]
ottaviano: removed, wrong branch for this code

src/Entity/CommitteeMembership.php
```diff
@@ -44,20 +44,20 @@ class CommitteeMembership
     use EntityIdentityTrait;
 
     /**
-     * The committee UUID.
+     * @var Adherent
      *
-     * @var UuidInterface
-     *
-     * @ORM\Column(type="uuid")
+     * @ORM\ManyToOne(targetEntity="Adherent", inversedBy="memberships")
```
cacahouwete: Is it absolutely needed to have bidirectional link ?

ottaviano: I think yes, it existed before and it's used in `Adherent` 
image

src/Repository/CommitteeMembershipRepository.php
```diff
@@ -361,61 +303,76 @@ function (CommitteeMembership $membership) {
...
+    }
 
-        $query = $qb
-            ->select('cm.committeeUuid')
-            ->leftJoin('cm.adherent', 'a')
-            ->where('LOWER(a.emailAddress) LIKE :emailAddress')
-            ->andWhere($qb->expr()->in('cm.privilege', CommitteeMembership::getHostPrivileges()))
-            ->setParameter('emailAddress', '%'.strtolower($emailAddress).'%')
-            ->getQuery()
+    public function findCommitteesUuid(array $criteria): array
```
B-Galati: IMHO it's better to have some private methods that can build a base line querybuilder than trying to build something generic with an array of `$criteria`. 

At the beginning that sort of method are small and simple and they end big and cumbersome for no good reasons than trying to make generic code.

thomasbisignani:
:medal_sports:

PR #2444, opened 2018-05-22
Add CitizenProject relation to CitizenProjectMembership object

ottaviano:


src/Entity/CitizenProjectMembership.php
```diff
@@ -212,7 +212,8 @@ public function getSubscriptionDate(): \DateTimeImmutable
 
     public function matches(Adherent $adherent, CitizenProject $citizenProject): bool
     {
-        return $adherent->equals($this->getAdherent()) && $this->citizenProjectUuid->equals($citizenProject->getUuid());
+        return $adherent->equals($this->getAdherent())
+            && $this->citizenProject->getUuid()->equals($citizenProject->getUuid());
```
cacahouwete: It could be better to have :
```php
        return $adherent->equals($this->getAdherent()) &&
            $this->citizenProject->getUuid()->equals($citizenProject->getUuid());
```
It's just to know the convention int the project about that

ottaviano: we don't have any convention about that, it's not in PSR.
why it's better? 😃

B-Galati: I do prefer the version of this PR with the leading operator

ottaviano:
I rebased this branch and push force :D

src/Entity/CitizenProjectMembership.php
Remg: :popcorn:

B-Galati: This is not a valid comparison IMO. On one hand, the trailing coma does not bring any useful information and it can be on every single line if it is at the end of each line. You cannot if the coma is at beginning of each line

On the other hand, a logical operator such as `&&` or `||` do have a lot of meaning so that if you place them at the beginning of each line they can be seen very easily and it becomes obvious how they are used.

src/Repository/CitizenProjectMembershipRepository.php
```diff
@@ -116,73 +101,65 @@ public function countAdministratorMembers(string $citizenProjectUuid): int
...
             $privileges[] = CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR;
         }
 
-        return $this->findPriviledgedMembers($citizenProjectUuid, $privileges);
+        return $this->findPrivilegedMembers($citizenProject, $privileges);
     }
 
     /**
-     * Returns the list of all priviledged memberships of a citizen project.
+     * Returns the list of all privileged memberships of a citizen project.
```
thomasbisignani: :medal_sports: :man_cartwheeling:

src/Entity/CitizenProjectMembership.php
ottaviano: @cacahouwete, do you prefer the first or second method? 
```php

private function supportUser1($user): bool
{
    return $user instanceof Adherent
        && (
            $user->isadsfasdffsdasdfasdfasdfasdfasdfasdfasdfasdfaHost()
            || $user->isSupervisor131231313123123()
            || $user->isasdfasdf()
            || $user->isCiasdfasdatizenProjectAdminiasdfasdfasdfasdfasdfastrator()
        );
}

private function supportUser2($user): bool
{
    return $user instanceof Adherent &&
        (
            $user->isadsfasdffsdasdfasdfasdfasdfasdfasdfasdfasdfaHost() ||
            $user->isSupervisor131231313123123() ||
            $user->isasdfasdf() ||
            $user->isCiasdfasdatizenProjectAdminiasdfasdfasdfasdfasdfastrator()
        );
}
```

PR #2449, opened 2018-05-22
CitizenProject & CitizenAction synchronization with API

Remg:


app/config/services/services.xml
```diff
@@ -171,5 +171,32 @@
 
         <!-- History -->
         <service id="AppBundle\History\EmailSubscriptionHistoryHandler" />
+
+        <!-- Sync with API -->
+
+        <!-- User subscriber -->
+        <service id="AppBundle\Membership\EventListener\UserSubscriber">
+            <argument key="$producer" type="service" id="old_sound_rabbit_mq.sync_api_producer"/>
```
cacahouwete: It's a question, why not using bind to bind service old_sound_rabbit_mq.sync_api_producer to param name $producer ?

B-Galati: This is very fragile/error prone for such a service to use bindings. We could quickly make some mistakes without noticing anything.
Without bind It's explicit, easy to read, easy to understand.

B-Galati: Last thing, we are using `$producer` variable in lots of places but they point to different classes most of the time.

thomasbisignani: I'm agree with @B-Galati, it's too overkill for this case.

features/sync_with_api.feature
```diff
@@ -1,3 +1,4 @@
+@bonsoir
```
illusionOfParadise: Bonsoir? =P

PR #2462, opened 2018-05-25
Add endpoint for filtered events count by month + search bar with autocomplete

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/3612/2e-partie-section-comit%C3%A9s-scroll-down-double-graphique-choisir-un-comit%C3%A9-1

### DIFF
--- a/src/Controller/Api/EventsController.php
+++ b/src/Controller/Api/EventsController.php
@@ -2,20 +2,48 @@
 
 namespace AppBundle\Controller\Api;
 
+use AppBundle\Entity\Committee;
+use AppBundle\Repository\EventRepository;
+use AppBundle\Statistics\StatisticsParametersFilter;
+use Psr\Log\InvalidArgumentException;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
+/**
+ * @Route("/events")
+ */
 class EventsController extends Controller
 {
     /**
-     * @Route("/events", name="api_committees_events")
+     * @Route("", name="api_committees_events")
      * @Method("GET")
      */
-    public function getUpcomingCommitteesEventsAction(Request $request)
+    public function getUpcomingCommitteesEventsAction(Request $request): Response
     {
         return new JsonResponse($this->get('app.api.event_provider')->getUpcomingEvents($request->query->getInt('type')));
+    }
+
+    /**
+     * @Route("/count-by-month", name="app_committee_events_count_by_month")
+     * @Method("GET")
+     * @Security("is_granted('ROLE_REFERENT')")
+     */
+    public function eventsCountInReferentManagedAreaAction(Request $request, EventRepository $eventRepository): Response
+    {
+        $referent = $this->getUser();
+        try {
+            $filter = StatisticsParametersFilter::createFromRequest($request, $this->getDoctrine()->getRepository(Committee::class));
+        } catch (InvalidArgumentException $e) {
+            throw new BadRequestHttpException($e->getMessage());
+        }
+        $events = $eventRepository->countCommitteeEventsInReferentManagedArea($referent, $filter);
+
+        return new JsonResponse($events);
     }
 }
--- a/src/Controller/Api/ReferentController.php
+++ b/src/Controller/Api/ReferentController.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace AppBundle\Controller\Api;
+
+use AppBundle\Repository\CommitteeRepository;
+use AppBundle\Repository\EventRepository;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @Route("/referent")
+ * @Security("is_granted('ROLE_REFERENT')")
+ */
+class ReferentController extends Controller
+{
+    private const AUTOCOMPLETE_TYPE_CITY = 'city';
+    private const AUTOCOMPLETE_TYPE_COMMITTEE = 'committee';
+    private const AUTOCOMPLETE_TYPE_COUNTRY = 'country';
+
+    private const AUTOCOMPLETE_TYPES = [
+        self::AUTOCOMPLETE_TYPE_CITY,
+        self::AUTOCOMPLETE_TYPE_COMMITTEE,
+        self::AUTOCOMPLETE_TYPE_COUNTRY,
+    ];
+
+    /**
+     * @Route("/search/autocomplete", name="api_referent_space_search_autocomplete")
+     * @Method("GET")
+     */
+    public function searchAutocompleteAction(Request $request, CommitteeRepository $committeeRepository, EventRepository $eventRepository): Response
+    {
+        if (!($type = $request->query->get('type')) || !($value = $request->query->get('value'))) {
+            throw new BadRequestHttpException('The parameters "type" and "value" are required.');
+        }
+
+        if (!in_array($type, self::AUTOCOMPLETE_TYPES, true)) {
+            throw new BadRequestHttpException(sprintf('Invalid parameter "type" "%" given.', $type));
+        }
+        $referent = $this->getUser();
+
+        switch ($type) {
+            case self::AUTOCOMPLETE_TYPE_CITY:
+                return new JsonResponse(['cities' => array_unique(array_merge(
+                    $committeeRepository->findCitiesForReferentAutocomplete($referent, $value),
+                    $eventRepository->findCitiesForReferentAutocomplete($referent, $value)),
+                    SORT_REGULAR
+                )]);
+            case self::AUTOCOMPLETE_TYPE_COMMITTEE:
+                return new JsonResponse(['committees' => $committeeRepository->findApprovedForReferentAutocomplete($referent, $value)]);
+            case self::AUTOCOMPLETE_TYPE_COUNTRY:
+                return new JsonResponse(['countries' => $referent->getManagedArea()->getOnlyManagedCountryCodes($value)]);
+        }
+    }
+}
--- a/src/Entity/ReferentManagedArea.php
+++ b/src/Entity/ReferentManagedArea.php
@@ -80,6 +80,15 @@ class ReferentManagedArea
         $this->tags->removeElement($tag);
     }
 
+    public function getOnlyManagedCountryCodes($value): array
+    {
+        return array_values(array_filter(array_map(function (ReferentTag $tag) use ($value) {
+            if (ctype_alpha($tag->getCode()) && 0 === stripos($tag->getName(), $value)) {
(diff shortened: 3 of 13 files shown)


app/config/services/services.xml
```diff
@@ -174,5 +174,8 @@
 
         <!-- History -->
         <service id="AppBundle\History\EmailSubscriptionHistoryHandler" />
+
+        <!-- Stats -->
+        <service id="AppBundle\Statistics\StatisticsParametersFilter" public="true" />
```
B-Galati: This class should not be a service as it is something mutable. 

It's just a place where we store all filters asked by the client.

It could be instantiated like this in controller instead `StatisticsParametersFilter::fromRequest()`

src/Entity/ReferentManagedArea.php
```diff
@@ -80,6 +80,15 @@ public function removeTag(ReferentTag $tag): void
         $this->tags->removeElement($tag);
     }
 
+    public function getOnlyManagedCountryCodes($value): array
+    {
+        return array_values(array_filter(array_map(function (ReferentTag $tag) use ($value) {
+            if (ctype_alpha($tag->getCode()) && 0 === stripos($tag->getName(), $value)) {
+                return [$tag->getCode() => $tag->getName()];
```
B-Galati: Why returning a key if at the end we do `array_values` ?

src/Repository/CommitteeRepository.php
```diff
@@ -358,6 +358,80 @@ public function countApprovedForReferent(Adherent $referent): int
         ;
     }
 
+    public function findApprovedForReferentAutocomplete(Adherent $referent, $value): array
```
B-Galati: Please find a method that tell us what's the goal/intention of this method. It's really unclear what it is doing right now other that it is made for an autocomplete stuff.

B-Galati: Same comment for all new methods of this class.

src/Repository/CommitteeRepository.php
```diff
@@ -358,6 +358,80 @@ public function countApprovedForReferent(Adherent $referent): int
...
+    {
+        if (!$referent->isReferent()) {
+            throw new \InvalidArgumentException('Adherent must be a referent.');
+        }
+
+        $qb = $this->createQueryBuilder('committee')
+            ->select('committee.uuid, committee.name')
+            ->join('committee.referentTags', 'tag')
+            ->where('committee.status = :status')
+            ->andWhere('committee.name LIKE :word')
```
B-Galati: `searchedName` or something like this would be more accurate.

src/Repository/CommitteeRepository.php
```diff
@@ -358,6 +358,80 @@ public function countApprovedForReferent(Adherent $referent): int
...
+        ;
+
+        $tagsFilter = $qb->expr()->orX();
+
+        foreach ($referent->getManagedArea()->getTags() as $key => $tag) {
+            $tagsFilter->add("tag.code IN (:tag_$key)");
+            $tagsFilter->add(
+                $qb->expr()->andX(
+                    'committee.postAddress.country = \'FR\'',
+                    $qb->expr()->like('committee.postAddress.postalCode', ":tag_prefix_$key")
```
B-Galati: ~~`:tag_prefix_$key`~~ `:postal_code_$key` would be clearer

src/Repository/CommitteeRepository.php
```diff
@@ -358,6 +358,80 @@ public function countApprovedForReferent(Adherent $referent): int
...
+            ->andWhere('committee.name LIKE :word')
+            ->setParameter('word', $value.'%')
+            ->setParameter('status', Committee::APPROVED)
+            ->orderBy('committee.name')
+        ;
+
+        $tagsFilter = $qb->expr()->orX();
+
+        foreach ($referent->getManagedArea()->getTags() as $key => $tag) {
+            $tagsFilter->add("tag.code IN (:tag_$key)");
```
B-Galati: Could use `=` instead of `IN` 
And could use entity directly instead of the `$tag->getCode()`.

B-Galati: I think the tag condition can be put outside of the foreach no ?

```php
$tagsFilter
    ->add("tag IN (:tags)")
    ->setParameter('tags', $referent->getManagedArea()->getTags())
;
```

src/Entity/ReferentManagedArea.php
illusionOfParadise: It is the subarray keys and not the ones removed by `array_values`.

B-Galati: :+1: thank you !

src/Repository/CommitteeRepository.php
illusionOfParadise: It's not easy to find another appropriate name. The goal is to return the values for autocomplete, no other purpose. If I describe what method does it will give `getUuidAndNameOfApprouvedCommitteesSearchedByNameInReferentManagedArea` name, which is not practical. Also there is too little chance that it will be used elsewhere.

src/Controller/Api/EventsController.php
```diff
@@ -2,20 +2,48 @@
...
+    /**
+     * @Route("/count-by-month", name="app_committee_events_count_by_month")
+     * @Method("GET")
+     * @Security("is_granted('ROLE_REFERENT')")
+     */
+    public function eventsCountInReferentManagedAreaAction(Request $request, EventRepository $eventRepository): Response
+    {
+        $referent = $this->getUser();
+        try {
+            $filter = StatisticsParametersFilter::fromRequest($request, $this->getDoctrine()->getRepository(Committee::class));
```
B-Galati: Why is CommitteeRepository not injected into method name ?

src/Controller/Api/ReferentController.php
```diff
@@ -0,0 +1,60 @@
...
+    private const AUTOCOMPLETE_TYPE_COUNTRY = 'country';
+
+    private const AUTOCOMPLETE_TYPES = [
+        self::AUTOCOMPLETE_TYPE_CITY,
+        self::AUTOCOMPLETE_TYPE_COMMITTEE,
+        self::AUTOCOMPLETE_TYPE_COUNTRY,
+    ];
+
+    /**
+     * @Route("/search/autocomplete", defaults={"_enable_campaign_silence"=true}, name="api_referent_space_search_autocomplete")
```
ottaviano: can you remove `defaults={"_enable_campaign_silence"=true}` pls?

src/Repository/CommitteeRepository.php
```diff
@@ -358,6 +359,48 @@ public function countApprovedForReferent(Adherent $referent): int
...
+            ->where('committee.status = :status')
+            ->andWhere('committee.name LIKE :searchedName')
+            ->andWhere('tag.id IN (:tags)')
+            ->setParameter('searchedName', $value.'%')
+            ->setParameter('status', Committee::APPROVED)
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
+            ->orderBy('committee.name')
+        ;
+
+        return array_map(function (array $committee) {
```
ottaviano: is it really needed?
what about return just array of :
```
[
  {
    "uuid": "uuid_1",
    "name": "name_1"
  },
  {
    "uuid": "uuid_2",
    "name": "name_2"
  }
]
```

instead of

```
[
  {
    "uuid_1": "name_1"
  },
  {
    "uuid_2": "nama_2"
  }
]
```

src/Repository/CommitteeRepository.php
```diff
@@ -358,6 +359,48 @@ public function countApprovedForReferent(Adherent $referent): int
...
+            ->where('committee.status = :status')
+            ->andWhere('committee.postAddress.cityName LIKE :searchedCityName')
+            ->andWhere('tag.id IN (:tags)')
+            ->setParameter('searchedCityName', $value.'%')
+            ->setParameter('status', Committee::APPROVED)
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
+            ->orderBy('city')
+        ;
+
+        return array_map(function (array $city) {
```
ottaviano: use `array_column` pls

src/Repository/EventRepository.php
```diff
@@ -131,9 +134,7 @@ protected function createSlugQueryBuilder(string $slug): QueryBuilder
      */
     public function findManagedBy(Adherent $referent): array
     {
-        if (!$referent->isReferent()) {
-            return [];
-        }
+        $this->checkReferent($referent);
```
ottaviano: not the same thing, `return []` vs `throw new Exception`

src/Repository/EventRepository.php
```diff
@@ -425,4 +426,89 @@ public function paginate(int $offset = 0, int $limit = SearchParametersFilter::D
...
+            ->andWhere('event.committee IS NOT NULL')
+            ->andWhere('event.postAddress.cityName LIKE :searchedCityName')
+            ->andWhere('tag.id IN (:tags)')
+            ->setParameter('searchedCityName', $value.'%')
+            ->setParameter('status', BaseEvent::STATUS_SCHEDULED)
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
+            ->orderBy('city')
+        ;
+
+        return array_map(function (array $city) {
```
ottaviano: `array_column`

src/Repository/EventRepository.php
```diff
@@ -425,4 +426,89 @@ public function paginate(int $offset = 0, int $limit = SearchParametersFilter::D
...
+            return $city['city'];
+        }, $qb->getQuery()->getArrayResult());
+    }
+
+    public function queryCountByMonth(Adherent $referent, int $months = 5): QueryBuilder
+    {
+        return $this->createQueryBuilder('event')
+            ->select('COUNT(event) AS count, YEAR_MONTH(event.beginAt) as yearmonth')
```
ottaviano: can you use the same style for `AS` pls :)

src/Statistics/StatisticsParametersFilter.php
```diff
@@ -0,0 +1,75 @@
...
+    private const PARAMETER_CITY_NAME = 'city';
+    private const PARAMETER_COUNTRY_CODE = 'country';
+
+    private $committee = null;
+    private $cityName = null;
+    private $countryCode = null;
+
+    public static function fromRequest(Request $request, CommitteeRepository $committeeRepository): StatisticsParametersFilter
+    {
+        $filter = new StatisticsParametersFilter();
```
ottaviano: `new self`

src/Statistics/StatisticsParametersFilter.php
```diff
@@ -0,0 +1,75 @@
...
+
+        if ($uuid = $request->query->get(self::PARAMETER_COMMITTEE_UUID)) {
+            if ($committee = $committeeRepository->findOneByUuid($uuid)) {
+                $filter->setCommittee($committee);
+            } else {
+                throw new InvalidArgumentException("There is no committee with UUID '$uuid'.");
```
ottaviano: `\InvalidArgumentException`

src/Statistics/StatisticsParametersFilter.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    private const PARAMETER_COMMITTEE_UUID = 'committee';
+    private const PARAMETER_CITY_NAME = 'city';
+    private const PARAMETER_COUNTRY_CODE = 'country';
+
+    private $committee = null;
+    private $cityName = null;
+    private $countryCode = null;
+
+    public static function fromRequest(Request $request, CommitteeRepository $committeeRepository): StatisticsParametersFilter
```
ottaviano: `createFromRequest`

src/Repository/EventRepository.php
illusionOfParadise: Yes, but in this case it's better to throw an exception because this method should not be called with non referent.

src/Controller/Api/EventsController.php
illusionOfParadise: It will also be used in EventRepository.

src/Repository/CommitteeRepository.php
illusionOfParadise: It was defined with front dev.

src/Controller/Api/EventsController.php
```diff
@@ -2,20 +2,48 @@
...
+    /**
+     * @Route("/count-by-month", name="app_committee_events_count_by_month")
+     * @Method("GET")
+     * @Security("is_granted('ROLE_REFERENT')")
+     */
+    public function eventsCountInReferentManagedAreaAction(Request $request, EventRepository $eventRepository): Response
+    {
+        $referent = $this->getUser();
+        try {
+            $filter = StatisticsParametersFilter::createFromRequest($request, $this->getDoctrine()->getRepository(Committee::class));
```
B-Galati: The repository could be injected through the method.

PR #2466, opened 2018-05-28
rename transaction table to transactions #anto ;)

cacahouwete:


src/Entity/Transaction.php
```diff
@@ -7,7 +7,7 @@
 
 /**
  * @ORM\Entity(repositoryClass="AppBundle\Repository\TransactionRepository")
- * @ORM\Table(indexes={
+ * @ORM\Table(name="transactions", indexes={
```
ottaviano: @atamano suggested `donation_transactions`, no?

cacahouwete: ok

src/Entity/Transaction.php
```diff
@@ -7,7 +7,7 @@
 
 /**
  * @ORM\Entity(repositoryClass="AppBundle\Repository\TransactionRepository")
- * @ORM\Table(indexes={
+ * @ORM\Table(name="donation_transactions", indexes={
  *     @ORM\Index(name="transaction_result_idx", columns={"paybox_result_code"})
```
illusionOfParadise: To be consistent with the renaming: `transaction_result_idx` => `donation_transaction_result_idx`

PR #2467, opened 2018-05-29
[ch4192] Add payment method in historic donation

cacahouwete:
https://app.clubhouse.io/enmarche/story/4192/ajouter-le-moyen-de-paiement-dans-l-historique-des-dons-sur-l-onglet-mes-dons

### DIFF
--- a/app/migrations/Version20180529102348.php
+++ b/app/migrations/Version20180529102348.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20180529102348 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE donation_transactions ADD type VARCHAR(255) NOT NULL DEFAULT \'cb\'');
+        $this->addSql('ALTER TABLE donation_transactions CHANGE type type VARCHAR(255) NOT NULL');
+    }
+
+    public function down(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE donation_transactions DROP type');
+    }
+}
--- a/src/Entity/Transaction.php
+++ b/src/Entity/Transaction.php
@@ -23,6 +23,10 @@ class Transaction
     public const PAYBOX_CARD_UNAUTHORIZED = '000021';
     public const PAYBOX_PAYMENT_PAGE_TIMEOUT = '000030';
 
+    public const TYPE_CB = 'cb';
+    public const TYPE_CHECK = 'check';
+    public const TYPE_TRANSFER = 'transfer';
+
     /**
      * @var int
      *
@@ -75,6 +79,13 @@ class Transaction
     private $payboxSubscriptionId;
 
     /**
+     * @var string
+     *
+     * @ORM\Column
+     */
+    private $type;
+
+    /**
      * @var \DateTimeImmutable
      *
      * @ORM\Column(type="datetime_immutable")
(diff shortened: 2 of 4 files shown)


src/Entity/Transaction.php
```diff
@@ -23,6 +23,10 @@ class Transaction
     public const PAYBOX_CARD_UNAUTHORIZED = '000021';
     public const PAYBOX_PAYMENT_PAGE_TIMEOUT = '000030';
 
+    public const TYPE_PAYBOX = 'paybox';
+    public const TYPE_CHEQUE = 'cheque';
```
ottaviano: why switch to FR for this const? can you stay in ENG pls?

src/Entity/Transaction.php
```diff
@@ -23,6 +23,10 @@ class Transaction
     public const PAYBOX_CARD_UNAUTHORIZED = '000021';
     public const PAYBOX_PAYMENT_PAGE_TIMEOUT = '000030';
 
+    public const TYPE_PAYBOX = 'paybox';
```
ottaviano: it's not a payment type, it's a provider, can you rename for `CB` pls?

app/migrations/Version20180529102348.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20180529102348 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE transaction ADD type VARCHAR(255) NOT NULL DEFAULT \'paybox\'');
```
ottaviano: `donation_` ?

src/Entity/Transaction.php
cacahouwete: in english it's cheque or check when I try google translate. :s

app/migrations/Version20180529102348.php
cacahouwete: At this point it wasn't renamed ^^, but I will change it

src/Entity/Transaction.php
ottaviano: lol you are right, cheque is in eng too :D thanks wiki

translations/messages.fr.yml
```diff
@@ -313,3 +313,10 @@ timeline.measure.status.upcoming: à venir
 # UserDocument
 #
 document.upload.success: Le document a été uploadé avec succès.
+
+#
+# Transactions
+#
+transaction.type.paybox: Carte bleue
```
ottaviano: cd

translations/messages.fr.yml
```diff
@@ -313,3 +313,10 @@ timeline.measure.status.upcoming: à venir
...
+transaction.type.cheque: Chèque
```
ottaviano: check

app/migrations/Version20180529102348.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20180529102348 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE donation_transactions ADD type VARCHAR(255) NOT NULL DEFAULT \'cb\'');
```
B-Galati: Not a big fan of having a default value on the column.
It's fragile.
If for whatever reason we try to insert a new line in this table it would work while we potentially have a bug in the codebase.

cacahouwete: That's why I alter it with an other change to remove default value ^^,

B-Galati: My bad, thank you for pointing this out ;-)

PR #2470, opened 2018-05-29
[ch2541] Put all OG image dimensions on fix value

cacahouwete:
https://app.clubhouse.io/enmarche/story/2541/redimensionner-les-images-open-graph-et-twitter-card


src/MediaGenerator/Command/AbstractCitizenProjectMediaCommand.php
```diff
@@ -24,7 +24,7 @@
     /**
      * @var File
      *
-     * @Assert\Image(minWidth=1200)
+     * @Assert\Image(minWidth=1600)
```
ottaviano: I think not need change this, it's for generate an image for facebook cover which should have 1200px for width

JulHb:
Attention c'est du 1600x840, pas 1600x800

PR #2474, opened 2018-05-31
[CH3647] Add route to count total event partipants in referent areas

B-Galati:
https://app.clubhouse.io/enmarche/story/3647/3e-partie-section-evenements-box-kpis-inscrits-events

### DIFF
--- a/src/Controller/Api/EventsController.php
+++ b/src/Controller/Api/EventsController.php
@@ -2,7 +2,7 @@
 
 namespace AppBundle\Controller\Api;
 
-use AppBundle\Entity\Committee;
+use AppBundle\Repository\CommitteeRepository;
 use AppBundle\Repository\CitizenActionRepository;
 use AppBundle\Repository\EventRepository;
 use AppBundle\Statistics\StatisticsParametersFilter;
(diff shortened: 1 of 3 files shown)


src/Repository/EventRepository.php
```diff
@@ -515,6 +515,28 @@ public function queryCountByMonth(Adherent $referent, int $months = 5): QueryBui
...
+    {
+        $this->checkReferent($referent);
+
+        $results = $this->createQueryBuilder('event')
+            ->select('event.participantsCount')
+            ->innerJoin('event.referentTags', 'tag')
+            ->where('tag IN (:tags)')
+            ->andWhere('event.committee IS NOT NULL')
+            ->andWhere("event.status = '".Event::STATUS_SCHEDULED."'")
+            ->andWhere('event.organizer != :referent')
```
illusionOfParadise: Technically a referent could be a committee supervisor or host and also committee event organiser. And in this case, it's not correct to add this condition.
To be sure that we get only committee  events it's enough to check that committee is not null as you did here https://github.com/EnMarche/en-marche.fr/pull/2474/files#diff-265bef353d2bcbcdada5c1736d191b53R526

src/Repository/EventRepository.php
```diff
@@ -515,6 +515,26 @@ public function queryCountByMonth(Adherent $referent, int $months = 5): QueryBui
...
+            ->andWhere('event.committee IS NOT NULL')
+            ->andWhere("event.status = '".Event::STATUS_SCHEDULED."'")
+            ->andWhere('event.participantsCount > 0')
+            ->groupBy('event.id')
+            ->setParameter('tags', $referent->getManagedArea()->getTags())
+            ->getQuery()
+            ->getArrayResult()
+        ;
+
+        return array_sum(array_column($results, 'participantsCount'));
```
ottaviano: I think it's better to do this stuff with sgbd to avoid the data transfert between db and php.

B-Galati: Agree I'll see if I can do something better but not sure.

src/Controller/Api/EventsController.php
```diff
@@ -66,4 +66,18 @@ public function allTypesEventsCountInReferentManagedAreaAction(EventRepository $
...
     }
+
+    /**
+     * @Route("/count-participants", name="app_committee_events_count_participants")
+     * @Method("GET")
+     * @Security("is_granted('ROLE_REFERENT')")
+     */
+    public function eventsCountInReferentManagedArea(EventRepository $eventRepository): Response
+    {
+        $referent = $this->getUser();
```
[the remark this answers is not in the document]
B-Galati: It will be in #2477 :D

PR #2477, opened 2018-06-01
[ch3641] monthly event participants

B-Galati:


src/Repository/EventRegistrationRepository.php
```diff
@@ -188,11 +188,44 @@ public function countEventParticipantsInReferentManagedArea(Adherent $referent,
...
         return RepositoryUtils::aggregateCountByMonth($query->getQuery()->getArrayResult(), 'event_participants');
     }
 
+    public function countEventParticipantsInReferentManagedAreaInAtLeastOneCommittee(Adherent $referent, StatisticsParametersFilter $filter = null, int $months = 5): array
+    {
+        $this->checkReferent($referent);
+
+        $query = $this->createQueryBuilder('event_registrations')
+            ->select('DISTINCT event_registrations.emailAddress, COUNT(DISTINCT event_registrations) AS count, YEAR_MONTH(event.beginAt) as yearmonth')
+            ->join(Event::class, 'event', Join::WITH, 'event_registrations.event = event.id')
```
ottaviano: `->join('event_registrations.event', 'event')`, no?

src/Repository/EventRegistrationRepository.php
```diff
@@ -188,11 +188,44 @@ public function countEventParticipantsInReferentManagedArea(Adherent $referent,
...
+            ->join('event.referentTags', 'evenTag')
```
ottaviano: eventTag

src/Repository/EventRegistrationRepository.php
```diff
@@ -188,11 +188,44 @@ public function countEventParticipantsInReferentManagedArea(Adherent $referent,
...
 
         return RepositoryUtils::aggregateCountByMonth($query->getQuery()->getArrayResult(), 'event_participants');
     }
 
+    public function countEventParticipantsInReferentManagedAreaInAtLeastOneCommittee(Adherent $referent, StatisticsParametersFilter $filter = null, int $months = 5): array
+    {
+        $this->checkReferent($referent);
+
+        $query = $this->createQueryBuilder('event_registrations')
+            ->select('DISTINCT event_registrations.emailAddress, COUNT(DISTINCT event_registrations) AS count, YEAR_MONTH(event.beginAt) as yearmonth')
```
ottaviano: why do you need the email address? It's useless with `->groupBy('yearmonth')` no?

src/Repository/EventRegistrationRepository.php
B-Galati: yes thanks !

B-Galati: In fact no, it's because I only want to join on `Event` and not `BaseEvent` (Basically I don't want citizenAction)

src/Repository/EventRegistrationRepository.php
B-Galati: I do agree but perhaps we are missing something. @illusionOfParadise What do you think ?

illusionOfParadise: One person with the same email adresse can be registered to several events in one month and we need to take it into account only once.

B-Galati: In fact no, we talk in PM with @illusionOfParadise and we figured out that one person should be taken into account more than once.

B-Galati: So this distinct is removed.

src/Repository/EventRegistrationRepository.php
```diff
@@ -173,26 +173,61 @@ public function countEventParticipantsInReferentManagedArea(Adherent $referent,
     {
         $this->checkReferent($referent);
 
-        $query = $this->createQueryBuilder('event_registrations')
-            ->select('DISTINCT event_registrations.emailAddress, COUNT(DISTINCT event_registrations) AS count, YEAR_MONTH(event.beginAt) as yearmonth')
-            ->join(Event::class, 'event', Join::WITH, 'event_registrations.event = event.id')
-            ->leftJoin(Adherent::class, 'adherent', Join::WITH, 'adherent.uuid = event_registrations.adherentUuid')
+        $query = $this->createQueryBuilder('eventRegistrations')
+            ->select('COUNT(DISTINCT eventRegistrations) AS count, YEAR_MONTH(event.beginAt) as yearmonth')
```
illusionOfParadise: We should not modify the count because this method is already used in another endpoint to calcul number of different persons participated in all events in managed area. 
We should add a parameter to define a count calculation or create another method.

src/Repository/EventRegistrationRepository.php
```diff
@@ -173,26 +173,61 @@ public function countEventParticipantsInReferentManagedArea(Adherent $referent,
...
+            ->join(Event::class, 'event', Join::WITH, 'eventRegistrations.event = event.id')
+            ->leftJoin(Adherent::class, 'adherent', Join::WITH, 'adherent.uuid = eventRegistrations.adherentUuid')
```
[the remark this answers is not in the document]
B-Galati: Thanks!

PR #2480, opened 2018-06-04
Bugfix - Add missing lib webmozart/assert

B-Galati:


composer.lock
```diff
@@ -7893,116 +7893,6 @@
             ],
             "time": "2018-04-30T19:57:29+00:00"
         },
-        {
-            "name": "symfony/polyfill-ctype",
```
ottaviano: normal this deletion?

B-Galati: Yes I think it's because someone ran composer update with his local composer instead of the one provided by docker.

PR #2491, opened 2018-06-05
Create a functional test for paybox ipn callback

cacahouwete:


app/config/config_test.yml
```diff
@@ -108,3 +108,6 @@ old_sound_rabbit_mq:
                     - event.*
                     - committee.*
             callback: Tests\AppBundle\Consumer\DummyConsumer
+
+lexik_paybox:
+    public_key: "%kernel.project_dir%/var/public.key"
```
B-Galati: use env var here too.

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * @depends testOnPayboxIpnResponse
+     */
+    public function testAllTransactionCreated(): void
+    {
+        $transactions = $this->transactionRepository->findAllSuccessfulTransactionByEmail('jacques.picard@en-marche.fr');
+
+        // 4 because there are the initiale transaction when donationFixtures are loaded
```
Remg: b/c there are initial transactions*

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+        static::assertCount(4, $transactions);
+    }
+
+    private function createSignature(array &$params): void
```
Remg: is this reference needed here?

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+            'subscription' => '1234',
+        ];
+    }
+
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->init([
+            LoadDonationData::class,
```
Remg: Master of all lords (tmtc @B-Galati) teached me we should declare fixtures dependencies explicitly to avoid confusion.

Please add `LoadAdherentData::class` since `LoadDonationdata::class` depends on it.

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+     * @dataProvider payloadProvider
+     */
+    public function testOnPayboxIpnResponse(string $transactionId): void
+    {
+        $params = $this->createPayload($transactionId);
+        static::assertNull($this->transactionRepository->findByPayboxTransactionId($transactionId));
+
+        $this->createSignature($params);
+
+        $this->client->request('POST', '/don/payment-ipn/1528114726', $params);
```
Remg: Please use constants for such values: `Symfony\Component\HttpFoundation\Request::METHOD_POST`

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+use AppBundle\DataFixtures\ORM\LoadDonationData;
+use AppBundle\Repository\DonationRepository;
+use AppBundle\Repository\TransactionRepository;
+use Tests\AppBundle\Controller\ControllerTestTrait;
+use Tests\AppBundle\MysqlWebTestCase;
+
+/**
+ * @group functional
+ */
+class TransactionSubscriberTest extends MysqlWebTestCase
```
Remg: You can extend `Tests\AppBundle\SqliteWebTestCase` if these tests do not require mysql-specific functions.
Functional test suite will be faster \o/

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * @dataProvider payloadProvider
+     */
+    public function testOnPayboxIpnResponse(string $transactionId): void
+    {
+        $params = $this->createPayload($transactionId);
+        static::assertNull($this->transactionRepository->findByPayboxTransactionId($transactionId));
+
+        $this->createSignature($params);
```
Remg: Could we add the signature directly in the `createPayload` method ?

tests/Donation/TransactionSubscriberTest.php
```diff
@@ -0,0 +1,109 @@
...
+        yield 'transaction3' => ['424243'];
+    }
+
+    /**
+     * @dataProvider payloadProvider
+     */
+    public function testOnPayboxIpnResponse(string $transactionId): void
+    {
+        $params = $this->createPayload($transactionId);
+        static::assertNull($this->transactionRepository->findByPayboxTransactionId($transactionId));
```
Remg: Could be done first since if this fails then `$params` is useless

tests/Donation/TransactionSubscriberTest.php
cacahouwete: There is 1 initial transaction then 1+3 = 4

tests/Donation/TransactionSubscriberTest.php
cacahouwete: I think it's useless because if tomorrow we decide to remove LoadAdherent, the fixture will be loaded for nothing and technically the test does not use Adherents datas.

Btw Adherent's data are used to generate donations data by using data already set in adherent but we can totaly remove it and set data with a faker or ... , the only link between donation and user account is the email address . there are no foreign key, ... .
```php
        $donation = new Donation(
            $uuid = Uuid::uuid4(),
            $uuid->toString().'_'.$this->slugify->slugify($adherent->getFullName()),
            $amount * 100,
            $adherent->getGender(),
            $adherent->getFirstName(),
            $adherent->getLastName(),
            $adherent->getEmailAddress(),
            $adherent->getPostAddress(),
            $adherent->getPhone(),
            '127.0.0.1',
            $duration
        );
```

tests/Donation/TransactionSubscriberTest.php
cacahouwete: Don't know, it was for pleasure to use references ^^, and set 'Sign' in the method

tests/Donation/TransactionSubscriberTest.php
Remg: fair enough !

PR #2496, opened 2018-06-07
[ch3637] count should be done on adherent and total

B-Galati:
https://app.clubhouse.io/enmarche/story/3637/1ère-partie-section-adhérents-graph-double-histogramme-emails-subscription

### DIFF
--- a/src/Controller/Api/EventsController.php
+++ b/src/Controller/Api/EventsController.php
@@ -77,8 +77,8 @@ class EventsController extends Controller
 
         return new JsonResponse([
             'total' => $eventRepository->countParticipantsInReferentManagedArea($referent),
-            'event_participants' => $eventRegistrationRepository->countEventParticipantsInReferentManagedArea($referent),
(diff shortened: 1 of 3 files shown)


src/Controller/Api/EventsController.php
```diff
@@ -77,8 +77,8 @@ public function eventsCountInReferentManagedArea(EventRepository $eventRepositor
 
         return new JsonResponse([
             'total' => $eventRepository->countParticipantsInReferentManagedArea($referent),
-            'event_participants' => $eventRegistrationRepository->countEventParticipantsInReferentManagedArea($referent),
-            'in_at_least_one_committee' => $eventRegistrationRepository->countEventParticipantsInReferentManagedAreaInAtLeastOneCommittee($referent),
```
cacahouwete: Change naming will have no problems with apps using it ?

B-Galati: Not used yet ;-)

PR #2497, opened 2018-06-08
[MOOC] Added MOOC API endpoint

thomasbisignani:


src/Normalizer/MoocNormalizer.php
```diff
@@ -0,0 +1,120 @@
...
+
+    private function normalizeFiles(Collection $files): array
+    {
+        $attachmentFiles = [];
+
+        /** @var AttachmentFile $file */
+        foreach ($files as $file) {
+            $attachmentFiles = [
+                'title' => $file->getTitle(),
+                'file' => $this->router->generate('mooc_get_file', ['id' => $file->getId()], UrlGenerator::ABSOLUTE_URL),
```
Remg: as discussed, this should be the file uuid instead of id.

thomasbisignani: Yes, I am waiting for @cacahouwete  modification on his branch.

PR #2499, opened 2018-06-11
[ch4294] Add autocomplete address via google maps api

cacahouwete:
https://app.clubhouse.io/enmarche/story/4194/int%C3%A9grer-le-place-autocomplete-au-formulaire-de-don

### DIFF
--- a/front/pages/donation_informations.js
+++ b/front/pages/donation_informations.js
@@ -1,7 +1,16 @@
 import formValidator from '../validator/formValidator';
+import GooglePlaceAutocomplete from '../services/address/GooglePlaceAutocomplete';
 
 export default (formType) => {
-    const form = dom('form[name="app_donation"]');
+    formValidator(formType, dom('form[name="app_donation"]'));
 
-    formValidator(formType, form);
+    const autocomplete = new GooglePlaceAutocomplete(
+        dom('.address-autocomplete'),
+        dom('#app_donation_address'),
+        dom('#app_donation_cityName'),
+        dom('#app_donation_postalCode'),
+        dom('#app_donation_country'),
+        'form form--full form__field'
+    );
+    autocomplete.build();
 };
--- a/front/services/address/GooglePlaceAutocomplete.js
+++ b/front/services/address/GooglePlaceAutocomplete.js
@@ -1,0 +1,71 @@
+export default class GooglePlaceAutocomplete {
+    constructor(wrapper, addressField, cityField, zipCodeField, countryField, inputClassNames = '') {
+        this._wrapper = wrapper;
+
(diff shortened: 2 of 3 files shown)


templates/membership/join.html.twig
```diff
@@ -26,8 +27,65 @@
...
+                        domElement.value += ' ';
+                    }
+                    domElement.value += val;
+                    domElement.dispatchEvent(new InputEvent('input'));
+                }
+            }
+        }
     </script>
     <script src="https://www.google.com/recaptcha/api.js"></script>
+    <script src="https://maps.googleapis.com/maps/api/js?key={{ google_maps_api_key }}&libraries=places&callback=initAutocomplete"
```
hhamon: `&` should be `&amp;`

templates/membership/join.html.twig
```diff
@@ -11,6 +11,7 @@
 {% block open_graph_image_height 630 %}
 
 {% block stylesheets %}
+    <link type="text/css" rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
```
B-Galati: I guess we don't want to use public CDN for this case no ?

templates/membership/join.html.twig
```diff
@@ -22,12 +23,19 @@
...
+                /** @type {!HTMLInputElement} */(document.getElementById('autocomplete')),
+                {types: ['geocode']});
+            Kernel.onLoad(function() {
+                App.runJoin('{{ form.vars.form_type_class|e('js') }}', autocomplete);
+                {{ js.address(form.address) }}
+            });
+        }
     </script>
     <script src="https://www.google.com/recaptcha/api.js"></script>
+    <script src="https://maps.googleapis.com/maps/api/js?key={{ google_maps_api_key }}&amp;libraries=places&amp;callback=initAutocomplete"
```
B-Galati: It looks like a security issue.
We should not exposed our secret in our sourcecode.
I saw that we did the same in many places in the code. Someone have an explanation ?
@ottaviano @illusionOfParadise @thomasbisignani @Remg  ?

templates/membership/join.html.twig
ottaviano: fixed, thanks @hhamon 👍

templates/membership/join.html.twig
ottaviano: removed :D

templates/membership/join.html.twig
ottaviano: @B-Galati, to secure the API key @atamano should restrict access for this key on GCloud console if not done yet

PR #2502, opened 2018-06-11
[MOOC] Fixed Upload files and updated Mooc Admin

thomasbisignani:


app/migrations/Version20180611115330.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20180611115330 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE mooc_attachment_file ADD slug VARCHAR(255) NOT NULL, DROP uuid');
```
B-Galati: The unique index constraint is missing

templates/admin/mooc/list_type.html.twig
```diff
@@ -0,0 +1,12 @@
+{% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
+
+{% block field %}
+    {% if object.getType is defined %}
+        {% if object.getType == constant('ELEMENT_TYPE_VIDEO', object) %}
```
B-Galati: Would be easier with a new method in `object` class.

templates/admin/mooc/list_type.html.twig
```diff
@@ -0,0 +1,12 @@
...
+            <span class="label label-danger">Youtube</span><br/>
```
B-Galati: (Optional) you could use a translation key instead of `if` statements

templates/admin/mooc/list_type.html.twig
```diff
@@ -0,0 +1,12 @@
...
+        {% endif %}
+        {% if object.getType == constant('ELEMENT_TYPE_QUIZ', object) %}
```
B-Galati: `elseif` would be better as we know it's possible to get there if the 1st if condition is met.

app/migrations/Version20180611115330.php
thomasbisignani: This PR is in WIP. I will push soon a new commit, thx :100: :1st_place_medal: :man_cartwheeling:

src/Entity/Mooc/BaseMoocElement.php
```diff
@@ -212,6 +212,16 @@ public function removeFile(AttachmentFile $file): void
 
     abstract public function getType(): string;
 
+    public function isVideo(): bool
+    {
+        return BaseMoocElement::ELEMENT_TYPE_VIDEO == $this->getType();
```
B-Galati: triple equal please :D

thomasbisignani: My bad. It's fixed :joy: :+1:

PR #2507, opened 2018-06-12
Add deputy role to adherent admin

illusionOfParadise:


src/Admin/DistrictAdmin.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $formMapper
+            ->add('name', TextType::class, [
+                'label' => 'Nom',
+                'filter_emojis' => true,
+            ])
+            ->add('countries', TextType::class, [
+                'label' => 'Pays',
+            ])
+            ->add('geoShape', TextType::class, [
```
ottaviano: it's really needed this admin? if yes, do you try create one district?
I think a `TextGeoShapeDataTransformer` is missing here, no?

src/Admin/DistrictAdmin.php
```diff
@@ -0,0 +1,48 @@
...
+class DistrictAdmin extends AbstractAdmin
+{
+    protected function configureFormFields(FormMapper $formMapper)
+    {
+        $formMapper
+            ->add('name', TextType::class, [
+                'label' => 'Nom',
+                'filter_emojis' => true,
+            ])
+            ->add('countries', TextType::class, [
```
ottaviano: `CountryType` ?

src/Entity/Adherent.php
```diff
@@ -1120,7 +1120,25 @@ public function getManagedDistrict(): ?District
 
     public function setManagedDistrict(?District $district): void
     {
+        if ($district
+            && $district->getAdherent()
+            && $district->getAdherent()->getUuid()->toString() !== $this->getUuid()->toString()) {
```
ottaviano: `!$this->equals($district->getAdherent())`

src/Admin/DistrictAdmin.php
illusionOfParadise: No, it's needed only to modify a district from AdherentAdmin.

illusionOfParadise: This admin is not shown in the admin dashboard and menu.

ottaviano: oh I see, then the fields `countries` and `geoShape` are useless here, no? as you use only name field in AdherentAdmin.

PR #2509, opened 2018-06-12
Change quizz fixture to have a real typeform for front test

cacahouwete:


src/Normalizer/MoocNormalizer.php
```diff
@@ -75,14 +77,14 @@ private function normalizeElement(BaseMoocElement $element): array
             'attachments' => $this->normalizeFiles($element->getFiles()),
         ];
 
-        if (BaseMoocElement::ELEMENT_TYPE_VIDEO === $element->getType()) {
+        if ($element instanceof Video) {
```
thomasbisignani: You can now use `if ($element->isVideo()) {`

src/Normalizer/MoocNormalizer.php
```diff
@@ -75,14 +77,14 @@ private function normalizeElement(BaseMoocElement $element): array
...
             $moocElement['youtubeId'] = $element->getYoutubeId();
             $moocElement['youtubeThumbnail'] = $element->getYoutubeThumbnail();
             $moocElement['duration'] = $element->getDuration()->format('H:i:s');
         }
 
-        if (BaseMoocElement::ELEMENT_TYPE_QUIZ === $element->getType()) {
-            $moocElement['typeformEmbed'] = $element->getTypeForm();
+        if ($element instanceof Quizz) {
```
thomasbisignani: You can now use `if ($element->isQuiz()) {`

src/Entity/Mooc/Quizz.php
```diff
@@ -16,30 +16,31 @@ class Quizz extends BaseMoocElement
...
      * @var string
      *
-     * @ORM\Column(type="text")
+     * @ORM\Column
      *
      * @Assert\NotBlank
+     * @Assert\Url
      */
-    private $typeForm;
+    private $typeFormUrl;
```
thomasbisignani: Not a big deal but we can name it `$typeformUrl` without CamelCase for Typeform.

src/Normalizer/MoocNormalizer.php
cacahouwete: If I use this I will have to add a doc bloc for ide understand wich class is it ^^,

src/Normalizer/MoocNormalizer.php
cacahouwete: If I use this I will have to add a doc bloc for ide understand wich class is it ^^,

PR #2514, opened 2018-06-13
[MOOC] Added Twig test extension and Refacto elements

thomasbisignani:
- Renames Quizz for Quiz
- Adds Twig Test extension for Element type
- Removes `BaseMoocElement` issers and uses `instanceof` and a twig extension instead
- Removes type folder for the `AttachmentFile`

### DIFF
--- a/src/Normalizer/MoocNormalizer.php
+++ b/src/Normalizer/MoocNormalizer.php
@@ -7,11 +7,12 @@ use AppBundle\Entity\Mooc\AttachmentLink;
 use AppBundle\Entity\Mooc\BaseMoocElement;
 use AppBundle\Entity\Mooc\Chapter;
 use AppBundle\Entity\Mooc\Mooc;
-use AppBundle\Entity\Mooc\Quizz;
+use AppBundle\Entity\Mooc\Quiz;
 use AppBundle\Entity\Mooc\Video;
 use Doctrine\Common\Collections\Collection;
 use Symfony\Component\Routing\Generator\UrlGenerator;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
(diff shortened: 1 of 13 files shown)


src/Normalizer/MoocNormalizer.php
```diff
@@ -128,4 +128,17 @@ private function normalizeFiles(Collection $files): array
 
         return $attachmentFiles;
     }
+
+    private function getElementType(BaseMoocElement $element)
```
Remg: missing return type

src/Normalizer/MoocNormalizer.php
```diff
@@ -128,4 +128,17 @@ private function normalizeFiles(Collection $files): array
...
+    {
+        if ($element instanceof Video) {
+            return 'video';
+        }
+
+        if ($element instanceof Quiz) {
+            return 'quiz';
+        }
+
+        return '';
```
Remg: why not `null` here ?

src/Twig/MoocElementExtension.php
```diff
@@ -0,0 +1,23 @@
...
+use AppBundle\Entity\Mooc\Quiz;
+use AppBundle\Entity\Mooc\Video;
+use Twig\Extension\AbstractExtension;
+
+class MoocElementExtension extends AbstractExtension
+{
+    public function getTests()
+    {
+        return [
+            new \Twig_SimpleTest('video', function (BaseMoocElement $element) {
```
Remg: to avoid any future confusion i think it should be `if object is mooc_video`, and not `is video`
on this application another kind of object can be more generic video: `AppBundle\Entity\Media`.

what do you think?

src/Twig/MoocElementExtension.php
```diff
@@ -0,0 +1,23 @@
...
+                return $element instanceof Video;
+            }),
+            new \Twig_SimpleTest('quiz', function (BaseMoocElement $element) {
```
Remg: same here i guess ?

PR #2517, opened 2018-06-14
Add deputy role to adherent role filter in admin

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/4438/dans-l-admin-rajouter-dans-le-filtre-r%C3%B4le-celui-de-d%C3%A9put%C3%A9

### DIFF
--- a/src/Admin/AdherentAdmin.php
+++ b/src/Admin/AdherentAdmin.php
@@ -440,6 +440,12 @@ class AdherentAdmin extends AbstractAdmin
                         $qb->setParameter('committee_privileges', $privileges);
                     }
 
+                    // Deputy
+                    if (\in_array(AdherentRoleEnum::DEPUTY, $value['value'], true)) {
+                        $qb->leftJoin(sprintf('%s.managedDistrict', $alias), 'district');
+                        $where->add('district IS NOT NULL');
+                    }
+
                     // Board Member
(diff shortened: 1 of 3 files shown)


src/Admin/AdherentAdmin.php
```diff
@@ -440,6 +440,12 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                         $qb->setParameter('committee_privileges', $privileges);
                     }
 
+                    // Deputy
+                    if (\in_array(AdherentRoleEnum::DEPUTY, $value['value'], true)) {
+                        $qb->leftJoin(sprintf('%s.managedDistrict', $alias), 'district');
```
ottaviano: what about replace `leftJoin` by `innerJoin` and remove `where` condition?

illusionOfParadise: @ottaviano I agree with `innerJoin`, but we need a condition when it's a filter by role https://github.com/EnMarche/en-marche.fr/pull/2517/files#diff-d435c0b62bf9827706902379559ff9c8R421 or we should add check that `where` condition has some parts to add it or not to query.

ottaviano: oh I see, keep your solution ;)

PR #2520, opened 2018-06-15
Speed up tests for CI 

B-Galati:
- Store MySQL data in memory
- Authenticate user without going to /login


thomasbisignani:
:1st_place_medal: :man_cartwheeling:

PR #2523, opened 2018-06-19
[MOOC] Updated share elements fields

thomasbisignani:


src/Admin/MoocElementAdmin.php
```diff
@@ -86,18 +86,18 @@ protected function configureFormFields(FormMapper $formMapper)
...
                 ])
-                ->add('emailObject', TextType::class, [
+                ->add('shareEmailSubject', TextType::class, [
                     'label' => 'Sujet de l\'email de partage',
                 ])
-                ->add('emailBody', PurifiedTextareaType::class, [
+                ->add('shareEmailBody', PurifiedTextareaType::class, [
                     'label' => 'Corps de l\'email de partage',
-                    'attr' => ['class' => 'ck-editor'],
+                    'attr' => ['rows' => 5],
```
ottaviano: what about add maxlength html attribute?

ottaviano: why do you transform this field from ckeditor to simple textarea?

src/Entity/Mooc/BaseMoocElement.php
```diff
@@ -111,47 +111,47 @@
...
      */
-    private $emailObject;
+    private $shareEmailSubject;
 
     /**
-     * @ORM\Column(length=800)
+     * @ORM\Column(length=500)
      *
-     * @Assert\Length(min=5, max=800)
+     * @Assert\Length(min=5, max=500)
```
ottaviano: NotBlank is missing

app/migrations/Version20180619100109.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20180619100109 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE mooc_elements ADD share_twitter_text VARCHAR(255) NOT NULL, ADD share_facebook_text VARCHAR(255) NOT NULL, ADD share_email_subject VARCHAR(255) NOT NULL, ADD share_email_body VARCHAR(500) NOT NULL, DROP twitter_text, DROP facebook_text, DROP email_object, DROP email_body');
```
ottaviano: do you have some data in dropped columns? if yes, maybe use CHANGE action instead of `ADD & DROP`?

PR #2525, opened 2018-06-19
Creates a command to update Donations reference

Remg:


src/Command/DonationUpdateReferenceCommand.php
```diff
@@ -0,0 +1,121 @@
...
+    private function updateDonationsReference(OutputInterface $output): void
+    {
+        $progressBar = new ProgressBar($output, $this->countDonations());
+
+        $count = 0;
+        foreach ($this->getDonations() as $result) {
+            $donation = reset($result);
+
+            $reference = $this->donationRequestUtils->buildDonationReference(
+                Uuid::fromString($donation->getUuid()),
```
ottaviano: `getUuid` returns UuidInterface object

src/Command/DonationUpdateReferenceCommand.php
```diff
@@ -0,0 +1,121 @@
...
+            $donation = reset($result);
+
+            $reference = $this->donationRequestUtils->buildDonationReference(
+                $donation->getUuid(),
+                sprintf('%s %s', $donation->getFirstName(), $donation->getLastName())
+            );
+
+            $donation->setPayboxOrderRef($reference);
+
+            $this->em->persist($donation);
```
B-Galati: I don't think it's needed to call `persist` here as the object comes from doctrine ORM and thus should be already attached.

PR #2530, opened 2018-06-20
Reduce memory leak for tests

B-Galati:


src/DataFixtures/ORM/LoadMissionTypeData.php
```diff
@@ -22,6 +23,12 @@ class LoadMissionTypeData implements FixtureInterface
 
     public function load(ObjectManager $manager)
     {
+        if (!$manager instanceof  EntityManagerInterface) {
```
salahm: you may use a trait here, with a function which takes 2 args : $manger, $tableName

app/config/config_test.yml
```diff
@@ -111,3 +111,10 @@ old_sound_rabbit_mq:
 
 lexik_paybox:
     public_key: "%env(SSL_PUBLIC_KEY)%"
+
+doctrine:
+    dbal:
+        connections:
```
salahm: you may also consider using at least an arrayCache pour doctrine metadata ;)

src/DataFixtures/ORM/LoadMissionTypeData.php
Remg: `$manger` & `$tableName` are great var names, same lexical field and all... :trollface:

app/config/config_test.yml
B-Galati: We will do that with [DoctrineTestBundle](https://github.com/dmaicher/doctrine-test-bundle) in another PR

src/DataFixtures/ORM/LoadMissionTypeData.php
B-Galati: I do agree

B-Galati: with @salahm lol

B-Galati:
Closing, I'll do another PR will all changes.

PR #2533, opened 2018-06-20
Redis instance is not required anymore for cache warmup

B-Galati:


app/config/services/cms.xml
```diff
@@ -52,7 +52,7 @@
         <service id="League\Flysystem\Filesystem" alias="app.storage" />
 
         <!-- Medias -->
-        <service id="app.glide.cache_adapter" class="Danhunsaker\Flysystem\Redis\RedisAdapter" public="false">
+        <service id="app.glide.cache_adapter" class="Danhunsaker\Flysystem\Redis\RedisAdapter" public="false" lazy="true">
```
B-Galati: Here is the trick

Dockerfile
```diff
@@ -98,7 +92,6 @@ RUN chmod 0444 gcloud-service-key.json && \
     mv docker/prod/nginx.conf /etc/nginx/nginx.conf && \
     mv docker/prod/supervisord.conf /etc/supervisor/conf.d/ && \
 
-    rm -rf docker && \
-    rm -rf web/app_dev.php
+    rm -rf docker composer.json composer.lock
```
ottaviano: `HttpKernel\Kernel` class uses the composer.json file to find the `project_dir`

B-Galati: Best catch ever :+1: Thank you !

PR #2536, opened 2018-06-20
Remove /campus/mooc page

Remg:


Remg:
@atamano please remove the cloudflare redirection once this is deployed

PR #2543, opened 2018-06-21
Change district import command to process another import file

illusionOfParadise:


Remg:
Should you change the loader injected in the [ImportDistrictsCommand](https://github.com/EnMarche/en-marche.fr/blob/master/src/Command/ImportDistrictsCommand.php) then? or only in fixtures is enough?

src/Deputy/DistrictLoader.php
```diff
@@ -60,20 +59,21 @@ public function batchInsertOrUpdateDistricts(array $districts): void
...
 
-            $geoDistricts = array_filter($this->geoDistricts, function ($geoDistrict) use ($code) {
-                return $geoDistrict['properties']['ID'] === $code;
+            $geoDistricts = array_filter($this->geoDistricts, function ($geoDistrict) use ($codeDepartement, $number) {
+                return $geoDistrict['fields']['departement'] === $codeDepartement && (int) $geoDistrict['fields']['circonscription'] === $number;
             });
 
             if (0 === count($geoDistricts)) {
-                throw new \RuntimeException("Districts GeoJSON file doesn't contain district with code '$code'");
+                throw new \RuntimeException("\"Districts GeoJSON file doesn't contain district with number '$number' and department code '$codeDepartement'");
```
ottaviano: `"\"D` -> `"D`

src/Deputy/LightFileDistrictLoader.php
```diff
@@ -0,0 +1,112 @@
...
+use Symfony\Component\Serializer\Encoder\JsonEncoder;
+use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
+use Symfony\Component\Serializer\Serializer;
+
+/**
+ * Class used to load districts using as $districtsFile the file /src/DataFixtures/deputy/france_circonscriptions_legislatives.json
+ * or https://www.data.gouv.fr/s/resources/carte-des-circonscriptions-legislatives-2012-et-2017/20170721-135742/france-circonscriptions-legislatives-2012.json
+ * Actually used for loading fixtures.
+ */
+class LightFileDistrictLoader
```
ottaviano: Is this class used only on fixtures?

illusionOfParadise:
The loader `DistrictLoader` used in the `ImportDistrictsCommand` is a new.  And `LightFileDistrictLoader` is the old loader that I continue to use for the fixtures. The new import file is much heavier that's why I prefer to keep the old import file and loader for fixtures.

src/Deputy/LightFileDistrictLoader.php
illusionOfParadise: Yes.

PR #2547, opened 2018-06-26
Add republican silence for deputy space

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/4440/int%C3%A9grer-l-espace-d%C3%A9put%C3%A9-dans-le-silence-r%C3%A9publicain

### DIFF
--- a/features/deputy.feature
+++ b/features/deputy.feature
@@ -40,7 +40,7 @@ Feature:
   Scenario: As deputy of 1st Paris district I can send message to the adherents.
     Given I am logged as "deputy@en-marche-dev.fr"
     When I am on "/espace-depute/utilisateurs/message"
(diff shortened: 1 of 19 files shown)


features/deputy.feature
```diff
@@ -106,6 +106,16 @@ Feature:
               "target_message": "Content of a deputy message",
               "target_firstname": "D\u00e9put\u00e9"
           }
+      },
+      {
+        "Email": "deputy-ch-li@en-marche-dev.fr",
+        "Name": "D\u00e9put\u00e9 CHLI FDESIX",
```
Remg: just fyi (no need to change anything here), thanks to PHPMatcher we now use in the `JsonContext`, we can use special characters in these tests.
you could write `"Name": "Député CHLI FDESIX"` here for example.
i think it makes tests more readable.

PR #2548, opened 2018-06-27
Updated newsletter success redirection for articles

thomasbisignani:


src/Controller/EnMarche/NewsletterController.php
```diff
@@ -31,6 +36,14 @@ public function subscriptionAction(Request $request)
         if ($form->isSubmitted() && $form->isValid()) {
             $this->get('app.newsletter_subscription.handler')->subscribe($subscription);
 
+            if ($path = $newsletterSubscriptionProcess->getSuccessRedirectPath() ?? null) {
+                $newsletterSubscriptionProcess->terminate();
+
+                $this->addFlash('info', $translator->trans('newsletter.subscription.success'));
```
ottaviano: flash message is translated in base.html.twig, you can use the key here

thomasbisignani: Nice catch! thx you :+1:

app/config/services/services.xml
```diff
@@ -213,5 +213,8 @@
 
         <!-- Deputy -->
         <service id="AppBundle\Deputy\DistrictLoader" public="true" />
+
+        <!-- Newsletter -->
+        <service id="AppBundle\Newsletter\NewsletterSubscriptionProcess"/>
```
HeahDude: It seems there is already a section for it https://github.com/EnMarche/en-marche.fr/blob/master/app/config/services/community.xml#L6. What about removing those and registering the namespace here?

PR #2553, opened 2018-06-28
Formation organisation

marineb:


templates/components/_footer.html.twig
```diff
@@ -56,14 +56,24 @@
...
                     </h4>
                 </li>
                 <li>
-                    <a href="{{ path('page_campus') }}">Toutes les ressources</a>
+                    <a href="https://touspolitiques.fr" target="_blank" rel="noopener noreferrer">
+                        L'institut de formation <em>Tous Politiques !</em>
+                    </a>
+                </li>
+                <li>
+                    <a href="https://mooc.en-marche.fr" target="_blank" rel="noopener noreferrer">
```
Remg: Should this link be `https://mooc.en-marche.fr/faire-de-son-coup-de-fourchette-un-acte-politique` instead?

When `mooc.en-marche.fr` will be updated with a new mooc, the label "Le MOOC sur l'alimentation" will still be relevant then

templates/page/campus/home.html.twig
```diff
@@ -63,9 +76,9 @@ gratuite et ouverte à tous.' %}
                 <div class="campus__text">
                     <h2>Se former</h2>
                     <p>
-                        Vous trouverez sur cette page nos cours sur la vie politique. 
-                        Ces « microlearning » sont des parcours éducatifs et ludiques à suivre en ligne, 
-                        à votre rythme, pour mieux comprendre le fonctionnement de la vie politique 
+                        Vous trouverez sur cette page nos cours sur la vie politique.
+                        Ces « microlearning » sont des parcours éducatifs et ludiques à suivre en ligne,
```
Remg: You wrote "microlearnings" with a "s" here : https://github.com/EnMarche/en-marche.fr/pull/2553/files#diff-1a752ab87d882a61d58251f2e3f00e07R75

but not here?

templates/components/_footer.html.twig
marineb: Yes, you're right. I'll update it !

PR #2558, opened 2018-07-03
Fix donation status

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/4498/faire-une-migration-pour-corriger-les-statuts-des-dons

### DIFF
--- a/app/migrations/Version20180703113033.php
+++ b/app/migrations/Version20180703113033.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Migrations;
+
+use AppBundle\Entity\Donation;
+use AppBundle\Entity\Transaction;
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20180703113033 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.duration != 0 AND donations.subscription_ended_at IS NULL',
+            ['status' => Donation::STATUS_SUBSCRIPTION_IN_PROGRESS]
+        );
+        $this->addSql(
+            'UPDATE donations '.
+            'JOIN '.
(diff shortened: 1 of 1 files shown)


app/migrations/Version20180703113033.php
```diff
@@ -0,0 +1,58 @@
...
+            ['status' => Donation::STATUS_SUBSCRIPTION_IN_PROGRESS]
+        );
+        $this->addSql(
+            'UPDATE donations '.
+            'JOIN '.
+                '(SELECT MAX(id) max_id, donation_id FROM donation_transactions GROUP BY donation_id) dt_max '.
+            'ON (dt_max.donation_id = donations.id) '.
+            'JOIN donation_transactions ON (donation_transactions.id = dt_max.max_id) '.
+            'SET donations.status = :status '.
+            'WHERE donation_transactions.paybox_result_code != :code',
```
Remg: shouldn't this be `WHERE dt_max.paybox_result_code != :code` ?
as we want to check if the last transaction is in error or not

illusionOfParadise: No, because it's already the last transaction `JOIN donation_transactions ON (donation_transactions.id = dt_max.max_id)`.

app/migrations/Version20180703113033.php
```diff
@@ -0,0 +1,58 @@
...
+final class Version20180703113033 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql(
+            'UPDATE donations SET donations.status = :status WHERE donations.duration != 0 AND donations.subscription_ended_at IS NULL',
+            ['status' => Donation::STATUS_SUBSCRIPTION_IN_PROGRESS]
+        );
+        $this->addSql(
+            'UPDATE donations '.
```
HeahDude: You can use heretic syntax instead of those concatenations and whitespace at end of lines:
```php
    $this->addSql(<<<EOSQL
UPDATE donations
JOIN
    (SELECT MAX(id) max_id, donation_id FROM donation_transactions GROUP BY donation_id) dt_max
...
EOSQL
    );
```

PR #2564, opened 2018-07-09
[Donation] Added alert feature for high amount

thomasbisignani:
- [X] **If the user makes a high donation, we propose him a monthly donation :**
capture d ecran de 2018-07-10 14-26-11

- [X] **This PR fixes the amount render on the donation form and the confirmation page :**

Before : 
capture d ecran de 2018-07-10 11-41-46

After : 
capture d ecran de 2018-07-10 11-37-35

- [X] **This PR adds also the monthly label on the retry form :** 

capture d ecran de 2018-07-10 11-45-47

### DIFF
--- a/src/Form/DonationRequestType.php
+++ b/src/Form/DonationRequestType.php
@@ -4,6 +4,7 @@ namespace AppBundle\Form;
 
 use AppBundle\Donation\DonationRequest;
 use AppBundle\Donation\DonationRequestUtils;
+use AppBundle\Donation\PayboxPaymentSubscription;
 use AppBundle\Entity\Adherent;
 use AppBundle\Form\DataTransformer\FloatToStringTransformer;
 use AppBundle\Membership\MembershipRegistrationProcess;
@@ -11,6 +12,7 @@ use AppBundle\Repository\AdherentRepository;
 use Misd\PhoneNumberBundle\Form\Type\PhoneNumberType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
 use Symfony\Component\Form\Extension\Core\Type\HiddenType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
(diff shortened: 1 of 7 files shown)


templates/form_theme.html.twig
```diff
@@ -228,3 +228,22 @@
...
     {% endif %}
 {% endblock %}
+
+{% block _app_donation_confirmDonationType_row %}
+    {{ form_errors(form.parent.children.confirmSubscriptionAmount) }}
+    {%- for child in form %}
+        <p style="text-align: left">
+            {{- form_widget(child) -}}
+
+            {% if child.vars.value == constant('AppBundle\\Form\\DonationRequestType::CONFIRM_DONATION_TYPE_UNIQUE') %}
```
Remg: i guess the condition can be inside the `<label>` tag

src/Form/DonationRequestType.php
```diff
@@ -57,6 +67,40 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                        $request->get('montant')
+                    ),
+                    'filter_emojis' => true,
+                    'attr' => ['size' => 2, 'maxlength' => 4],
+                ])
+                ->addEventListener(FormEvents::POST_SUBMIT, function (FormEvent $formEvent) {
+                    $form = $formEvent->getForm();
+
+                    if (self::CONFIRM_DONATION_TYPE_SUBSCRIPTION === $form->get('confirmDonationType')->getData()
+                        && $donationRequest = $formEvent->getData() ?? null
```
ottaviano: `?? null` is useless here, no ?

templates/form_theme.html.twig
```diff
@@ -228,3 +228,20 @@
...
 {% endblock %}
+
+{% block _app_donation_confirmDonationType_row %}
+    {{ form_errors(form.parent.children.confirmSubscriptionAmount) }}
+    {%- for child in form %}
+        <p style="text-align: left">
+            {{- form_widget(child) -}}
+             <label for="{{ child.vars.id }}">
+
+            {% if child.vars.value == constant('AppBundle\\Form\\DonationRequestType::CONFIRM_DONATION_TYPE_UNIQUE') %}
```
ottaviano: CS: indent for if, you are into <label> element

templates/form_theme.html.twig
```diff
@@ -228,3 +228,20 @@
         <a href="{{ path('mooc_get_file', { slug: form.vars.value.slug, extension: form.vars.value.extension }) }}" target="_blank">Télécharger le fichier</a>
     {% endif %}
 {% endblock %}
+
+{% block _app_donation_confirmDonationType_row %}
+    {{ form_errors(form.parent.children.confirmSubscriptionAmount) }}
+    {%- for child in form %}
+        <p style="text-align: left">
```
ottaviano: 1) can you replace `p` with `div` pls? it's more correct
2) you can use `class="text--left"` to align the text instead of `style` attribute.
If I need to apply a style to my element, I am looking for it first in `/front/style`, maybe it already exists one class for this style.

PR #2571, opened 2018-07-11
[Donation] Updated civility and removed phone fields

thomasbisignani:
- Updated gender status `Madame`, `Monsieur` instead of `Femme`, `Homme`, but keep the same values in bdd
- Removed phone fields

### DIFF
--- a/templates/donation/informations.html.twig
+++ b/templates/donation/informations.html.twig
@@ -36,7 +36,7 @@
                 Toutes ces informations sont obligatoires pour que votre don soit validé par la commission des comptes.
             </div>
             <div class="l__row l__row--h-stretch b__nudge--bottom-tiny form__radio-button">
-                {{ form_widget(form.gender, { attr: { class: 'register__form__gender' } }) }}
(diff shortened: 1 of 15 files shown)


templates/donation/informations.html.twig
```diff
@@ -33,7 +33,7 @@
                 Toutes ces informations sont obligatoires pour que votre don soit validé par la commission des comptes.
             </div>
             <div class="l__row l__row--h-stretch b__nudge--bottom-tiny form__radio-button">
-                {{ form_widget(form.gender, { attr: { class: 'register__form__gender' } }) }}
+                {{ form_widget(form.gender, { attr: { class: 'register__form__gender civility' } }) }}
```
ottaviano: you can use `b__nudge--bottom-small` 

https://github.com/EnMarche/en-marche.fr/blob/master/front/style/layout/_layout.scss#L355

thomasbisignani: Nice catch, thx !

PR #2572, opened 2018-07-11
Change canonical links on event invitation and registration pages

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/4607/modifier-les-canonicals-des-pages-invitation-et-inscription-pour-les-%C3%A9v%C3%A9nements


templates/citizen_action/attend.html.twig
```diff
@@ -1,5 +1,7 @@
 {% extends 'events/attend.html.twig' %}
 
+{% block canonical_url url('app_citizen_action_attend', {slug: citizen_action.slug}) %}
```
thomasbisignani: Just a micro CS, can you add whitespaces inside arrays ? `{ slug: citizen_action.slug }`
Same for other

PR #2592, opened 2018-07-17
Add link to replace autocomplete if address is unknown by google

ottaviano:


src/Form/DonationRequestType.php
```diff
@@ -117,7 +117,9 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ->add('cityName', TextType::class, [
                 'required' => false,
             ])
-            ->add('country', UnitedNationsCountryType::class)
+            ->add('country', UnitedNationsCountryType::class, [
+                'preferred_choices' => ['FR'],
```
HeahDude:
```diff
- 'FR'
+ Address::FRANCE
```
?

templates/donation/informations.html.twig
```diff
@@ -65,7 +65,12 @@
         <div>
             {{ form_errors(form.address) }}
 
-            <div class="form__row address-autocomplete"></div>
+            <div class="form__row">
+                <div class="address-autocomplete"></div>
+                <p class="text--smallest text--muted text--left visually-hidden" id="address-autocomplete-help-message">
```
HeahDude: If I force displaying this field client side, can I click on it (though bypassing the autocomplete)?

ottaviano: yes, you can do it, as you can disable JS and bypass the autocomplete, or inter unknown address and this message will appear automatically

HeahDude: Ok!

src/Form/DonationRequestType.php
ottaviano: fixed, thanks 👍

PR #2599, opened 2018-07-19
 Add webhook for mailchimp when user change email subscriptions

illusionOfParadise:
https://app.clubhouse.io/enmarche/story/3687/step-2-a-webhooks-notifs-vers-mailchimp-ou-mailjet-etq-adh%C3%A9rent-si-je-modifie-mes-notifications-cela-doit-se-r%C3%A9percuter-sur

It's ok that I change an old migration https://github.com/EnMarche/en-marche.fr/pull/2599/files#diff-72c8fe2840e5ed19a45b93778c5c3644R40, confirmed with @Remg.

### DIFF
--- a/features/bootstrap/RabbitMqContext.php
+++ b/features/bootstrap/RabbitMqContext.php
@@ -108,7 +108,8 @@ class RabbitMqContext implements Context
             );
 
             if (!$found) {
-                throw new \LogicException('Expected message not found');
+                $messagesString = implode("\n", array_map(function ($msg) { return $msg->getBody(); }, $messages));
+                throw new \LogicException("Expected message not found among these ones:\n$messagesString");
             }
         }
     }
--- a/src/Form/WebHookType.php
+++ b/src/Form/WebHookType.php
@@ -4,6 +4,7 @@ namespace AppBundle\Form;
 
 use AppBundle\Entity\WebHook\WebHook;
 use AppBundle\WebHook\Event;
+use AppBundle\WebHook\Service;
 use Sonata\AdminBundle\Form\Type\CollectionType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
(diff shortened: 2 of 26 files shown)


src/Form/WebHookType.php
```diff
@@ -29,6 +30,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'by_reference' => false,
                 'error_bubbling' => false,
             ])
+            ->add('service', ChoiceType::class, [
+                'required' => false,
+                'placeholder' => '--',
+                'choices' => Service::toArray(),
+                'choice_label' => function ($value) {return $value; },
```
HeahDude: CS:
```diff
- 'choice_label' => function ($value) {return $value; },
+ 'choice_label' => function ($value) { return $value; },
```

src/WebHook/Event.php
```diff
@@ -9,10 +9,12 @@
...
  * @method static USER_MODIFICATION()
  * @method static USER_DELETION()
+ * @method static USER_UPDATE_SUBSCRIPTIONS()
  */
 class Event extends Enum
 {
     public const USER_CREATION = UserEvents::USER_CREATED;
     public const USER_MODIFICATION = UserEvents::USER_UPDATED;
     public const USER_DELETION = UserEvents::USER_DELETED;
+    public const USER_UPDATE_SUBSCRIPTIONS = UserEvents::USER_UPDATE_SUBSCRIPTIONS;
```
HeahDude: When is this used?

HeahDude: IMO declaring the constant in `UserEvents` is useless, it should be done here directly.
I can see that you follow the path taken before, but I think we should either use enums, or not.
Using both final classes and enum does not make any sense.

features/bootstrap/RabbitMqContext.php
```diff
@@ -108,7 +108,11 @@ function (bool $carry, AMQPMessage $message) use ($expectedHash): bool {
             );
 
             if (!$found) {
-                throw new \LogicException('Expected message not found');
+                $strMessages = '';
```
HeahDude: Messages looks glued without space using that code 🤔, though unreadable. I would do it like this:
```php
$messagesString = implode("\n", array_map(function ($msg) { return $msg->getBody(); }, $messages));

throw new \LogicException("Expected message not found among these ones:\n$messagesString");
```
WDYT?

PR #2600, opened 2018-07-19
Remove useless comment

ottaviano:
https://twitter.com/Stof70/status/1019515955083403264  🎉 🎉 🎉

### DIFF
--- a/src/Admin/Extension/AlgoliaIndexedEntityAdminExtension.php
+++ b/src/Admin/Extension/AlgoliaIndexedEntityAdminExtension.php
@@ -17,7 +17,6 @@ class AlgoliaIndexedEntityAdminExtension extends AbstractAdminExtension
     }
 
     /**
-     * @param AdminInterface                $admin
      * @param AlgoliaIndexedEntityInterface $object
      */
     public function postPersist(AdminInterface $admin, $object)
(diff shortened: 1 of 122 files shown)


src/Admin/Extension/AlgoliaIndexedEntityAdminExtension.php
```diff
@@ -17,7 +17,6 @@ public function __construct(AlgoliaIndexedEntityManager $manager)
     }
 
     /**
-     * @param AdminInterface                $admin
      * @param AlgoliaIndexedEntityInterface $object
```
HeahDude: Should be removed and added in the signature (same for other occurrences)

src/CitizenProject/CitizenProjectFactory.php
```diff
@@ -91,8 +87,6 @@ public function createFromCitizenProjectCreationCommand(CitizenProjectCreationCo
      * The format must be something like "33 0102030405"
      *
      * @param string $phoneNumber
```
HeahDude: Should be type hinted in the signature instead.

src/Entity/CitizenProject.php
```diff
@@ -507,7 +507,6 @@ public static function createForAdherent(
     /**
      * Marks this citizen project as approved.
      *
-     * @param string $timestamp
      *
```
HeahDude: this empty line can be removed too (should be fixed in the fixer btw)

src/CitizenProject/CitizenProjectFactory.php
HeahDude: (many occurrences)

thomasbisignani:
:1st_place_medal:

PR #2601, opened 2018-07-19
Execute runManageParticipants only if host to avoid JS error

ottaviano:
image


templates/citizen_action/list_participants.html.twig
```diff
@@ -2,17 +2,20 @@
 
 {% block page_title 'Participants à l\'action citoyenne' %}
 
+{% set is_host = is_citizen_project_administrator(citizen_action.citizenProject) %}
```
HeahDude: You can use `is_administrator` from the inherited layout.

PR #2608, opened 2018-07-25
Improve subscription type migration command

illusionOfParadise:
It's a one-shot command.

### DIFF
--- a/src/Command/MigrateSubscriptionTypeCommand.php
+++ b/src/Command/MigrateSubscriptionTypeCommand.php
@@ -5,29 +5,37 @@ namespace AppBundle\Command;
 use AppBundle\Entity\Adherent;
 use AppBundle\Entity\CitizenProjectMembership;
 use AppBundle\Entity\District;
-use AppBundle\Repository\AdherentRepository;
-use AppBundle\Repository\SubscriptionTypeRepository;
+use AppBundle\Entity\SubscriptionType;
 use AppBundle\Subscription\SubscriptionTypeEnum;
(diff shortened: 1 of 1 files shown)


src/Command/MigrateSubscriptionTypeCommand.php
```diff
@@ -5,71 +5,121 @@
...
+                    $adherentTypes[] = SubscriptionTypeEnum::CITIZEN_PROJECT_HOST_EMAIL;
+                }
+
+                if (array_diff($adherentTypes, $adherent->getEmailsSubscriptions())) {
+                    foreach ($adherentTypes as $type) {
+                        $adherent->addSubscriptionType($subscriptionTypes[$type]);
+                    }
+                }
+
+                if (0 === $index % 500) {
```
Remg: Can we make this batch size a constant and increase it to 1000 maybe?

src/Command/MigrateSubscriptionTypeCommand.php
```diff
@@ -5,71 +5,121 @@
...
+                        $adherent->addSubscriptionType($subscriptionTypes[$type]);
+                    }
+                }
+
+                if (0 === $index % 1000) {
+                    $this->manager->flush();
+                    $this->manager->clear(Adherent::class);
+                    $this->manager->clear(CitizenProjectMembership::class);
+                    $this->manager->clear(District::class);
+                    $progressBar->advance(500);
```
HeahDude: Shouldn't this value be changed accordingly?

PR #2610, opened 2018-07-26
[MOOC] Added Sitemap API Endpoint

thomasbisignani:


src/Controller/Api/MoocController.php
```diff
@@ -34,4 +46,12 @@ public function moocAction(Mooc $mooc, MoocNormalizer $normalizer): Response
             true
         );
     }
+
+    private function createXmlResponse(string $content): Response
+    {
+        $response = new Response($content);
+        $response->headers->set('Content-Type', 'text/xml');
```
HeahDude: You can use the third argument, thus return the object directly.

PR #2630, opened 2018-08-13
Improve SEO of CP actors list page

ottaviano:


templates/citizen_project/list_actors.html.twig
```diff
@@ -1,6 +1,7 @@
 {% extends 'citizen_project/_layout.html.twig' %}
 
-{% block page_title 'Acteurs du projet citoyen' %}
+{% block page_title 'Acteurs du projet citoyen ' ~ citizen_project.name %}
+{% block meta_description 'Découvrez les ' ~ followers|length ~ ' acteurs de ' ~ citizen_project.name ~ ', une action créée dans le cadre des projets citoyens de La République En Marche ! et portée à ' ~ citizen_project.cityName ~ '.' %}
```
thomasbisignani: IMO you can use the string interpolation here (  `#{variable}` )

ottaviano: done 👍

PR #2631, opened 2018-08-13
Remove unused uuid param from URL

ottaviano:


ottaviano:
I will do it in other PR

PR #2641, opened 2018-08-16
Add command for sending activation email to unconfirmed users

ottaviano:


app/config/services/services.xml
```diff
@@ -94,6 +94,7 @@
         <service id="AppBundle\Donation\PayboxFormFactory" />
 
         <!-- Membership -->
+        <service id="AppBundle\Command\RemindUnconfirmedMembershipCommand"/>
```
thomasbisignani: You can declare this service in the `<!--Command-->` section, what do you think ?

src/Command/RemindUnconfirmedMembershipCommand.php
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    /**
+     * @return Adherent[]
+     */
+    private function findUnconfirmed(\DateTime $from, int $limit): array
+    {
+        return $this->adherentRepository
+            ->createQueryBuilder('a')
```
thomasbisignani: Not a big deal, but `adherent` instead of an alias `a` is little more readable

ottaviano: agree 👍 , fixed

PR #2642, opened 2018-08-16
Change TurnkeyProject endpoints + add position

illusionOfParadise:


features/api/turnkey_projects.feature
```diff
@@ -28,30 +28,34 @@ Feature:
       "category":"Nature et Environnement",
       "title":"Stop mégots !",
       "slug":"stop-megots",
-      "subtitle":"Campagnes de sensibilisation et de revalorisation des mégots jetés"
+      "subtitle":"Campagnes de sensibilisation et de revalorisation des mégots jetés",
+      "position": 1
```
ottaviano: why do we need to have `position` field in the response, the ordered list (without position field) is not enough?

src/Controller/Api/TurnkeyProjectController.php
```diff
@@ -13,59 +13,62 @@
...
-            [],
-            true
+            ['total' => $repository->countApproved()],
+            JsonResponse::HTTP_OK
         );
     }
 
     /**
-     * @Route("/turnkey-projects/count", name="api_count_approved_turnkey_projects")
+     * @Route("/is-pinned", name="api_pinned_turnkey_project")
```
ottaviano: 1) what about just `/pinned`?
2) `api_pinned_turnkey_project` => `api_pinned_turnkey_projectS` 
3) is the list of pinned projects ordered?

illusionOfParadise: 1. Why not `is-pinned`?
2. No, there is only one pinned turnkey project, it's not possible to have more than one.
3. Same that 2.

features/api/turnkey_projects.feature
illusionOfParadise: Yes, there is no need of position. I removed it.

src/Controller/Api/TurnkeyProjectController.php
```diff
@@ -13,59 +13,62 @@
...
      * @Method("GET")
      */
-    public function countApprovedTurnkeyProjectAction(TurnkeyProjectRepository $repository): Response
+    public function getPinnedTurnkeyProjectAction(TurnkeyProjectRepository $repository, Serializer $serializer): Response
     {
+        $turnkeyProject = $repository->findPinned();
```
thomasbisignani: This variable is useless, it is used once

src/Controller/Api/TurnkeyProjectController.php
thomasbisignani: I agree with the naming `/pinned`, it is more in accordance with the action name, and `is-pinned` make us think boolean logic.

PR #2659, opened 2018-09-04
[MOOC] Added Landing page API endpoint

thomasbisignani:


src/Controller/Api/MoocController.php
```diff
@@ -19,6 +21,22 @@
...
 {
+    /**
+     * @Route("", name="api_mooc_landing")
+     * @Method("GET")
+     */
+    public function moocLandingPageAction(MoocRepository $moocRepository, MoocListNormalizer $normalizer): Response
+    {
+        $serializer = new Serializer([$normalizer], [new JsonEncoder()]);
+
+        return new JsonResponse(
```
ottaviano: why do you not use `return $this->json()` with `@Groups()` annotation for your 4 properties in Mooc object without creating a specific Normaliser?

thomasbisignani: I created this Normalizer to handle the order of the fields but ok ;-)

PR #2666, opened 2018-09-06
Add some visual changes on the CA page

illusionOfParadise:


src/CitizenAction/CitizenActionManager.php
```diff
@@ -84,10 +91,20 @@ public function populateRegistrationWithAdherentsInformations(EventRegistrationC
                 'postalCode' => $adherentFind->getPostalCode(),
                 'cityName' => $adherentFind->getCityName(),
                 'createdAt' => $registration->getCreatedAt(),
+                'administrator' => false,
```
ottaviano: why do you not use `'administrator' => $adherent->isAdministratorOf($citizenProject),`?

tests/Repository/EventRepositoryTest.php
```diff
@@ -63,6 +64,20 @@ public function testSearchAllEvents()
         $search = $this->get(SearchParametersFilter::class)->handleRequest($request);
 
         $this->assertSame(1, count($this->repository->searchAllEvents($search)));
+
+        http://enmarche.code/recherche?q=&r=50&c=Paris%208e,%20France&t=citizen_actions&offset=0&ec=citizen_actions
```
ottaviano: need remove?

src/Entity/EventRegistration.php
```diff
@@ -144,4 +144,11 @@ public function matches(BaseEvent $event, Adherent $adherent = null): bool
 
         return $adherent && $this->adherentUuid->equals($adherent->getUuid());
     }
+
+    public function getLastNameInitial(): string
```
ottaviano: what about using `EntityPersonNameTrait` in `EventRegistration` too for avoiding CC this method?

app/config/config.yml
```diff
@@ -229,6 +229,7 @@ sonata_doctrine_orm_admin:
         types:
             list:
                 array: admin/CRUD/list_array.html.twig
+                datetime: admin/CRUD/list_datetime.html.twig
```
ottaviano: is this type really used?

illusionOfParadise: Yes, for all date representation in admin, it was added to change a format and show month in french.

src/CitizenAction/CitizenActionManager.php
illusionOfParadise: Citizen Project is not accessible in this method.

ottaviano: not false, but it can be, no?
```php
function populateRegistrationWithAdherentsInformations(..., CitizenProject $cp)
```

illusionOfParadise: What is the advantage to add an argument to the method and do an additional heavy call https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/Adherent.php#L945 for each iteration?

src/Entity/EventRegistration.php
```diff
@@ -17,8 +17,9 @@
  */
 class EventRegistration
 {
-    use EntityIdentityTrait;
     use EntityCrudTrait;
+    use EntityIdentityTrait;
+    use EntityPersonNameTrait;
```
ottaviano: so you can remove `private $firstName` and `$lastName` from EventRegistration?

PR #2677, opened 2018-09-12
Show text not form field for some fields  on the CitizenProject creation page when creating from Turnkey project

illusionOfParadise:


templates/citizen_project/_form.html.twig
```diff
@@ -13,37 +13,53 @@
...
-    {{ form_widget(form.name, {
-        attr: {
-            'class': 'form--full',
-            'placeholder': 'Un nom court et clair',
-        }
-    }) }}
-    <small>Ex: "Un stage pour tous"</small>
+    {% if form.name.vars.disabled %}
+        <div id="citizen_project_name_text">{{ form.vars.data.name }}</div>
+        <div class="visually-hidden">{{ form_widget(form.name) }}</div>
```
ottaviano: do we still need to render the form element here?

templates/citizen_project/_form.html.twig
```diff
@@ -69,33 +85,43 @@
...
-            'class': 'form--full form__field--textarea-120',
-            'placeholder': 'Un court paragraphe sur la ou les solutions que vous proposez',
-        }
-    }) }}
-    <small>
-        Ex: Le projet a pour objectif de mettre en relation ces élèves avec un réseau de professionnels
-        (entreprises, commerçants, artisans, etc.) volontaires pour les accueillir.
-    </small>
+    {% if form.proposed_solution.vars.disabled %}
+        <div id="citizen_project_proposed_solution_text">{{ form.vars.data.proposedSolution|raw }}</div>
```
ottaviano: why `|raw` filter?

templates/citizen_project/_form.html.twig
```diff
@@ -13,37 +13,53 @@
...
     {{ form_label(form.name, 'Donnez un nom à votre projet :', {label_attr: {class: 'label__header--large'}}) }}
-    {{ form_errors(form.name) }}
-    {{ form_widget(form.name, {
-        attr: {
-            'class': 'form--full',
-            'placeholder': 'Un nom court et clair',
-        }
-    }) }}
-    <small>Ex: "Un stage pour tous"</small>
+    {% if form.name.vars.disabled %}
```
ottaviano: you can use the getter `isDisabled()` of form element 
`if form.name.disabled` or `if form.name.isDisabled`

templates/citizen_project/_form.html.twig
illusionOfParadise: Yes, because this field contains the HTML tags. Its value was entered in the CKEditor and we need to display a formatted text.

templates/citizen_project/_form.html.twig
illusionOfParadise: It does not work, because it's a `FormView` class which have no method `isDisabled()`.

templates/citizen_project/_form.html.twig
illusionOfParadise: Yes, if no, they will be rendered with `form_rest()` of parent templates.

templates/adherent/create_citizen_project.html.twig
```diff
@@ -59,7 +59,11 @@
...
-                                        <label class="block" for="citizen_project_data_processing">J’adhère aux <a href="https://storage.googleapis.com/en-marche-prod/documents/projets-citoyens/CGU%20-%20Projets%20Citoyens.pdf" target="_blank" rel="noopener" class="link--blue">Conditions Générales d’Utilisation des Projets citoyens</a> et à la <a href="https://storage.googleapis.com/en-marche-prod/documents/projets-citoyens/La%20Charte%20des%20Projets%20Citoyens.pdf" target="_blank" class="link--blue">Charte des Projets citoyens</a></label>
+                                        <label class="block" for="citizen_project_data_processing">
+                                            <span>
+                                                J’adhère aux <a href="https://storage.googleapis.com/en-marche-prod/documents/projets-citoyens/CGU%20-%20Projets%20Citoyens.pdf" target="_blank" rel="noopener" class="link--blue">Conditions Générales d’Utilisation des Projets citoyens</a> et à la <a href="https://storage.googleapis.com/en-marche-prod/documents/projets-citoyens/La%20Charte%20des%20Projets%20Citoyens.pdf" target="_blank" class="link--blue">Charte des Projets citoyens</a>
```
thomasbisignani: Can you change the `’` char for `'` ?

templates/citizen_project/_form.html.twig
ottaviano: oh yes, you are right 👍

PR #2678, opened 2018-09-12
[Donation] Added country value by geolocation

thomasbisignani:


front/services/address/GooglePlaceAutocomplete.js
```diff
@@ -120,13 +120,15 @@ export default class GooglePlaceAutocomplete extends EventEmitter {
     }
 
     resetState() {
+        const defaultCountry = dom('select[name*=country] option[selected]');
```
ottaviano: We already have this value in `this._address.country` property

To avoid reselect the `<select>` dom element I propose 
1) to change update country field logic in `updateFields` method:
```js
// BEFORE
// this._address.setCountry(this._state.country && this._state.country.short_name || '');

// AFTER
if (this._state.country && this._state.country.short_name) {
    this._address.setCountry(this._state.country.short_name);
} else if (!this._address.getCountry()) {
    this._address.setCountry('FR');
}
```

2) but need set to null state.country value
```js
this._state = {
    street_number: null,
    route: null,
    locality: null,
    sublocality_level_1: null,
    postal_code: null,
    country: null,
};
```

3) and add the country getter method to AddressObject

what do you think about?

thomasbisignani: Nice catch ! thx

PR #2680, opened 2018-09-13
Remove a newsletter subscription when an unsubscription come from Mailchimp

illusionOfParadise:


src/Subscription/SubscriptionHandler.php
```diff
@@ -31,20 +36,22 @@ public function changeSubscription(string $type, string $email, string $listId):
...
+            if (self::ACTION_TYPE_SUBSCRIBE === $type && !$hasSubscription) {
+                $adherent->addSubscriptionType($subscriptionType);
+            } elseif (self::ACTION_TYPE_UNSUBSCRIBE === $type && $hasSubscription) {
+                $adherent->removeSubscriptionType($subscriptionType);
+            }
+        } elseif (($newsletterSubscription = $this->newsletterSubscriptionRepository->findOneByEmail($email))
+            && self::ACTION_TYPE_UNSUBSCRIBE === $type) {
+            // Newsletter subscription will remain in the table but with filled `deleted_at`field
+            $this->em->remove($newsletterSubscription);
         }
```
ottaviano: can we `flush()` here instead of in the controller? what do you think?

PR #2691, opened 2018-09-17
[Event] move report event link

JeremieSamson:
Deplacer "Signaler cet événement" et le mettre en bleu clair

### DIFF
--- a/templates/events/_header.html.twig
+++ b/templates/events/_header.html.twig
@@ -20,11 +20,6 @@
                 </div>
             {% endif %}
         </div>
-        <div class="b__nudge--top-40">
-            {% if is_granted('REPORT') %}
-                <a href="{{ report_path(event, app.request.pathInfo) }}" class="text--body link--red">Signaler cet événement</a>
-            {% endif %}
-        </div>
     </div>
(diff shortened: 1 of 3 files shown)


thomasbisignani:
:1st_place_medal:

front/style/pages/_committee.scss
```diff
@@ -471,6 +471,10 @@
         }
     }
 
+    .box-right {
```
ottaviano: can we add this general rule to `_layout.scss` file with `float-right` as the name?

JeremieSamson: We had the same thought but @thomasbisignani told me there is a pull-righ / pull-left from font awesome css we should use as it already exists (next commit).

ottaviano: hm, do we use FA?

JeremieSamson: yes sir => https://github.com/EnMarche/en-marche.fr/blob/master/templates/base.html.twig#L97

ottaviano: lol, thanks, I don't find it in package.json

PR #2694, opened 2018-09-17
Sort CP list of the adherent and add colors

illusionOfParadise:


src/Entity/Adherent.php
```diff
@@ -889,12 +889,18 @@ public function getMembershipFor(Committee $committee): ?CommitteeMembership
...
 
-    public function getCitizenProjectMemberships(): CitizenProjectMembershipCollection
+    public function getCitizenProjectMemberships($withoutRefused = false): CitizenProjectMembershipCollection
     {
         if (!$this->citizenProjectMemberships instanceof CitizenProjectMembershipCollection) {
             $this->citizenProjectMemberships = new CitizenProjectMembershipCollection($this->citizenProjectMemberships->toArray());
         }
 
+        if ($withoutRefused) {
+            $this->citizenProjectMemberships->filter(function (CitizenProjectMembership $membership) {
```
ottaviano: 1) `return` missing
2) maybe add `filterRefusedProject` method to `CitizenProjectMembershipCollection` class?

templates/adherent/list_my_administrated_citizen_projects.html.twig
```diff
@@ -1,12 +1,45 @@
...
-            </li>
-        {% endif %}
-    {% endfor %}
-</ul>
+{% if citizen_projects|length > 1 %}
+    <div class="nav-dropdown">
+        <i class="fa fa-caret-down" aria-hidden="true"></i>
+        <i class="fa fa-caret-up" aria-hidden="true"></i>
+        <ul class="list__links">
+            {% for citizen_project in citizen_projects|slice(0, 1) %}
```
ottaviano: what about fusion `for` with `if` instructions
```twig
{% if foo in bar if foo.isTrue() %}
```

src/Entity/Adherent.php
```diff
@@ -889,12 +889,16 @@ public function getMembershipFor(Committee $committee): ?CommitteeMembership
...
 
-    public function getCitizenProjectMemberships(): CitizenProjectMembershipCollection
+    public function getCitizenProjectMemberships($withoutRefused = false): CitizenProjectMembershipCollection
     {
         if (!$this->citizenProjectMemberships instanceof CitizenProjectMembershipCollection) {
             $this->citizenProjectMemberships = new CitizenProjectMembershipCollection($this->citizenProjectMemberships->toArray());
         }
 
+        if ($withoutRefused) {
+            $this->citizenProjectMemberships = $this->citizenProjectMemberships->filterRefusedProjects();
```
ottaviano: hm, we are into the getter (not setter), why do you modify the initial value?

PR #2695, opened 2018-09-17
Added country filter for committees during the adhesion process

thomasbisignani:


tests/Committee/CommitteeManagerTest.php
```diff
@@ -103,6 +104,34 @@ public function testGetNearbyCommittees()
...
+
+        $this->assertCount($expectedCount, $committees);
+
+        /** @var Committee $committee */
+        foreach ($committees as $committee) {
+            $this->assertSame($adherent->getCountry(), $committee->getCountry());
+        }
+    }
+
+    public function provideCommitteesByCoordinatesAndCountry(): iterable
```
illusionOfParadise: The name is not consistent with the provided data. It will be more explicit to name `provideAdherentUuidsAndCommitteesCount`. What do you think?

PR #2702, opened 2018-09-19
Create a separate table for District GeoData

illusionOfParadise:


src/Entity/District.php
```diff
@@ -162,20 +149,20 @@ public function getDepartmentCode(): string
...
+     * @return GeoData
      */
-    public function getGeoShape(): GeometryInterface
+    public function getGeoData(): GeoData
     {
-        return $this->geoShape;
+        return $this->geoData;
     }
 
     public function getPolygons(): array
```
ottaviano: is this method used somewhere?

app/migrations/Version20180918171508.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
```
ottaviano: to remove?

thomasbisignani: Can you remove this line ?

PR #2708, opened 2018-09-20
Add files to Turnkey Projects

Remg:


src/Entity/TurnkeyProject.php
```diff
@@ -175,6 +177,18 @@ class TurnkeyProject
      */
     private $position = 1;
 
+    /**
+     * @var Collection|TurnkeyProjectFile[]
+     *
+     * @ORM\ManyToMany(targetEntity="TurnkeyProjectFile", cascade={"persist"}, orphanRemoval=true)
+     * @ORM\JoinTable(name="turnkey_projects_has_files")
```
ottaviano: why the default value `turnkey_projects_files` is not OK?
the `has` in table name is not really meaningful.

src/Entity/TurnkeyProject.php
```diff
@@ -175,6 +177,18 @@ class TurnkeyProject
...
+     * @ORM\OrderBy({"id": "ASC"})
```
ottaviano: it's the order by default, no?

thomasbisignani:
:1st_place_medal:

PR #2709, opened 2018-09-20
[Biography] Added organization

thomasbisignani:
This PR improves the old organization page and allows to manage the members via the BO.

### DIFF
--- a/app/config/services/admin.xml
+++ b/app/config/services/admin.xml
@@ -610,6 +610,19 @@
 
         <!--
         ######################################################
+        # Biography
+        ######################################################
+        -->
+        <service id="app.admin.biography_executive_office_member" class="AppBundle\Admin\ExecutiveOfficeMemberAdmin">
+            <tag name="sonata.admin" manager-type="orm" label="Membres BurEx" group="Biographies" />
+
+            <argument />
+            <argument>AppBundle\Entity\Biography\ExecutiveOfficeMember</argument>
+            <argument />
+        </service>
+
+        <!--
+        ######################################################
         # OAuth
         ######################################################
          -->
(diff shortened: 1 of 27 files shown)


app/config/services/admin.xml
```diff
@@ -609,6 +609,19 @@
             </call>
         </service>
 
+        <!--
+        ######################################################
+        # Biography
+        ######################################################
+        -->
+        <service id="app.admin.biography_executive_office_member" class="AppBundle\Admin\ExecutiveOfficeMemberAdmin">
```
Remg: can we use the class as id ?

src/Controller/EnMarche/BiographyController.php
```diff
@@ -0,0 +1,23 @@
...
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+
+class BiographyController extends Controller
+{
+    /**
+     * @Route("/le-mouvement/notre-organisation", name="app_our_organization")
+     * @Method("GET")
+     */
+    public function executiveOfficeMemberListAction(ExecutiveOfficeMemberRepository $repository)
```
Remg: can you add a return type?

tests/Controller/EnMarche/PageControllerTest.php
```diff
@@ -55,7 +55,6 @@ public function providePages()
         yield ['/emmanuel-macron/revolution'];
         yield ['/emmanuel-macron/videos'];
         yield ['/le-mouvement'];
-        yield ['/le-mouvement/notre-organisation'];
```
Remg: why remove this line?

app/config/services/admin.xml
thomasbisignani: Sonata needs this syntax for the security IMO.

PR #2712, opened 2018-09-20
Sort on user committees and add colors

JeremieSamson:


src/Collection/CommitteeCollection.php
```diff
@@ -24,10 +24,16 @@ public function getOrderedCommittees(CommitteeMembershipCollection $memberships)
             return in_array((string) $committee->getUuid(), $uuids);
         });
 
+        $uuids = $memberships->getCommitteeUnApprovedMemberships(CommitteeMembershipCollection::EXCLUDE_SUPERVISORS)->getCommitteeUuids();
```
Remg: is the argument useless or not handled in the method ?
`public function getCommitteeUnApprovedMemberships(): self`

src/Collection/CommitteeMembershipCollection.php
```diff
@@ -66,7 +66,7 @@ public function getCommitteeHostMemberships(int $withSupervisors = self::INCLUDE
     public function getCommitteeFollowerMemberships(): self
     {
         return $this->filter(function (CommitteeMembership $membership) {
-            return $membership->isFollower();
+            return $membership->isFollower() && !$membership->getCommittee()->isPending();
```
thomasbisignani: What do you think to rename the `getCommitteeFollowerMemberships`method to `getCommitteeFollowerMembershipsInProcess` ?

src/Collection/CommitteeMembershipCollection.php
```diff
@@ -77,6 +77,13 @@ public function getCommitteeSupervisorMemberships(): self
         });
     }
 
+    public function getCommitteeUnapprovedMemberships(): self
```
thomasbisignani: What do you think to rename the `getCommitteeUnapprovedMemberships`method to `getCommitteeMembershipsInPending` ?

illusionOfParadise: @JeremieSamson Are you sure that we need a list of only pending committees without pre-approved and pre-refused ones?

src/Collection/CommitteeMembershipCollection.php
illusionOfParadise: @thomasbisignani the actual name seems good. It's the opposite, we check that the committee status is not pending.

src/Collection/CommitteeMembershipCollection.php
Remg: is it confirmed ? @JeremieSamson

PR #2713, opened 2018-09-20
Demote hosts & supervisor on source committee during merge

Remg:


src/Committee/CommitteeMergeCommandHandler.php
```diff
@@ -56,6 +58,12 @@ public function handle(CommitteeMergeCommand $committeeMergeCommand): void
                 $this->em->persist($this->createCommitteeMembershipHistory($membership));
             }
 
+            foreach ($sourceHostsMemberships as $sourceHostMembership) {
```
ottaviano: we can do this with this way, no? :

```
$repository
    ->findHostMemberships($sourceCommittee)
    ->map(function (Membership $m) {
        $m->setPrivilege(CommitteeMembership::COMMITTEE_FOLLOWER);
    })
;
```

Remg: yes, it would be a shortcut to write this.
but then the 'select' statement would be inside the transaction.

the way i handle this is: select's statements, begin transaction, update statements,  commit/rollback

what do you think?

ottaviano: SELECT in a transaction can select inserted/updated rows in the current transaction. It's not your use case.

PR #2717, opened 2018-09-22
Create React apps system

tgalopin:


tgalopin:
@Remg I think this is ready.

PR #2720, opened 2018-09-24
Fix CP variable name

ottaviano:
https://sentry.io/en-marche-i7/en-marchefr/issues/699173001/?referrer=slack


thomasbisignani:
:man_cartwheeling:

PR #2723, opened 2018-09-25
Fix CommitteeManager deprecations

Remg:
```User Deprecated: The "AppBundle\Committee\CommitteeManager" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.```

### DIFF
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -35,6 +35,10 @@ services:
         arguments:
             - '%env(SSL_PRIVATE_KEY)%'
 
+    app.committee.manager:
+        alias: 'AppBundle\Committee\CommitteeManager'
+        public: true
+
 parameters:
     env(DATABASE_NAME): "enmarche_test"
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by thomasbisignani** [comment]
:ok_hand:

File: app/config/config_test.yml [hunk 1]
```diff
@@ -35,6 +35,10 @@ services:
         arguments:
             - '%env(SSL_PRIVATE_KEY)%'
 
+    app.committee.manager:
```
ottaviano: why is it public? :D

Remg: because tests.
https://github.com/EnMarche/en-marche.fr/blob/master/tests/Controller/EnMarche/CommitteeManagerControllerTest.php#L601

ottaviano:
It is no longer relevant with Sf4.4 merged recently

PR #2726, opened 2018-09-25
[Biography] Added show view

thomasbisignani:


src/Admin/ExecutiveOfficeMemberAdmin.php
```diff
@@ -167,10 +165,8 @@ protected function configureListFields(ListMapper $listMapper)
                 'label' => 'Dernière mise à jour',
             ])
             ->add('_action', null, [
-                'actions' => [
-                    'edit' => [],
-                    'delete' => [],
-                ],
+                'virtual_field' => true,
```
ottaviano: you can keep `edit` and `delete` actions and add you custom action (link) with:
```
->add('_action', null, [
  'actions' => [
    'link' => [
      'template' => 'template_path.html.twig', // with only one link
    ],
    'edit' => [],
    'delete' => [],
  ],
])
```

templates/page/le-mouvement/notre-organisation.html.twig
```diff
@@ -38,50 +38,62 @@
...
-        </div>
-    {% endif %}
-
-    {% if (executiveOfficeMembers) %}
-        <div id="executive-office-members">
-            <h2 class="text--center">Le Bureau Exécutif</h2>
-            <ul class="team">
-                {% for member in executiveOfficeMembers %}
+    <div id="biography">
+        {% if (executiveOfficer) %}
```
ottaviano: you can remove the `(` & `)` for `if` on twig

templates/social_share/social_networks.html.twig
```diff
@@ -0,0 +1,12 @@
+{% if object.facebookProfile %}
+    <a class="link--no-decor" href="{{ object.facebookProfile }}" target="_blank"><i class="fa fa-facebook"></i></a>
```
ottaviano: what about adding `@AssertUrl` to these properties?

front/style/content/_movement.scss
```diff
@@ -166,14 +166,30 @@
     width: 100%;
 }
 
-#executive-officer .team {
-    justify-content: center !important;
+#biography {
+    .executive-officer {
+        .team {
+            justify-content: center !important;
```
JeremieSamson: Why do you use !important ?

thomasbisignani: It's an override of this property : https://github.com/EnMarche/en-marche.fr/blob/master/front/style/content/_movement.scss#L51

PR #2733, opened 2018-09-28
[Citizen Projects] Fixed members count for meta description

thomasbisignani:


templates/citizen_project/list_actors.html.twig
```diff
@@ -1,7 +1,7 @@
 {% extends 'citizen_project/_layout.html.twig' %}
 
 {% block page_title "Acteurs du projet citoyen #{citizen_project.name}" %}
-{% block meta_description "Découvrez les #{citizen_project.membersCounts|length} acteurs de #{citizen_project.name}, une action créée dans le cadre des projets citoyens de La République En Marche ! et portée à #{citizen_project.cityName}." %}
+{% block meta_description "Découvrez les #{citizen_project.membersCounts} acteurs de #{citizen_project.name}, une action créée dans le cadre des projets citoyens de La République En Marche ! et portée à #{citizen_project.cityName}." %}
```
ottaviano: `.membersCount` without `s`, no?

thomasbisignani: I agree :+1:

The property is defined with `s` : 
```php    
/**
     * The cached number of members (followers and hosts/administrators).
     *
     * @ORM\Column(type="smallint", options={"unsigned": true})
     *
     * @Algolia\Attribute
     *
     * @JMS\Groups({"public", "committee_read", "citizen_project_read"})
     * @JMS\SerializedName("membersCount")
     */
    protected $membersCounts;
```

The getter without : 
```php
    public function getMembersCount(): int
    {
        return $this->membersCounts;
    }
```
It is used with the `s` a lot of times in this project, i can use it without `s` here and fix all others in another PR.

PR #2736, opened 2018-10-01
Renamed BaseGroup members count property

thomasbisignani:


src/Entity/BaseGroup.php
```diff
@@ -72,7 +72,7 @@
      * @JMS\Groups({"public", "committee_read", "citizen_project_read"})
      * @JMS\SerializedName("membersCount")
```
Remg: this annotation is useless now ?

PR #2740, opened 2018-10-03
Fix css error on checking checkbox

JeremieSamson:


front/style/pages/_citizen-projects.scss
```diff
@@ -290,10 +290,12 @@
         padding: 10px;
     }
 
-    &-form {
+    &__form {
```
illusionOfParadise: There are templates that continue to use the old class `citizen-project-form`. You should modify them.

PR #2746, opened 2018-10-04
Remove menu in the deputy space and change deputy mail

illusionOfParadise:


templates/deputy/_layout.html.twig
```diff
@@ -16,33 +16,6 @@
             </header>
 
             <div class="l__wrapper procuration-manager__content">
-                {% set pages = [
```
ottaviano: do we need to block the controller actions too?

illusionOfParadise: Currently in the Controller we use canary mode and these actions are blocked in the prod.
When we will open deputy space for prod we will keep the canary mode for these cached actions.

illusionOfParadise: It's possible that actions will be used lately.

ottaviano: 👌

PR #2750, opened 2018-10-08
Add solution in REACT turnkey project detail

JeremieSamson:


web/apps/projets-citoyens/src/js/components/TurnkeyProjectDetail.js
```diff
@@ -45,7 +46,11 @@ const TurnkeyProjectDetail = props => {
                 )}
             </div>
             {!swap && <Title title={title} subtitle={subtitle} />}
-            <p style={{ width: `${props.textSize}` }}>{description}</p>
+            <p style={{ width: `${props.textSize}` }}>
+                {description}
+                <br/><br/>
+                {solution}
```
ottaviano: what about add a new `p` element with solution's text?

JeremieSamson: PO wanted less margin between texts when I proposed to have "solution" in another p element. 
Should I still use it and play with css or leave it as it is @ottaviano ?

ottaviano: 👌

PR #2751, opened 2018-10-09
Add thumbnail image on mooc

JeremieSamson:
Permettre l'upload d'une vignette lors de la création d'un MOOC

### DIFF
--- a/src/Admin/MoocAdmin.php
+++ b/src/Admin/MoocAdmin.php
@@ -3,18 +3,37 @@
 namespace AppBundle\Admin;
 
 use AppBundle\Entity\Mooc\Chapter;
+use AppBundle\Entity\Mooc\Mooc;
 use AppBundle\Form\PurifiedTextareaType;
+use AppBundle\Mooc\MoocManager;
 use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Route\RouteCollection;
 use Symfony\Bridge\Doctrine\Form\Type\EntityType;
+use Symfony\Component\Form\Extension\Core\Type\FileType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\Extension\Core\Type\TimeType;
 
 class MoocAdmin extends AbstractAdmin
 {
+    /**
+     * @var MoocManager
+     */
+    private $moocManager;
+
+    public function __construct(
+        string $code,
+        string $class,
+        string $baseControllerName,
+        MoocManager $moocManager
+    ) {
+        parent::__construct($code, $class, $baseControllerName);
+
+        $this->moocManager = $moocManager;
+    }
+
     protected function configureFormFields(FormMapper $formMapper)
     {
         $formMapper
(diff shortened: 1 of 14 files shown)


src/Entity/Mooc/Mooc.php
```diff
@@ -125,6 +121,19 @@ class Mooc
...
 
+    /**
+     * @var UploadedFile|null
+     *
+     * @Assert\Image(
+     *     mimeTypes={"image/jpeg", "image/png"},
+     *     maxSize="1M",
+     *     maxWidth="480",
+     *     maxHeight="360",
+     *     minWidth="211"
```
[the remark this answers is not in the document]
JeremieSamson: yep, confirmed by po

src/Admin/MoocAdmin.php
```diff
@@ -158,15 +152,23 @@ protected function configureListFields(ListMapper $listMapper): void
         ;
     }
 
-    protected function configureRoutes(RouteCollection $collection): void
+    private function addImageSrc(): ?string
+    {
+        return $this->getSubject() && $this->getSubject()->getImageName()
+            ? '<br/>><img src="'.$this->getSubject()->getAssetImagePath().'" />'
```
thomasbisignani: There is a useless `>` char : 
```php 
<br/>><img src
``` 
=>
```php
<br/><img src
```
What about using a method (`getImageHelpMessage()`) to directly return the image help message with the img tag if an image is uploaded  ?

src/Normalizer/MoocListNormalizer.php
```diff
@@ -0,0 +1,44 @@
...
+    private $router;
+
+    public function __construct(RouterInterface $router)
+    {
+        $this->router = $router;
+    }
+
+    public function normalize($object, $format = null, array $context = [])
+    {
+        return $this->normalizeMooc($object);
```
ottaviano: why do you add a new private method?

src/Normalizer/MoocListNormalizer.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function supportsNormalization($data, $format = null)
+    {
+        return self::FORMAT === $format && $data instanceof Mooc;
+    }
+
+    private function normalizeMooc(Mooc $mooc): array
+    {
+        return [
+            'chapter_count' => $mooc->countChapters(),
```
ottaviano: I think you don't need to manage other fields, but just `image`. You can decorate ItemNormalizer of ApiPlatform to do this.

templates/form_theme_admin.html.twig
```diff
@@ -0,0 +1,14 @@
+{% extends '@SonataAdmin/Form/form_admin_fields.html.twig' %}
+
+{% block app_admin_mooc_mooc_image_file_image_row -%}
+    {{ block('admin_image_with_url_row') }}
+{%- endblock app_admin_mooc_mooc_image_file_image_row %}
```
ottaviano: can you remove the block name here pls, as the second endblock L14 is without the name?

templates/form_theme_admin.html.twig
```diff
@@ -0,0 +1,14 @@
...
+
+{% block admin_image_with_url_row %}
+    {{ block('form_row') }}
+
+    {% set object = form.parent.vars.value %}
+    {% if object.imageName is defined and object.imageName is not null %}
```
ottaviano: `is not null` => `is not empty`

thomasbisignani:
:ok_hand:

PR #2752, opened 2018-10-09
Add args on phpunit in makefile

JeremieSamson:
In order to pass phpunit args like --filter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,10 @@ EXEC?=$(DOCKER_COMPOSE) exec app
 COMPOSER=$(EXEC) composer
 CONSOLE=bin/console
 PHPCSFIXER?=$(EXEC) php -d memory_limit=1024m vendor/bin/php-cs-fixer
+BEHAT=$(EXEC) vendor/bin/behat
 BEHAT_ARGS?=-vvv
+PHPUNIT=$(EXEC) vendor/bin/phpunit
+PHPUNIT_ARGS?=-v
 
 .DEFAULT_GOAL := help
 .PHONY: help start stop reset db db-diff db-diff-dump db-migrate db-rollback db-load watch clear clean test tu tf tj lint ls ly lt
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -120,10 +120,10 @@ test-behat:
 	$(EXEC) vendor/bin/behat $(BEHAT_ARGS)
 
 test-phpunit-functional:                                                                               ## Run phpunit fonctional tests
-	$(EXEC) vendor/bin/phpunit --group functional
+	$(EXEC) vendor/bin/phpunit $(PHPUNIT_ARGS) --group functional
 
 tu: vendor app/config/assets_version.yml                                                               ## Run the PHP unit tests
-	$(EXEC) vendor/bin/phpunit --exclude-group functional
+	$(EXEC) vendor/bin/phpunit $(PHPUNIT_ARGS) --exclude-group functional
```
tony-tran: Good idea. Maybe you can create a new step just to run phpunit with argument only and use this new step in `tu` ? WDYT ?

JeremieSamson:
@tony-tran makefile updated

Makefile
```diff
@@ -5,6 +5,8 @@ COMPOSER=$(EXEC) composer
 CONSOLE=bin/console
 PHPCSFIXER?=$(EXEC) php -d memory_limit=1024m vendor/bin/php-cs-fixer
 BEHAT_ARGS?=-vvv
+PHPUNIT=$(EXEC) vendor/bin/phpunit
```
thomasbisignani: Can you add a default args for the `PHPUNIT_ARGS` variable ?
example :  `PHPUNIT_ARGS?=-v`

Makefile
```diff
@@ -5,6 +5,8 @@ COMPOSER=$(EXEC) composer
...
+BEHAT=$(EXEC) vendor/bin/behat
```
thomasbisignani: Can you add a default args for the PHPUNIT_ARGS variable ?
example : `PHPUNIT_ARGS?=-v`

PR #2753, opened 2018-10-09
Feature/ch5523/referent filter

tony-tran:
Please merge this https://github.com/EnMarche/worker/pull/17 before

### DIFF
--- a/front/style/pages/_referent.scss
+++ b/front/style/pages/_referent.scss
@@ -37,6 +37,21 @@
                 margin: 0 5px;
             }
         }
+
+        &__search {
+            &__col {
+                flex: 1;
+                margin: 0 100px 0 0;
+
+                .l__row {
+                    margin: 0 0 20px 0;
+                }
+
+                &__area-code {
+                    margin: 0 0 10px 0 !important;
+                }
+            }
+        }
     }
 
     &__btn-create-event {
@@ -55,7 +70,6 @@
         }
 
         &__metadata {
-            margin-left: 10px;
             color: $gray--muted;
         }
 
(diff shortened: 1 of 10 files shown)


translations/messages.fr.yml
```diff
@@ -3,6 +3,8 @@
 #
 common.gender.man: Homme
 common.gender.woman: Femme
+common.gender.male: Homme
+common.gender.female: Femme
```
ottaviano: where are these keys used?

tests/Controller/EnMarche/ReferentControllerTest.php
```diff
@@ -18,6 +18,7 @@
 /**
  * @group functional
  * @group referent
+ * @group wip
```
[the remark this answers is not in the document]
tony-tran: Oh my bad... Thank you

translations/messages.fr.yml
tony-tran: I add this to translate the Genders class

tony-tran: Ok I see, there is `const CHOICES`, thank you

front/style/pages/_referent.scss
```diff
@@ -37,6 +37,21 @@
                 margin: 0 5px;
             }
         }
+
+        &__search {
+            &__col {
+                flex: 1;
+                margin: 0 100px 0px 0;
```
thomasbisignani: `margin: 0 100px 0 0;` 

or maybe just `margin-right: 100px` ? (if is not an override)

src/Entity/Projection/ReferentManagedUser.php
```diff
@@ -173,6 +173,11 @@ class ReferentManagedUser
      */
     private $createdAt;
 
+    /**
```
thomasbisignani: To be consistent with this class, you can add this PHPDoc : 
`* @var string|null`

PR #2757, opened 2018-10-11
ch5523 Fix listing of adherent

tony-tran:


templates/referent/users_list.html.twig
```diff
@@ -211,11 +211,11 @@
...
                     <th style="width: 100px;">Prénom</th>
                     <th style="width: 100px;">Nom</th>
-                    <th style="width: 100px;">Age</th>
-                    <th style="width: 100px;">Code postal</th>
+                    <th style="width: 100px;">Téléphone</th>
+                    <th style="width: 100px;">Ville</th>
                     <th style="width: 120px;">Code postal du comité géré</th>
-                    <th class="datagrid__table__col--left">Informations</th>
-                    <th style="width: 220px;" class="datagrid__table__col--left datagrid__table__col--hide-mobile">Comités</th>
+                    <th style="width: 100px;" class="datagrid__table__col--left">Informations</th>
```
JeremieSamson: I'm not a CSS guy but having width with px instead of % isn't very responsive right ?

tony-tran: I'm not too, but I think maybe the css parent manage the responsive ? In any case, I look in mobile version and it's okay.

PR #2758, opened 2018-10-11
Update Referent API endpoints

illusionOfParadise:


app/config/services/services.xml
```diff
@@ -351,5 +351,12 @@
 
         <!-- Storage -->
         <service id="AppBundle\Storage\ImageStorage" />
+
+        <!-- ParamConverter -->
+        <service id="AppBundle\Request\ParamConverter\DoctrineQueryStringConverter">
```
ottaviano: 1) I think `App\ParamConverter` as namespace will be enough 
2) Can you add Param in the class name `...StringConverter` => `...StringParamConverter`?

src/Controller/Api/AdherentsController.php
```diff
@@ -33,14 +36,15 @@ public function adherentsCountAction(AdherentRepository $adherentRepository): Re
     /**
      * @Route("/count-by-referent-area", name="app_adherents_count_for_referent_managed_area")
      * @Method("GET")
+     * @Entity("referent", expr="repository.findOneByEmail(referent)", converter="querystring")
```
ottaviano: I think we need to use `findReferent(referent)` instead of `findOneByEmail`

src/Request/ParamConverter/DoctrineQueryStringConverter.php
```diff
@@ -0,0 +1,308 @@
...
+use Symfony\Component\ExpressionLanguage\SyntaxError;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpFoundation\Request;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\ORM\NoResultException;
+
+/**
+ * Puts specific attribute parameter to query parameters and does the same job as DoctrineParamConverter.
+ */
+class DoctrineQueryStringConverter implements ParamConverterInterface
```
ottaviano: what do you think about this solution:
```php
class QueryStringParamConverter extends DoctrineParamConverter
{
    public function apply(Request $request, ParamConverter $configuration)
    {
        $name = $configuration->getName();

        if (!$request->query->has($name)) {
            throw new \InvalidArgumentException($name.' param is missing in the request');
        }

        $request->attributes->set($name, $request->query->get($name));

        return parent::apply($request, $configuration);
    }
}
```

with a service tag:
```
<tag name="request.param_converter" converter="querystring" priority="false" />
```

`priority=false` allow use this converter via `converter="querystring"` annotation and avoid adding is in the stack of all converters.

illusionOfParadise: Changed.

app/config/services/services.xml
illusionOfParadise: Done.

PR #2759, opened 2018-10-12
Update thumbnail template and add link on youtube id

JeremieSamson:


templates/admin/list/list_youtube_id.html.twig
```diff
@@ -0,0 +1,9 @@
+{% extends get_admin_template('base_list_field', admin.code) %}
+
+{% block field %}
+    <div style="text-align: center;">
```
thomasbisignani: This Sonata Admin includes Bootstrap 3, so you can apply this class here : 

```html
<div class="text-center">
```

JeremieSamson: Thanks CSS sensei

thomasbisignani: :man_cartwheeling:

PR #2763, opened 2018-10-12
Fix/ch5431/some fixes for cp

tony-tran:


front/style/pages/_citizen-projects.scss
```diff
@@ -304,6 +304,10 @@
             }
         }
     }
+
+    .committee-follow--anonymous__link {
```
JeremieSamson: What do you thin of putting it like this : 
```
.committee-follow {
  $--anonymous__link {
 }
}
```

tony-tran: Yes good idea but is it really necessary because there is just one css modifition for this class? Maybe to be consistent ?

src/Admin/CitizenProjectAdmin.php
```diff
@@ -361,6 +361,22 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                'show_filter' => true,
+                'field_type' => ChoiceType::class,
+                'field_options' => [
+                    'choices' => [
+                        'Projets simples' => CitizenProject::SIMPLE_TYPE,
+                        'Projets clés en main' => CitizenProject::TURKNEY_TYPE,
+                    ],
+                ],
+                'callback' => function (ProxyQuery $qb, string $alias, string $field, array $value) {
+                    $qb->andWhere(\sprintf('%s.turnkeyProject is %s', $alias, CitizenProject::TURKNEY_TYPE === $value['value'] ? 'not null' : 'null'));
```
thomasbisignani: Add a backslash on the `sprintf` function is not needed.

We use the `compiler_optimized` option in PHP CS Fixer. ( https://github.com/EnMarche/en-marche.fr/blob/master/.php_cs.dist#L23 )

You can see all functions where it is needed here : https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L350

front/style/pages/_citizen-projects.scss
JeremieSamson: Just to be consistent

templates/admin/citizen_project/type.html.twig
```diff
@@ -0,0 +1,5 @@
+{% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
+
+{% block field %}
+    {{ object.isFromTurnkeyProject ? 'Projet clé en main' : 'Projet simple' }}
```
ottaviano: why do you not use `object.typeOfProject()`?

src/Admin/CitizenProjectAdmin.php
```diff
@@ -361,6 +361,22 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
             ->add('featured', null, [
                 'label' => 'Coup de coeur',
             ])
+            ->add('turnkeyProject', CallbackFilter::class, [
+                'label' => 'Type de projet',
+                'show_filter' => true,
+                'field_type' => ChoiceType::class,
+                'field_options' => [
+                    'choices' => [
```
ottaviano: what about using your const here `'choices' => CitizenProject::TYPES,` ?

src/Entity/CitizenProject.php
```diff
@@ -51,6 +51,14 @@ class CitizenProject extends BaseGroup
         self::PRE_REFUSED,
     ];
 
+    public const SIMPLE_TYPE = 'simple';
+    public const TURNKEY_TYPE = 'turnkey';
+
+    public const TYPE = [
```
ottaviano: what about renaming the const on `TYPES` or `LABELS`?

src/Admin/CitizenProjectAdmin.php
```diff
@@ -372,6 +385,10 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
     protected function configureListFields(ListMapper $listMapper)
     {
         $listMapper
+            ->add('turnkeyProject', null, [
```
ottaviano: you can remove the template as it does nothing more than call one getter, you can do it with virtual field:
```
->add('projectType', null, [
  'virtual' => true,
  'label' => 'Type de projet',
])
```

src/Entity/CitizenProject.php
```diff
@@ -735,4 +743,13 @@ public function isNotFinalStatus(): bool
     {
         return \in_array($this->status, self::NOT_FINAL_STATUSES, true);
     }
+
+    public function getTypeOfProject(): string
```
ottaviano: can you rename on `getProjectType` pls?

src/Admin/CitizenProjectAdmin.php
```diff
@@ -361,6 +361,19 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
             ->add('featured', null, [
                 'label' => 'Coup de coeur',
             ])
+            ->add('turnkeyProject', CallbackFilter::class, [
+                'label' => 'Type de projet',
+                'show_filter' => true,
+                'field_type' => ChoiceType::class,
+                'field_options' => [
+                    'choices' => array_flip(CitizenProject::TYPE),
```
ottaviano: Good idea 👍

PR #2764, opened 2018-10-12
Fix HTML rendering

JeremieSamson:


web/apps/projets-citoyens/src/js/components/TurnkeyProjectDetail.js
```diff
@@ -46,11 +46,9 @@ const TurnkeyProjectDetail = props => {
...
             </div>
             {!swap && <Title title={title} subtitle={subtitle} />}
-            <p style={{ width: `${props.textSize}` }}>
-                {description}
-                <br/><br/>
-                {solution}
-            </p>
+
+            <p style={{ width: `${props.textSize}` }}>{description}</p>
+            <p className="solution" style={{ width: `${props.textSize}` }} dangerouslySetInnerHTML={{__html: solution}} />
```
ottaviano: What about use the div element and dont touch margin css rule of p element? 
And if the solution is a html code usine div element is more correct that p

PR #2770, opened 2018-10-17
Unregister on event

tony-tran:


front/services/api/ReqwestApiClient.js
```diff
@@ -158,6 +158,25 @@ export default class ReqwestApiClient {
         });
     }
 
+    unregisterEvent(eventSlug, token, callback) {
```
JeremieSamson: File is named "ReqwestApiClient", should be RequestApiClient no ?

src/Controller/EnMarche/EventController.php
```diff
@@ -173,4 +174,28 @@ public function invitationSentAction(Request $request, Event $event): Response
...
     }
+
+    /**
+     * @Route("/desinscription", name="app_event_unregistration", condition="request.isXmlHttpRequest()")
+     * @Method("GET|POST")
+     */
+    public function unregistrationAction(Request $request, Event $event): JsonResponse
+    {
+        if (!$this->isCsrfTokenValid('citizen_action.unregistration', $token = $request->request->get('token'))) {
+            throw $this->createAccessDeniedException('Invalid CSRF protection token to unregister from the citizen action.');
```
ottaviano: I think BadRequestHttpException is more consistent in this case

src/Controller/EnMarche/EventController.php
```diff
@@ -173,4 +174,28 @@ public function invitationSentAction(Request $request, Event $event): Response
...
+        }
+
+        $eventRegistrationManager = $this->get('app.event.registration_manager');
```
ottaviano: can you inject to the method this service?

src/Entity/BaseEvent.php
```diff
@@ -238,7 +238,9 @@ public function incrementParticipantsCount(int $increment = 1): void
 
     public function decrementParticipantsCount(int $increment = 1): void
     {
-        $this->participantsCount -= $increment;
+        if (0 < $this->participantsCount) {
```
ottaviano: good but not perfect, with `$this->participantsCount = 1` and `$increment = 2` it'll be negatif.

src/Twig/EventRuntime.php
```diff
@@ -0,0 +1,27 @@
...
+    private $eventRegistrationRepository;
+
+    public function __construct(EventRegistrationRepository $eventRegistrationRepository)
+    {
+        $this->eventRegistrationRepository = $eventRegistrationRepository;
+    }
+
+    public function isAlreadyParticipating(Event $event, ?UserInterface $user): bool
+    {
+        if (!$user instanceof Adherent || !$user) {
```
ottaviano: the check `!$user instanceof Adherent` is enough here, no?

src/Twig/EventExtension.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+class EventExtension extends AbstractExtension
+{
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('is_already_participating', [EventRuntime::class, 'isAlreadyParticipating']),
```
ottaviano: what about `is_event_already_participating`

templates/events/_header.html.twig
```diff
@@ -35,7 +35,11 @@
                 L'événement à été annulé
             </div>
         {% elseif not event.full%}
```
ottaviano: can you add one space here pls?

src/Controller/EnMarche/EventController.php
```diff
@@ -173,4 +174,28 @@ public function invitationSentAction(Request $request, Event $event): Response
...
         ]);
     }
+
+    /**
+     * @Route("/desinscription", name="app_event_unregistration", condition="request.isXmlHttpRequest()")
+     * @Method("GET|POST")
+     */
+    public function unregistrationAction(Request $request, Event $event): JsonResponse
+    {
+        if (!$this->isCsrfTokenValid('citizen_action.unregistration', $token = $request->request->get('token'))) {
```
ottaviano: what about renaming `citizen_action.unregistration` on `event.unregistration` for both CitizenAction and Event unsubscriptions? (as CitizenAction is an Event too)

templates/events/_header.html.twig
```diff
@@ -35,7 +35,11 @@
...
-            <a class="btn btn--no-border {{ isFull?' btn--full':''}} register-event" href="{{ path("app_#{event.type}_attend", {slug: event.slug}) }}">Je veux participer</a>
+            {% if is_already_participating(event, app.user) %}
+                <button class="btn btn--no-border {{ isFull?' btn--full':''}} unregister-event" data-csrf-token="{{ csrf_token('citizen_action.unregistration') }}">Je ne veux plus participer</button>
```
ottaviano: I think it's the wrong place for this button, because we are in `elseif not event.full`, when the event will be full I'll not see this button.

templates/search/_events.html.twig
```diff
@@ -58,7 +58,7 @@
                 {% elseif not event.full %}
                     <a href="{{ path("app_#{event.type}_show", { slug: event.slug }) }}"
                        class="btn btn--blue" target="_blank" rel="noopener noreferrer">
-                        Participer
+                        Voir plus
```
ottaviano: why this change?

front/services/api/ReqwestApiClient.js
tony-tran: Yes, I catch that, I propose to change this in another PR WDYT ?

src/Twig/EventRuntime.php
tony-tran: Yes because $user can be null (unauthenticated user)

templates/search/_events.html.twig
tony-tran: It's a P.O demand

src/Twig/EventRuntime.php
ottaviano: @tony-tran so if yes, and the first check is enough you will remove the second one?

src/Controller/EnMarche/EventController.php
```diff
@@ -173,4 +176,26 @@ public function invitationSentAction(Request $request, Event $event): Response
...
+     * @Method("GET|POST")
+     */
+    public function unregistrationAction(Request $request, Event $event, EventRegistrationManager $eventRegistrationManager): JsonResponse
+    {
+        if (!$this->isCsrfTokenValid('event.unregistration', $token = $request->request->get('token'))) {
+            throw new BadRequestHttpException('Invalid CSRF protection token to unregister from the citizen action.');
+        }
+
+        if (!($adherentEventRegistration = $eventRegistrationManager->searchRegistration($event, $this->getUser()->getEmailAddress(), null))) {
+            return new JsonResponse(
```
ottaviano: what about:
```php
throw $this->createNotFoundException(
    'Impossible d\'exécuter la désinscription de l\'évènement, votre inscription n\'est pas trouvée.'
)
```

src/Twig/EventRuntime.php
```diff
@@ -0,0 +1,27 @@
...
+        $this->eventRegistrationRepository = $eventRegistrationRepository;
+    }
+
+    public function isEventAlreadyParticipating(Event $event, ?UserInterface $user): bool
+    {
+        if (!$user instanceof Adherent) {
+            return false;
+        }
+
+        return $this->eventRegistrationRepository->findGuestRegistration($event->getUuid(), $user->getEmailAddress()) ? true : false;
```
ottaviano: as an alternative:
```
return (bool) $this->eventRegistrationRepository->findGuestRegistration(
    $event->getUuid(), 
    $user->getEmailAddress()
);
```

templates/events/_header.html.twig
```diff
@@ -34,12 +34,20 @@
             <div class="text--white text--center">
                 L'événement à été annulé
             </div>
-        {% elseif not event.full%}
-            <a class="btn btn--no-border {{ isFull?' btn--full':''}} register-event" href="{{ path("app_#{event.type}_attend", {slug: event.slug}) }}">Je veux participer</a>
+        {% elseif not event.full %}
```
ottaviano: can you refactor this part for writing once the button block, something like that: 
```
{% else %}
     {% if is_event_already_participating(event, app.user) %}
          <button .../>
     {% endif %}

     {% if not event.full %}
           // ...
     {% else %}
           // ...
     {% endif %}
{% endif %}
```

src/Twig/EventRuntime.php
tony-tran: I prefer don't cast the result to be sure to control the return.

ottaviano: it is the exactly what you are doing with 
```
return $object ? true : false;
=== 
return (bool) object;
```
to avaluate the result of `if ($object)` PHP cast the $object in boolean value.

PR #2777, opened 2018-10-18
[Jecoute] Expose choices id

thomasbisignani:


src/Entity/Jecoute/Choice.php
```diff
@@ -20,6 +20,8 @@ class Choice
      * @ORM\Column(type="integer")
      * @ORM\Id
      * @ORM\GeneratedValue(strategy="AUTO")
+     *
+     * @JMS\Groups({"survey_list"})*
```
ottaviano: can you remove the trailing `*` pls

PR #2778, opened 2018-10-18
Change a referent order in the export in the admin + add helper

illusionOfParadise:


app/config/services/admin.xml
```diff
@@ -268,6 +268,10 @@
             <argument>AppBundle\Entity\Referent</argument>
             <argument>AppBundle\Controller\Admin\AdminReferentController</argument>
             <argument type="service" id="app.form.data_transformer.referent_area_transformer" />
+            <call method="setTemplate">
+                <argument>list</argument>
```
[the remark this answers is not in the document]
illusionOfParadise: I need to override all block `pager_links` and not only `Pager/links.html.twig` template.

templates/admin/referent/admin_list.html.twig
```diff
@@ -0,0 +1,6 @@
+{% extends '@SonataAdmin/CRUD/base_list.html.twig' %}
+
+{% block pager_links %}
+    <div style="margin-top: 10px;">Faites dérouler ce dropdown pour exporter les équipes sélectionnées.</div>
```
JeremieSamson: Why do you add css in style instead of css files ?

illusionOfParadise: Because this template overrides a Sonata template that does not use the application css files. To use the existing css files we should add a link to css file, the inline css is lighter. We use this practice for Sonata templates in the project.

PR #2779, opened 2018-10-18
Admin for CommitteeMergeHistory

Remg:


src/Admin/CommitteeMergeHistoryAdmin.php
```diff
@@ -0,0 +1,98 @@
...
+                'show_filter' => true,
+                'field_type' => TextType::class,
+                'callback' => function (ProxyQuery $qb, string $alias, string $field, array $value) {
+                    if (!$value['value']) {
+                        return false;
+                    }
+
+                    $qb
+                        ->innerJoin("$alias.destinationCommittee", 'dc')
+                        ->andWhere('dc.name ILIKE :destinationName')
```
ottaviano: `LIKE`, it's already case_insensitive with the collation

src/Admin/CommitteeMergeHistoryAdmin.php
```diff
@@ -0,0 +1,98 @@
...
+                        ->setParameter('destinationName', '%'.$value['value'].'%')
+                    ;
+
+                    return true;
+                },
+            ])
+            ->add('createdAt', DateRangeFilter::class, [
+                'label' => 'Date de fusion',
+                'show_filter' => true,
+                'field_type' => 'sonata_type_date_range_picker',
```
ottaviano: `DateRangePickerType::class`

app/migrations/Version20181018125234.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20181018125234 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE committee_merge_histories ADD merged_by_id INT NOT NULL');
```
ottaviano: normally it'll crash if the table already has some rows because of `NOT NULL`

Remg: for some reason it's not crashing.
but i can't update a row that as such NULL value, indeed.

we plan to execute a SQL query after deploy so that existing lines has a merger by default

src/Entity/Reporting/CommitteeMergeHistory.php
```diff
@@ -47,6 +48,13 @@ class CommitteeMergeHistory
      */
     private $destinationCommittee;
 
+    /**
+     * @var string
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Administrator")
```
ottaviano: what about `@JoinColumn(nullable=false)`?

Remg: ?
``` ottaviano on 18 Oct Member

normally it'll crash if the table already has some rows because of NOT NULL
```

PR #2784, opened 2018-10-19
Add interets in event filter

tony-tran:


app/config/config.yml
```diff
@@ -85,6 +85,7 @@ twig:
         url_help_en_marche:     "https://aide.en-marche.fr"
         project_dir:            "%kernel.project_dir%"
         em_api_host:            "%env(EM_API_HOST)%"
+        adherent_interests:     "%adherent_interests%"
```
tony-tran: Do you prefer to inject to twig global or via the controller ?

src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -88,7 +100,9 @@ public function __construct(
         string $queryAreaCode,
         string $queryCity,
         string $queryId,
-        int $offset = 0
+        int $offset = 0,
+        array $interests = [],
+        string $gender
```
ottaviano: a mandatory argument must be before the optional.

src/Referent/ManagedUsersFilter.php
```diff
@@ -98,6 +102,7 @@ public function handleRequest(Request $request): self
         $this->queryFirstName = $query->get(self::PARAMETER_FIRST_NAME, '');
         $this->queryAgeMinimum = $query->getInt(self::PARAMETER_AGE_MIN);
         $this->queryAgeMaximum = $query->getInt(self::PARAMETER_AGE_MAX);
+        $this->queryInterests = $query->get(self::PARAMETER_INTEREST, []);
```
ottaviano: I think you should case the result in array

src/Repository/Projection/ReferentManagedUserRepository.php
```diff
@@ -159,6 +159,15 @@ private function createFilterQueryBuilder(Adherent $referent, ManagedUsersFilter
             $qb->andWhere($cityExpression);
         }
 
+        if (\count($interests = $filter->getQueryInterests()) > 0) {
```
ottaviano: you can avoid the if statement here and start with `foreach` directly

templates/referent/users_list.html.twig
```diff
@@ -143,8 +143,23 @@
...
-                            Exemple: "33450, 12, 8, 75002"
+                            Exemple: "59400, 18, CH"
                         </legend>
+
+                        <legend class="label__header label__header__interest">Centres d'intêret</legend>
+                        <div class="referent__filters__search__col__interest">
+                            <div class="form l__row l__row--wrap">
+                                {% for key, value in adherent_interests %}
+                                    <div class="form__checkbox form__checkbox--col">
+                                        <input type="checkbox" id="referent_pin_interests_{{ loop.index0 }}" name="{{ constant('PARAMETER_INTEREST', filter) }}[{{ key }}]" value="{{ key }}" {{ filter.queryInterests[key] is defined ? 'checked' }}>
```
ottaviano: I think you can use just `[]` for the name (without `{{ key }}`)
so you will have `i[]=int1&i[]=int10` and you can check `{{ key in filter.queryInterests ? 'checked' }}`
WDYT?

PR #2788, opened 2018-10-22
Change subscribe to an event button display and fix some tests

tony-tran:


templates/events/_header.html.twig
```diff
@@ -23,29 +23,31 @@
...
-                L'événement à été annulé
-            </div>
-        {% else %}
-            {% if is_event_already_participating(event, app.user) %}
-                <button class="btn btn--no-border {{ isFull?' btn--full':''}} unregister-event" data-csrf-token="{{ csrf_token('event.unregistration') }}">Je ne veux plus participer</button>
-            {% elseif not event.full %}
-                <a class="btn btn--no-border {{ isFull?' btn--full':''}} register-event" href="{{ path("app_#{event.type}_attend", {slug: event.slug}) }}">Je veux participer</a>
-            {% else %}
-                <button class="btn btn--no-border{{ isFull?' btn--full':''}} btn--text-blue btn--disabled register-event" disabled="disabled">
+        {% if not is_event_from_user_connected(event, app.user) %}
```
ottaviano: do we really need to add a twig function instead of using `event.organizer is same(app.user)`?

PR #2791, opened 2018-10-22
add interest in listing

tony-tran:


templates/referent/users_list.html.twig
```diff
@@ -259,10 +259,16 @@
...
+                                    Animateur local {{- is_host ? ' / ' -}}
+                                {%- elseif is_host -%}
+                                    Co-animateur local
+                                {%- else -%}
+                                    Adhérent
+                                {% endif %}
+                            </small>
+                            <br />
+                            <small class="text--italic">
+                                {{ user.interests ? "\n" ~ user.interests | join(', ') }}
```
ottaviano: what does it do `"\n" ~`?

templates/referent/users_list.html.twig
```diff
@@ -259,10 +259,16 @@
...
+                                    Animateur local {{- is_host ? ' / ' -}}
+                                {%- elseif is_host -%}
+                                    Co-animateur local
+                                {%- else -%}
+                                    Adhérent
+                                {% endif %}
+                            </small>
+                            <br />
+                            <small class="text--italic">
+                                {{ user.interests | join(', ') }}
```
thomasbisignani: Not a big deal but can you just write `|join` instead of `| join` ? we don't add a whitespace for filters

PR #2793, opened 2018-10-22
Update Referent API endpoint for autocomplete: get all results if an empty string in 'value' param

illusionOfParadise:


src/Entity/ReferentManagedArea.php
```diff
@@ -83,7 +83,8 @@ public function removeTag(ReferentTag $tag): void
     public function getOnlyManagedCountryCodes($value): array
     {
         return array_values(array_filter(array_map(function (ReferentTag $tag) use ($value) {
-            if (ctype_alpha($tag->getCode()) && 0 === stripos($tag->getName(), $value)) {
+            if (ctype_alpha($tag->getCode())
+                && (!$value || ($value && 0 === stripos($tag->getName(), $value)))) {
```
ottaviano: `!$value || ($value && A ...` === `!$value || (A ...`

PR #2798, opened 2018-10-24
change supervisor filter

tony-tran:


tests/Controller/EnMarche/ReferentControllerTest.php
```diff
@@ -18,6 +18,7 @@
 /**
  * @group functional
  * @group referent
+ * @group wip
```
ottaviano: :D 
try phpstorm to run one isolated test method :)

tony-tran: Rohhh my bad, sorry.

thomasbisignani: Another way, you can run it with :  

`docker-compose run --rm app vendor/bin/phpunit --filter ReferentControllerTest`

with the class / method name for the filter

templates/referent/users_list.html.twig
```diff
@@ -150,7 +146,7 @@
                             Exemple: "59400, 18, CH"
                         </legend>
 
-                        <legend class="label__header label__header__interest">Centres d'intêret</legend>
+                        <legend class="label__header label__header__interest">Centres d'intêrêt</legend>
```
ottaviano: `d'intérêt` ?

PR #2804, opened 2018-10-29
Referent Area - CP Filter

tony-tran:


src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -252,4 +260,9 @@ public function setAgeMaximum(?int $ageMaximum): void
     {
         $this->ageMaximum = $ageMaximum;
     }
+
+    public function includeCP(): bool
```
ottaviano: what about `getIncludeCitizenProject`? `includeCP` it is not clear if it's one getter or setter..

src/Referent/ManagedUsersFilter.php
```diff
@@ -263,4 +268,9 @@ public function setQueryInterests(array $queryInterests): void
     {
         $this->queryInterests = $queryInterests;
     }
+
+    public function includeCP(): bool
```
ottaviano: the same remark

src/Repository/Projection/ReferentManagedUserRepository.php
```diff
@@ -199,6 +199,10 @@ private function createFilterQueryBuilder(Adherent $referent, ManagedUsersFilter
             $typeExpression->add($and);
         }
 
+        if ($filter->includeCP()) {
+            $typeExpression->add('u.citizenProjects IS NOT NULL or u.citizenProjectsOrganizer IS NOT NULL');
```
ottaviano: what happens if the `citizenProjects` column has `[]` as value?

templates/referent/users_list.html.twig
```diff
@@ -240,6 +250,19 @@
                         <td>{{ user.lastName }}</td>
                         <td>{{ user.phone ? user.phone|phone_number_format : '~' }}</td>
                         <td>{{ user.postalCode }} <br/> {{ user.city }}, {{ user.country }}</td>
+                        <td class="datagrid__table__col--left">
+                            {% for slug, name in user.citizenProjectsOrganizer %}
+                                <ul>
```
ottaviano: Do you create one `<ul>` by link?

templates/referent/users_list.html.twig
```diff
@@ -240,6 +250,19 @@
...
+                                    <li><a href="{{ path('app_citizen_project_show', {'slug': slug}) }}"><strong>{{ name }}</strong></a></li>
+                                </ul>
+                            {% endfor %}
+
+                            {% for slug, name in user.citizenProjects %}
+                                <ul>
```
ottaviano: the same question

src/Repository/Projection/ReferentManagedUserRepository.php
ottaviano: 1) 
```
$filter->getIncludeCitizenProject() && \count($filter->getIncludeCitizenProject()) > 0
```
the second condition is useless here, the first one checks that the array is not empty.


2) By my initial comment, I mean that the column `u.citizenProjects` can be `null` and can have `[]` as a valid JSON value and the rows with `[]` as a value will be selected with your request.

PR #2806, opened 2018-10-29
Add an async sending of deputy emails

illusionOfParadise:


src/Entity/DeputyManagedUsersMessage.php
```diff
@@ -0,0 +1,62 @@
...
+        return new self(
+            $message->getUuid(),
+            $message->getFrom(),
+            $message->getSubject(),
+            $message->getContent(),
+            $message->getDistrict()
+        );
+    }
+
+    public function getDistrict(): ?District
```
ottaviano: it is never `null`?

app/config/config_test.yml
```diff
@@ -24,6 +24,10 @@ services:
     old_sound_rabbit_mq.referent_message_dispatcher_producer:
         class: 'Tests\AppBundle\Test\Producer\ReferentMessageNullProducer'
 
+
```
ottaviano: too much empty lines

PR #2808, opened 2018-10-29
Popup to let deputy and referent be informed of the new chart

JeremieSamson:
no tests to check if user chartAccepted has been changed, is there any doctrine context ?
!important in scss is because <object> add css on element

### DIFF
--- a/features/popup_chart.feature
+++ b/features/popup_chart.feature
@@ -1,0 +1,51 @@
+@javascript
+Feature:
+  As deputy or a referent
+  I must have a popup when I am on my space only if I didn't accept it before
+
+  Background:
+    Given the following fixtures are loaded:
+      | LoadAdherentData      |
+      | LoadDistrictData      |
+
+  Scenario: As deputy, I should see a popup when I'm in the deputy space
+    Given I am logged as "deputy@en-marche-dev.fr"
+    And I am on "parametres/mon-compte"
+    When I follow "Espace député"
+    Then I should be on "espace-depute/utilisateurs/message"
+    And I should see "Charte de bonne utilisation des outils numériques - à l’usage des députés"
+
+  Scenario: As deputy, I should be redirected if I close the popup when I'm in the deputy space
+    Given I am logged as "deputy@en-marche-dev.fr"
+    And I am on "espace-depute/utilisateurs/message"
+    And I should see "Charte de bonne utilisation des outils numériques - à l’usage des députés"
+    When I press "close"
+    Then I should be on "/"
+
+  Scenario: As deputy, I should stay on the same page if I accept the condition
+    Given I am logged as "deputy@en-marche-dev.fr"
+    And I am on "espace-depute/utilisateurs/message"
+    And I should see "Charte de bonne utilisation des outils numériques - à l’usage des députés"
+    When I press "J'accepte"
+    Then I should be on "espace-depute/utilisateurs/message"
+
+  Scenario: As referent, I should see a popup when I'm in the referent space
+    Given I am logged as "referent-75-77@en-marche-dev.fr"
+    And I am on "parametres/mon-compte"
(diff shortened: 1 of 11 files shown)


src/Controller/EnMarche/UserController.php
```diff
@@ -196,4 +197,23 @@ public function terminateMembershipAction(Request $request): Response
             'form' => $form->createView(),
         ]);
     }
+
+    /**
+     * @Route("/chart",
+     *     name="app_user_set_accept_chart",
+     *     condition="request.isXmlHttpRequest()"
+     * )
+     * @Method("GET")
```
ottaviano: should use POST or PUT method here

src/Controller/EnMarche/UserController.php
```diff
@@ -196,4 +197,23 @@ public function terminateMembershipAction(Request $request): Response
...
+     */
+    public function chartAcceptationAction(): JsonResponse
+    {
+        $this->getUser()->setChartAccepted(true);
+
+        $this->getDoctrine()->getManager()->flush();
```
ottaviano: can you inject `ObjectManager` as an argument of the `chartAcceptationAction` method?

templates/referent/_layout.html.twig
```diff
@@ -68,5 +74,29 @@
                 {% block referent_content %}{% endblock %}
             </div>
         </section>
+
+        {% if not app.user.isChartAccepted %}
+            {% include 'chart_dialog.html.twig' with {
+                'title': 'referent.popup.title'|trans,
+                'subtitle': 'referent.popup.subtitle'|trans,
+                'fileUrl': 'referent.popup.file_url'|trans
```
ottaviano: I think it will be better to stock these URLs in the config file, `twig->globals` for example. WDYT?

templates/javascript.js.twig
```diff
@@ -356,3 +356,30 @@ $(document).ready(function() {
...
+            "{{ 'referent.button.accept'|trans }}": function () {
+                $.ajax({
+                    url: "{{ path('app_user_set_accept_chart')|e('js') }}"
+                });
+
+                $(this).dialog("destroy");
+            }
+        },
+        beforeClose: function (event, ui) {
+            window.location.replace("{{ path('homepage') }}");
```
ottaviano: why this redirection?

JeremieSamson: Asked by PO, user has to accept or he is redirected to the homepage

features/popup_chart.feature
```diff
@@ -0,0 +1,56 @@
+Feature:
```
thomasbisignani: You can add the `@javascript` tag here instead of each scenario

templates/referent/_layout.html.twig
JeremieSamson: In a logic of multi-language website and because the file is wrote in French, I thought it was better to add it there. 
But it could remain in this file and the key can be added in the twig globals @ottaviano ?

templates/javascript.js.twig
thomasbisignani: can you make this macro more generic to reuse it easly ? or rename it ?

src/Controller/EnMarche/UserController.php
```diff
@@ -196,4 +198,23 @@ public function terminateMembershipAction(Request $request): Response
...
     }
+
+    /**
+     * @Route("/chart",
+     *     name="app_user_set_accept_chart",
+     *     condition="request.isXmlHttpRequest()"
+     * )
+     * @Method("PUT")
+     */
+    public function chartAcceptationAction(EntityManager $em): JsonResponse
```
thomasbisignani: You can inject directly the `ObjectManager` here

```php
public function chartAcceptationAction(ObjectManager $manager): JsonResponse
```

PR #2809, opened 2018-10-29
Remove maps from event pages

marineb:


templates/events/show.html.twig
```diff
@@ -63,49 +63,6 @@
...
-                map: map,
-                title: '{{ event.name|e('js') }}'
-            });
-        }
-        initMap();
-
-        Kernel.onLoad(function(di) {
-            var unregistrationButton = dom('.unregister-event');
-
-            unregistrationButton && on(unregistrationButton, 'click', function () {
```
ottaviano: do we really want to remove the event unregistration process?

marineb: oups. nope

PR #2815, opened 2018-10-30
Update referent contrat popup

marineb:


front/style/layout/_layout.scss
```diff
@@ -467,6 +467,21 @@
...
+        width: 97% !important;
+        left: 0 !important;
+    }
+
+    .ui-widget-content {
+        border: 0;
+    }
+
+    .ui-dialog-titlebar {
+        background: white;
```
thomasbisignani: You can use the `$white` variable here no ?

PR #2821, opened 2018-10-31
[Jecoute] Fixed not nullable for survey field

thomasbisignani:


src/Entity/Jecoute/DataSurvey.php
```diff
@@ -109,6 +109,8 @@ class DataSurvey
     /**
      * @ORM\ManyToOne(targetEntity="Survey")
      * @ORM\JoinColumn(onDelete="CASCADE", nullable=false)
+     *
+     * @Assert\NotNull
```
ottaviano: `NotBlank` will be better, no?

PR #2829, opened 2018-11-07
[API Sync] Sync lat,long, and district CP props

ottaviano:


src/Entity/CitizenProject.php
```diff
@@ -260,21 +262,6 @@ public function getPostAddress(): NullablePostAddress
         return $this->postAddress;
     }
 
-    public function getLatitude()
```
ottaviano: these getters already exist in https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/EntityNullablePostAddressTrait.php#L94

PR #2830, opened 2018-11-08
Feature 5451 - Ban adherent feature

tony-tran:


thomasbisignani:
You should add also the new role `ROLE_ADMIN_BAN` in the AdministratorAdmin et the Admin fixtures : 

https://github.com/EnMarche/en-marche.fr/blob/master/src/Admin/AdministratorAdmin.php#L106
https://github.com/EnMarche/en-marche.fr/blob/master/src/DataFixtures/ORM/LoadAdminData.php#L58

src/Validator/AdherentBannedValidator.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    private $objectManager;
+
+    public function __construct(ObjectManager $objectManager)
+    {
+        $this->objectManager = $objectManager;
+    }
+
+    public function validate($value, Constraint $constraint)
+    {
```
thomasbisignani: You should also check the constraint `AdherentBanned` instance and throw a `UnexpectedTypeException`
Example :
```php
        if (!$constraint instanceof AdherentBanned) {
            throw new UnexpectedTypeException($constraint, AdherentBanned::class);
        }
```

src/Controller/Admin/AdminAdherentController.php
```diff
@@ -28,4 +33,30 @@ public function exitImpersonationAction(): Response
 
         return $this->redirectToRoute('admin_app_adherent_list');
     }
+
+    /**
+     * @Route("adherent/{id}/ban", name="app_admin_ban_adherent")
```
ottaviano: can prefix by `/` the URI?

templates/admin/adherent/edit.html.twig
```diff
@@ -1,6 +1,13 @@
 {% extends 'SonataAdminBundle:CRUD:base_edit.html.twig' %}
 
 {% block actions %}
+    <script>
+        $(function(){
+            $('.ban-link').click(function(){
+                $('#ban').submit();
```
ottaviano: why do you not use `<input type=submit>` button?

tony-tran: I prefer to keep the `<a href>` to have the same design of links.
If I add the `input` I need to add the click event on this input and IMHO it's no necessary. I prefer to submit directly the form. WDYT ?

PR #2838, opened 2018-11-09
Fix wrong logic in edition and update css

JeremieSamson:


templates/adherent/profile.html.twig
```diff
@@ -104,8 +104,12 @@
                      <span class="b__nudge--bottom-tiny text--gray">{{ 'adherent.form.elected'|trans }}</span>
                      <span class="text--gray text--italic">({{ 'common.optional'|trans }})</span>
                  </div>
-                {{ form_widget(form.elected) }}
-                {{ form_errors(form.elected) }}
+
+                 <div class="form__checkbox">
+                     {{ form_widget(form.elected) }}
+                     {{ form_errors(form.elected) }}
+                     <label class="form form__label" for="adherent_elected"></label>
```
thomasbisignani: Why set the label here in the view and set `'label' => '',`  ?

JeremieSamson: I wanted to change the checkbox style.
Tried form_raw but the label was rendered as field name "Elected" next to the checkbox
Tried form_label but same behavior, no possibility to remove label 
So I added it manually with an empty content

PR #2846, opened 2018-11-13
Fix Algolia search on the CP page

illusionOfParadise:


templates/react.html.twig
```diff
@@ -34,13 +34,15 @@
...
 {% block content %}
-    <noscript>Vous devez activer Javascript pour voir cette page.</noscript>
-
-    <div id="root">
-        <div class="react-app-loader">
-            <img src="{{ asset('images/loader-sm.gif') }}" />
-            <br />
-            Chargement ...
+    <main>
+        <noscript>Vous devez activer Javascript pour voir cette page.</noscript>
```
thomasbisignani: Not a big deal but can you write `JavaScript` with the `S` ?

PR #2849, opened 2018-11-14
Change way to count character in CKEditor

tony-tran:


templates/javascript.js.twig
```diff
@@ -57,14 +57,20 @@ CKEDITOR.on('dialogDefinition', function(ev) {
     }
 });
 
+var decodeHTML = function (html) {
```
thomasbisignani: IMO we can use `const` and `let` here :

```
const decodeHTML = function (html) {
     let txt = document.createElement('textarea');
     txt.innerHTML = html;

     return txt.value;
 };
```

tony-tran: I would like to be consistent with the rest of the file. WDYT ?

thomasbisignani: Ok, we can modify that in another PR :+1:

PR #2864, opened 2018-11-19
Fix ckeditor counter and change CP sub menu

ottaviano:


templates/adherent/list_my_administrated_citizen_projects.html.twig
```diff
@@ -1,20 +1,13 @@
...
         <ul class="list__links">
-            {% for citizen_project in citizen_projects|slice(0, 1) if app.user.isAdministratorOf(citizen_project) %}
-                <li>
-                    <a href="{{ path('app_citizen_project_show', { slug: citizen_project.slug }) }}"
-                       title="{{ citizen_project.nameWithDistrict }}">
-                        {{- citizen_project.nameWithDistrict -}}
-                    </a>
-                </li>
-            {% endfor %}
+            <li><a href="#">{{ 'my_citizen_projects'|trans }}</a></li>
```
thomasbisignani: Why the `a` tag is used with a `href="#"` here ?

ottaviano: It's to have a link style and the `:hover` behavior.

PR #2869, opened 2018-11-21
Update mandate to array

JeremieSamson:
Update mandate choice in join and profil edition to mulitple choices

capture du 2018-11-21 16-34-36

### DIFF
--- a/front/style/pages/_register.scss
+++ b/front/style/pages/_register.scss
@@ -89,6 +89,22 @@
         &--flex {
             display: flex;
             align-items: center;
+
+            #form__mandates {
+                .select2-container {
+                    width: 100%;
+
+                    .select2-search-choice-close {
+                        border-bottom: none;
+                    }
+                }
+
+                .form__field {
+                    padding-left: 0;
+                    padding-right: 0;
+                    height: auto;
+                }
+            }
         }
 
         &--line {
--- a/src/Entity/Adherent.php
+++ b/src/Entity/Adherent.php
@@ -255,9 +255,9 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
     private $chartAccepted = false;
 
     /**
-     * @ORM\Column(type="string", length=22, nullable=true)
+     * @ORM\Column(type="simple_array", nullable=true)
      */
-    private $mandate;
+    private $mandates;
 
     public function __construct()
     {
@@ -281,7 +281,7 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
         string $registeredAt = 'now',
         ?array $tags = [],
         ?array $referentTags = [],
-        ?string $mandate = null
+        ?array $mandates = null
     ) {
         $adherent = new self();
 
@@ -300,7 +300,7 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
         $adherent->registeredAt = new \DateTime($registeredAt);
         $adherent->tags = new ArrayCollection($tags);
         $adherent->referentTags = new ArrayCollection($referentTags);
-        $adherent->mandate = $mandate;
+        $adherent->mandates = $mandates;
 
         return $adherent;
     }
@@ -642,7 +642,7 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
         $this->position = $membership->position;
         $this->phone = $membership->getPhone();
         $this->emailAddress = $membership->getEmailAddress();
-        $this->mandate = $membership->getMandate();
+        $this->mandates = $membership->getMandates();
 
         if (!$this->postAddress->equals($postAddress)) {
             $this->postAddress = $postAddress;
@@ -1198,8 +1198,18 @@ class Adherent implements UserInterface, GeoPointInterface, EncoderAwareInterfac
         $this->chartAccepted = $chartAccepted;
     }
 
-    public function getMandate(): ?string
+    public function getMandates(): ?array
     {
-        return $this->mandate;
+        return $this->mandates;
+    }
+
+    public function setMandates(?array $mandates): void
+    {
+        $this->mandates = $mandates;
+    }
+
+    public function hasMandate(): bool
+    {
+        return !empty($this->mandates);
     }
 }
--- a/src/Membership/MembershipRequest.php
+++ b/src/Membership/MembershipRequest.php
@@ -132,14 +132,13 @@ class MembershipRequest implements MembershipInterface
     private $elected = false;
 
     /**
-     * @var string|null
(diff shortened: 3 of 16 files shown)


front/style/pages/_register.scss
```diff
@@ -89,6 +89,22 @@
         &--flex {
             display: flex;
             align-items: center;
+
+            #form__mandate {
+                .select2-container {
+                    width: 100%;
+
+                    .select2-search-choice-close {
+                        border-bottom: unset;
```
thomasbisignani: `unset` is not compatible with some browsers. (IE / IE Mobile & co.) 
See : https://caniuse.com/#feat=css-unset-value

You should use the default value, example for this property : 
`border-bottom: none;`

front/style/pages/_register.scss
```diff
@@ -89,6 +89,22 @@
...
+                    }
+                }
+
+                .form__field {
+                    padding-left: unset;
```
thomasbisignani: Same here, and also for the `padding-right` and `height`

src/Membership/MembershipRequest.php
```diff
@@ -170,7 +169,7 @@ public static function createFromAdherent(Adherent $adherent): self
         $dto->phone = $adherent->getPhone();
         $dto->emailAddress = $adherent->getEmailAddress();
         $dto->mandate = $adherent->getMandate();
-        $dto->elected = !\is_null($adherent->getMandate());
+        $dto->elected = \count($adherent->getMandate());
```
thomasbisignani: What about adding `hasMandate()` method to return a boolean and use it here ?

templates/adherent/profile.html.twig
```diff
@@ -10,6 +16,10 @@
         Kernel.onLoad(function() {
             {{ js.address(form.address, { cityAutocomplete: true }) }}
         });
+
+        $(document).ready(function() {
+            $('.select2').select2();
```
thomasbisignani: Can you rename this class to have more specific name please ?

templates/membership/join.html.twig
```diff
@@ -21,13 +22,18 @@
...
     {% import 'javascript.js.twig' as js %}
     {% include 'membership/_mandate_script.html.twig' %}
 
     <script type="text/javascript">
         Kernel.onLoad(function() {
             App.runJoin('{{ form.vars.form_type_class|e('js') }}');
             {{ js.address(form.address) }}
+
+            $(document).ready(function() {
+                $('.select2').select2();
```
thomasbisignani: Same here for the class name

templates/adherent/profile.html.twig
```diff
@@ -10,6 +16,10 @@
         Kernel.onLoad(function() {
             {{ js.address(form.address, { cityAutocomplete: true }) }}
         });
+
+        $(document).ready(function() {
```
thomasbisignani: IMO we can call the ` $('.select2').select2();` inside the `Kernel.onLoad` no ?

src/Entity/Adherent.php
```diff
@@ -281,7 +281,7 @@ public static function create(
         string $registeredAt = 'now',
         ?array $tags = [],
         ?array $referentTags = [],
-        ?string $mandate = null
+        ?array $mandate = null
```
thomasbisignani: IMO we can now rename `$mandate` to `$mandates` , when we have an `simple_array`, we can pluralize the property.

src/Entity/Adherent.php
```diff
@@ -1198,7 +1198,7 @@ public function setChartAccepted(bool $chartAccepted): void
         $this->chartAccepted = $chartAccepted;
     }
 
-    public function getMandate(): ?string
+    public function getMandate(): ?array
```
thomasbisignani: IMO we can now rename `getMandate()` to `getMandates()`, WDYT ?

templates/adherent/profile.html.twig
JeremieSamson: :heavy_check_mark:

templates/membership/join.html.twig
JeremieSamson: :heavy_check_mark:

templates/adherent/profile.html.twig
JeremieSamson: :heavy_check_mark:

src/Membership/MembershipRequest.php
JeremieSamson: :heavy_check_mark:

front/style/pages/_register.scss
JeremieSamson: :heavy_check_mark:

front/style/pages/_register.scss
JeremieSamson: :heavy_check_mark:

src/Entity/Adherent.php
JeremieSamson: :heavy_check_mark:

src/Entity/Adherent.php
JeremieSamson: :heavy_check_mark:

src/Entity/Adherent.php
```diff
@@ -1198,8 +1198,18 @@ public function setChartAccepted(bool $chartAccepted): void
...
+
+    public function setMandates(array $mandates): void
+    {
+        $this->mandates = $mandates;
+    }
+
+    public function hasMandate(): bool
     {
-        return $this->mandate;
+        return \count($this->mandates) > 0;
```
ottaviano: it'll trigger an error it `$this->mandates` is null

ottaviano: and what about `isElectedPerson` as the name of method?

thomasbisignani: You can just use:
```php
return !empty($this->mandates)
```

src/Membership/MembershipRequest.php
JeremieSamson: IMO it's better to have `hasMandate` because it's closer than `isElectedperson` of the attribute.

src/Entity/Adherent.php
```diff
@@ -1198,8 +1198,18 @@ public function setChartAccepted(bool $chartAccepted): void
         $this->chartAccepted = $chartAccepted;
     }
 
-    public function getMandate(): ?string
+    public function getMandates(): ?array
+    {
+        return $this->mandates;
+    }
+
+    public function setMandates(array $mandates): void
```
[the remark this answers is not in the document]
JeremieSamson: :heavy_check_mark:

src/Entity/Adherent.php
JeremieSamson: :heavy_check_mark:

PR #2878, opened 2018-11-23
[Jecoute] Admin : Updated survey form

thomasbisignani:


src/Admin/JecouteSurveyAdmin.php
```diff
@@ -66,6 +66,15 @@ protected function configureListFields(ListMapper $listMapper)
         ;
     }
 
+    public function getTemplate($name)
```
ottaviano: can you use the service config instead pls?

```
<call method="setTemplate">
                <argument>edit</argument>
                <argument>admin/jecoute/survey_edit.html.twig</argument>
            </call>
```

templates/admin/jecoute/survey_edit.html.twig
```diff
@@ -0,0 +1,36 @@
+{% extends '@SonataAdmin/CRUD/base_edit.html.twig' %}
+
+{% block javascripts %}
+    {{ parent() }}
+
+    <script type="text/javascript">
+       $(document).ready(function() {
+           const displayChoices = function(questionType) {
+               const questionBloc = $(questionType).parents(".sonata-ba-field div[id$='_question']").first();
```
ottaviano: bloc => block ?

templates/admin/jecoute/survey_edit.html.twig
```diff
@@ -0,0 +1,36 @@
...
+           });
+
+           $('.question-disabled').each(function() {
+               $(this).find($("ul[id$='_question_type'] .radio label")).unbind('click');
+           });
+       });
+    </script>
+{% endblock %}
+
+{% block form %}
```
ottaviano: you don't need to override this block?

PR #2907, opened 2018-12-04
[ADI][ADMIN] Add theme admin

JeremieSamson:
Theme admin with image upload

theme_admin


src/Admin/IdeasWorkshopThemeAdmin.php
```diff
@@ -0,0 +1,109 @@
...
+    protected function configureListFields(ListMapper $listMapper)
+    {
+        $listMapper
+            ->add('name', null, [
+                'label' => 'Nom',
+            ])
+            ->add('enabled', null, [
+                'label' => 'Visibilité',
+            ])
+            ->add('image', null, [
```
thomasbisignani: You can use our custom `thumbnail` list type here, example :
```
            ->add('_image', 'thumbnail', [
                'virtual_field' => true,
                'label' => 'Image',
            ])```

File: src/Admin/IdeasWorkshopThemeAdmin.php [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+
+    public function __construct(
+        string $code,
+        string $class,
+        string $baseControllerName,
+        ThemeManager $themeManager
+    ) {
+        parent::__construct($code, $class, $baseControllerName);
+
+        $this->themeManager = $themeManager;
```
thomasbisignani: Instead of override the construct to inject the manager, you can inject it by setter injection.
You can create a `setThemeManager` method and configure the call method :  

```
<service id="app.admin.ideas_workshop_theme" class="AppBundle\Admin\IdeasWorkshopThemeAdmin">
     <tag name="sonata.admin" manager-type="orm" label="Thèmes des notes" group="Atelier des idées" />
     <argument />
     <argument>AppBundle\Entity\IdeasWorkshop\Theme</argument>
     <argument />

    <call method="setThemeManager">
        <argument type="service" id="AppBundle\IdeasWorkshop\ThemeManager"/>
    </call>
</service>
```

src/Entity/IdeasWorkshop/Theme.php
```diff
@@ -62,4 +77,14 @@ public function setEnabled(bool $enabled): void
     {
         $this->enabled = $enabled;
     }
+
+    public function getImagePath(): string
+    {
+        return sprintf('images/themes/%s', $this->getImageName());
```
illusionOfParadise: Add `ideas_workshop` to the images path: `images/ideas_workshop/themes/`.

src/IdeasWorkshop/ThemeManager.php
```diff
@@ -0,0 +1,39 @@
...
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+class ThemeManager
+{
+    public function __construct(ImageStorage $imageStorage)
+    {
+        $this->imageStorage = $imageStorage;
+    }
+
+    public function saveImage(Theme $theme): void
```
illusionOfParadise: This image management is repeated in multiple classes like
https://github.com/EnMarche/en-marche.fr/blob/master/src/Biography/BiographyManager.php
https://github.com/EnMarche/en-marche.fr/blob/master/src/Mooc/MoocManager.php
It would be nice to create a dedicated class.

JeremieSamson: IMO it should be in another PR, isolated from admin changes, but yes definitely.

src/Entity/IdeasWorkshop/Theme.php
JeremieSamson: done

PR #2919, opened 2018-12-12
[ADI][RFR] Add /ideas endpoint

JeremieSamson:


src/Serializer/IdeasWorkshopSerializerNormalizer.php
```diff
@@ -0,0 +1,39 @@
...
+use AppBundle\Repository\IdeaRepository;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+class IdeasWorkshopSerializerNormalizer implements NormalizerInterface
+{
+    private $normalizer;
+    private $ideaRepository;
+
+    public function __construct(
+        NormalizerInterface $normalizer,
```
thomasbisignani: multiple line useless here

src/Serializer/IdeasWorkshopSerializerNormalizer.php
```diff
@@ -0,0 +1,39 @@
...
+        IdeaRepository $ideaRepository
+    ) {
+        $this->normalizer = $normalizer;
+
+        $this->ideaRepository = $ideaRepository;
+    }
+
+    public function normalize($idea, $format = null, array $context = [])
```
thomasbisignani: We can keep the signature with the same variables names here : 
```php
public function normalize($object, $format = null, array $context = [])
```

PR #2921, opened 2018-12-12
Improve desintox page and add emoji support

marineb:


templates/desintox/list.html.twig
```diff
@@ -19,13 +19,17 @@
     </style>
 {% endblock %}
 
+{% block base_stylesheets %}
+    <link href="{{ asset('css/emoji.min.css') }}" rel="stylesheet">
```
Remg: you should add `{{ parent() }}` here

PR #2937, opened 2018-12-17
[ADI] Add themes, needs and idea categories list endpoints

illusionOfParadise:


app/config/config.yml
```diff
@@ -376,6 +376,21 @@ nelmio_cors:
...
 
 api_platform:
-    mapping:
-        paths: ['%kernel.project_dir%/src/Entity']
-    enable_docs: '%kernel.debug%'
+    name_converter: serializer.name_converter.camel_case_to_snake_case
+    resource_class_directories:
+        - '%kernel.project_dir%/src/Entity'
+    enable_docs: false
+    enable_swagger_ui: false
```
ottaviano: I think it is not very useful as you already disable the doc with `enable_docs: false`, WDYT?

app/config/config_test.yml
```diff
@@ -125,3 +125,11 @@ doctrine:
             default:
                 logging: false
                 profiling: false
+
+api_platform:
+    enable_docs: false
```
ottaviano: not needed as you do it in the global config file

src/Entity/IdeasWorkshop/EnabledInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace AppBundle\Entity\IdeasWorkshop;
```
ottaviano: what about moving it to `AppBundle\Entity`? This interface can be used by other entities

src/Entity/IdeasWorkshop/EnabledInterface.php
```diff
@@ -0,0 +1,7 @@
...
+
+interface EnabledInterface
+{
```
ottaviano: what about adding `public function isEnabled(): bool` method?

app/config/config_test.yml
illusionOfParadise: `config_test.yml` extends `config_dev.yml` that changes these values.

src/Entity/IdeasWorkshop/EnabledInterface.php
illusionOfParadise: Done.

src/Entity/IdeasWorkshop/EnabledInterface.php
illusionOfParadise: Done.

app/config/config.yml
illusionOfParadise: Done.

PR #2939, opened 2018-12-18
Add a dropdown to adherent header for committes

tony-tran:


src/Controller/EnMarche/AdherentController.php
```diff
@@ -287,12 +287,13 @@ public function listMyCommitteesAction(string $noResultMessage = null): Response
         ]);
     }
 
-    public function listCommitteesAlAction(): Response
+    public function listCommitteesAlAction(string $noResultMessage = null): Response
```
ottaviano: you can revert this

PR #2945, opened 2018-12-19
[ADI] ReportableInterface for IdeasWorkshop

Remg:


src/Entity/IdeasWorkshop/Thread.php
```diff
@@ -19,16 +22,10 @@
  */
 class Thread
```
ottaviano: interface Reportable missing

src/Controller/Api/ReportController.php
```diff
@@ -0,0 +1,89 @@
...
+
+        $errors = $validator->validate($reportCommand);
+
+        if (0 === $errors->count()) {
+            $handler->handle($reportCommand);
+
+            return new JsonResponse(null, Response::HTTP_CREATED);
+        }
+
+        $messages = [];
```
ottaviano: what about using the specific API normalizer: `ConstraintViolationListNormalizer` like this:
```php

$errors = $serializer->serialize($violations, 'jsonproblem')

return JsonResponse::fromJsonString($errors, Response::HTTP_BAD_REQUEST)
```

Remg: sweet :ok_hand: 

will do. thank you!  :heart:

PR #2975, opened 2018-12-26
fix Bind multi circos

Raulnet:


Raulnet:
[ch6147]

tests/Deputy/Subscriber/BindAdherentDistrictSubscriberTest.php
```diff
@@ -15,44 +15,52 @@
...
     {
         $adherent = $this->createAdherent();
-        $referentTag = $this->createReferentTag();
 
         $this->assertInstanceOf(ReferentTaggableEntity::class, $adherent);
         $this->assertSame(0, $adherent->getReferentTags()->count());
 
-        $this->manager->expects($this->once())->method('flush');
-        $this->districtRepository->expects($this->once())->method('findDistrictReferentTagByCoordinates')->willReturn($referentTag);
+        if ($count > 0) {
```
ottaviano: I think we can do:
```
if (0 < $count =  count($referentTags)) {
  // ...
}
```
and remove `int $count` argument from method signature, WDYT?

PR #3018, opened 2019-01-02
[ADI] Add /ideas endpoint to get ideas answers

JeremieSamson:


src/Entity/IdeasWorkshop/Question.php
```diff
@@ -28,6 +28,8 @@ class Question
     /**
      * @var int
      *
+     * @SymfonySerializer\Groups("idea_read")
```
ottaviano: do we really need to expose this info?

JeremieSamson: yes it's to match question's answer id

PR #3027, opened 2019-01-03
[ADI] Improve swagger

Remg:


src/Entity/IdeasWorkshop/ThreadCommentStatusEnum.php
```diff
@@ -15,4 +13,11 @@
         self::APPROVED,
         self::POSTED,
     ];
+
+    public const ALL_STATUSES = [
```
ottaviano: it can be a static public var and we can keep Enum parent class :) 
```
public static $all = [
  // ...
];
```

src/Swagger/SwaggerDecorator.php
```diff
@@ -39,8 +39,12 @@ public function normalize($object, $format = null, array $context = [])
                 $context = $resourceMetadata->getCollectionOperationAttribute('get', 'normalization_context', null, true);
 
                 if (isset($context['groups']) && ($groups = $context['groups']) && \is_array($groups)) {
-                    $docs = $this->overridePaginatedResponseFormat($docs, $resourceMetadata, $groups[0]);
+                    $group = $groups[0];
+                } else {
```
ottaviano: what about remove the `else` block and
```php
overridePaginatedResponseFormat($docs, $resourceMetadata, $group ?? null);
```

PR #3028, opened 2019-01-03
[ADI] #50 Volets "Mes idées" "Mes contributions"

Skytoooon:


web/apps/atelier-des-idees/src/redux/reducers/index.js
```diff
@@ -18,6 +19,7 @@ const rootReducer = history =>
         loading,
         modal,
         ideas,
+        myIdeas,
```
Haitaar: Manque myContributions

PR #3088, opened 2019-01-10
[ADI] #53 Name, need and categories filters

Haitaar:


Haitaar:
Rendre les select non searchable

PR #3090, opened 2019-01-10
[ADI] Add filter on category's name and order on contributors count

JeremieSamson:


src/Filter/ContributorsCountFilter.php
```diff
@@ -0,0 +1,59 @@
...
+            return;
+        }
+
+        $alias = $queryBuilder->getRootAliases()[0];
+
+        $queryBuilder
+            ->addSelect('COUNT(adherent) as HIDDEN contributorsCount')
+            ->leftJoin($alias.'.answers', 'answer')
+            ->leftJoin('answer.threads', 'thread')
+            ->leftJoin('thread.comments', 'threadComment')
```
ottaviano: why do you need to join comment table here?

src/Filter/ContributorsCountFilter.php
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        $description = [];
+        foreach ($this->properties as $property => $strategy) {
+            $description['contributorsCount'] = [
+                'property' => $property,
+                'type' => 'string',
+                'required' => false,
+                'swagger' => [
+                    'description' => 'Filter by contributors count.',
```
ottaviano: `Filter` -> `Order`

src/Filter/ContributorsCountFilter.php
JeremieSamson: count should be on comment table, last join was wrong, thanks :)

PR #3118, opened 2019-01-15
[ADI] Update getIdeaContributors method

JeremieSamson:


src/Repository/IdeaRepository.php
```diff
@@ -19,37 +21,46 @@ public function __construct(RegistryInterface $registry)
     }
 
     public function getIdeaContributors(Idea $idea): array
+    {
+        return array_unique(
```
ottaviano: `SORT_REGULAR` ?

PR #3143, opened 2019-01-16
add field nationality on donation entity

Raulnet:


src/Donation/DonationRequest.php
```diff
@@ -126,6 +132,7 @@ public function __construct(
         $this->country = 'FR';
         $this->setAmount($amount);
         $this->duration = $duration;
+        $this->nationality = 'FR';
```
ottaviano: `AppBundle\Address\Address::FRANCE`

src/Entity/Donation.php
```diff
@@ -114,7 +119,8 @@ public function __construct(
         string $emailAddress,
         PostAddress $postAddress,
         string $clientIp,
-        int $duration = PayboxPaymentSubscription::NONE
+        int $duration = PayboxPaymentSubscription::NONE,
+        string $nationality = 'FR'
```
ottaviano: `AppBundle\Address\Address::FRANCE`

src/Donation/DonationRequest.php
Raulnet: resolved

app/migrations/Version20190116162854.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20190116162854 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE donations ADD nationality VARCHAR(2) DEFAULT \'FR\' NOT NULL');
+        $this->addSql('ALTER TABLE donations CHANGE nationality nationality VARCHAR(2) NOT NULL');
```
JeremieSamson: I don't get the second line, was it generated by a diff ?

ottaviano: the first one adds a new column with a default value for existing rows and the second one remove this default value

PR #3151, opened 2019-01-17
[ADI] Add actions on Idea admin

JeremieSamson:


src/Controller/Admin/AdminIdeaController.php
```diff
@@ -22,10 +24,98 @@ class AdminIdeaController extends Controller
...
-        return $this->render('admin/ideas_workshop/idea_contributors.html.twig', [
+        return $this->render('admin/ideas_workshop/idea/idea_contributors.html.twig', [
             'idea' => $idea,
             'contributors' => $contributors = $repository->getContributors($idea),
             'contributors_count' => \count($contributors),
         ]);
     }
+
+    /**
+     * @Route("/{id}/enable", name="app_admin_idea_enable")
```
ottaviano: you can use `, method={"GET"}` of `@Route`

PR #3155, opened 2019-01-17
[Jecoute] Added Statistics pages

thomasbisignani:


src/Entity/Jecoute/SurveyQuestion.php
```diff
@@ -15,27 +17,27 @@
...
 {
+    use EntityIdentityTrait;
+
     /**
      * @ORM\Column(type="integer")
      * @ORM\Id
      * @ORM\GeneratedValue
      */
-    private $id;
+    protected $id;
```
Remg: not need if you use `EntityIdentityTrait`

src/Entity/Jecoute/Choice.php
```diff
@@ -29,7 +29,6 @@ class Choice
      * @var Question
      *
      * @ORM\ManyToOne(targetEntity="Question", inversedBy="choices")
-     * @ORM\JoinColumn(onDelete="CASCADE")
```
Remg: so we won't be able to delete any Question that has a Choice.
do you confirm? (we don't delete Questions, right?)

src/Entity/Jecoute/SurveyQuestion.php
thomasbisignani: This triggers a migration error because of the "unsigned" in the `EntityIdentityTrait`. So I override it.

templates/referent/surveys/data_answers_dialog_content.html.twig
```diff
@@ -0,0 +1,17 @@
+<table class="datagrid__table">
+    <thead>
+        <tr class="label__header--large form form__label">
+            <th>Réponse</th>
+            <th>Date</th>
+        </tr>
+        </thead>
```
JeremieSamson:
```suggestion
    </thead>
```

src/Jecoute/StatisticsProvider.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        $data['totalAnswered'] = $totalAnswered;
+
+        return $data;
+    }
+
+    private function calculateTotal(array $data, string $type): int
+    {
+        if (SurveyQuestionTypeEnum::SIMPLE_FIELD === $type) {
```
ottaviano: maybe:
```php
return array_sum(array_column(
    $data, 
    SurveyQuestionTypeEnum::SIMPLE_FIELD === $type ? 'textFieldsCount' : 'choicesCount'
));
```

src/Referent/SurveyExporter.php
```diff
@@ -38,7 +38,14 @@ public function exportAsJson(array $surveys): string
                         UrlGeneratorInterface::ABSOLUTE_URL
                     ),
                 ],
-            ];
+                'stats' => [
+                    'label' => "<span id='survey-stats-$i' class='btn btn--default'><i class='fa fa-bar-chart'></i></span>",
+                    'url' => $this->urlGenerator->generate(
+                        'app_referent_survey_stats', ['uuid' => $survey->getUuid()],
+                        UrlGeneratorInterface::ABSOLUTE_URL
```
ottaviano: do we really need the absolute URL for ajax call?

PR #3165, opened 2019-01-18
 'add choices during time "don"'

Raulnet:
'add choices during time "don"'

### DIFF
--- a/front/components/AmountChooser.js
+++ b/front/components/AmountChooser.js
@@ -57,6 +57,31 @@ export default class AmountChooser extends React.Component {
 
         return (
             <div className="amount-chooser">
+
+                <div className="amount-chooser__monthly form__radio">
+                    <input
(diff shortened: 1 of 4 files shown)


front/components/AmountChooser.js
```diff
@@ -57,6 +57,31 @@ export default class AmountChooser extends React.Component {
...
+
+                <div className="amount-chooser__monthly form__radio">
+                    <input
+                        type="radio"
+                        name="abonnement"
+                        id="donation-one-time"
+                        value="0"
+                        defaultChecked={!this.props.abonnement} />
+
+                    <label htmlFor="donation-one-time" id="donation-one-time_label">
```
ottaviano: do we really need to have `id="..."` on label?

Raulnet: not necessary, i removed them

PR #3181, opened 2019-01-22
[ADI] update getContributors query

JeremieSamson:


src/DataFixtures/ORM/LoadIdeaThreadCommentData.php
```diff
@@ -91,6 +93,13 @@ public function load(ObjectManager $manager)
             new \DateTime('-1 minute')
         );
 
+        $commentOfAdherent6OnIdeaWaste = ThreadComment::create(
+            Uuid::fromString(self::THREAD_COMMENT_09_UUID),
+            'Commentaire d`un adhérent',
```
thomasbisignani: Can you use `\'` or `"Commentaire d'un adhérent",` instead of the ` char please ?

src/Repository/IdeaRepository.php
```diff
@@ -33,33 +33,30 @@ public function getContributors(Idea $idea): array
...
-                    )
-                )
-            )
-            : 0
-        ;
+        $sqlIdeaContributors = <<<'SQL'
+        (
+            SELECT threadComment.author_id
+            FROM ideas_workshop_comment threadComment
+            LEFT JOIN ideas_workshop_thread thread ON thread.id = threadComment.thread_id
```
ottaviano: I think they should be only `INNER` joins here

PR #3187, opened 2019-01-22
[ADI] Urls: change /note with /proposition

Remg:


src/React/App/IdeasWorkshopApp.php
```diff
@@ -28,7 +28,7 @@ public function getRoutes(): array
             'contribute' => new Route('/atelier-des-idees/contribuer'),
             'propose' => new Route('/atelier-des-idees/proposer'),
             'create' => new Route('/atelier-des-idees/creer-ma-note'),
```
Haitaar:
```suggestion
            'create' => new Route('/atelier-des-idees/creer-ma-proposition'),
```

PR #3191, opened 2019-01-22
[ADI] Idea committee validation

Remg:


src/Validator/AuthorCommitteeValidator.php
```diff
@@ -0,0 +1,56 @@
...
+use AppBundle\Repository\CommitteeMembershipRepository;
+use Symfony\Component\Security\Core\Security;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+/**
+ * @Annotation
+ */
+class AuthorCommitteeValidator extends ConstraintValidator
```
ottaviano: `Author` -> `Member`?

src/Validator/AuthorCommitteeValidator.php
```diff
@@ -0,0 +1,56 @@
...
+        if (!$committee instanceof Committee) {
+            throw new UnexpectedTypeException($committee, Committee::class);
+        }
+
+        $adherent = $this->security->getUser();
+        if(!$adherent instanceof Adherent) {
+            throw new UnexpectedTypeException($adherent, Adherent::class);
+        }
+
+        if (!$this->committeeMembershipRepository->isAdherentInCommittee($adherent, $committee)) {
```
ottaviano: you can do 
```php
if (null === $user->getMembershipFor($committee)) {
  // ...
}
```

Remg: this will trigger a query fetching all committees...
i prefer using a query only counting what i need

what do you think ?

ottaviano: oh yes, good point 👍

PR #3197, opened 2019-01-22
Redirect user to create note page after connexion

ottaviano:


Remg:
```
Given i am on "/atelier-des-idees/proposer"
And I click on "Je propose mon idée"
Then I am redirected on "/login"
Then I go back
Then I am stuck on a page with a blank body
```
Sadly this test will pass .. :cry:
Is it front-related ?

PR #3211, opened 2019-01-23
[ADI] Add thread in count on total comments

JeremieSamson:


src/Repository/IdeaRepository.php
```diff
@@ -106,7 +106,27 @@ private function getCommentsContributors(Idea $idea): array
         ;
     }
 
-    public function countThreadComments(Idea $idea): int
+    public function countThreadAndThreadComments(Idea $idea): int
```
thomasbisignani: What do you think to rename this method by `countAllComments()` ?

PR #3214, opened 2019-01-23
[ADI] Remove emojis from text fields

illusionOfParadise:


src/Entity/EntityNameSlugTrait.php
```diff
@@ -51,8 +52,8 @@ public function getSlug(): string
 
     public function setName(string $name): void
     {
-        $this->name = $name;
-        $this->canonicalName = static::canonicalize($name);
+        $this->name = EmojisRemover::remove($name);
```
ottaviano: If we call `setName("")` with an empty string, then `EmojisRemover::remove($name);` will set `null` to the `$name` property...
what about doing this filter before persisting step with doctrine listener and not with a setter?

PR #3226, opened 2019-01-25
add new adminfiled zone && his filter

Raulnet:
add new adminfiled zone && his filter


src/Admin/JecouteSurveyAdmin.php
```diff
@@ -44,14 +45,32 @@ protected function configureFormFields(FormMapper $formMapper)
...
+        $datagridMapper
+            ->add('name')
+            ->add('author', null, [
+                'label' => 'Auteurs',
+                'show_filter' => true,
+            ])
+            ->add('author.referentTags', null, [
+                'label' => 'Zones',
+                'show_filter' => true,
+            ]);
```
thomasbisignani: CS indent : 

```php 
    $datagridMapper
        ->add('name')
        ->add('author', null, [
	     'label' => 'Auteurs',
	     'show_filter' => true,
	 ])
	 ->add('author.referentTags', null, [
	     'label' => 'Zones',
	     'show_filter' => true,
	 ])
    ;
```

PR #3229, opened 2019-01-25
bug la page de don a l adhesion ne s affiche

Raulnet:


templates/membership/donation.html.twig
```diff
@@ -5,45 +5,45 @@
 
 {% block javascripts %}
     <script type="text/javascript">
-        Kernel.onLoad(function() {
+        Kernel.onLoad(function () {
```
thomasbisignani: This change is not needed

Raulnet: remove

PR #3262, opened 2019-01-31
change filter to ModelAutocomplete

Raulnet:
change filter to ModelAutocomplete

### DIFF
--- a/src/Admin/JecouteSurveyAdmin.php
+++ b/src/Admin/JecouteSurveyAdmin.php
@@ -8,6 +8,7 @@ use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Route\RouteCollection;
+use Sonata\DoctrineORMAdminBundle\Filter\ModelAutocompleteFilter;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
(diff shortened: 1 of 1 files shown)


src/Admin/JecouteSurveyAdmin.php
```diff
@@ -49,13 +50,21 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
                 'show_filter' => true,
             ])
-            ->add('author.referentTags', null, [
+            ->add('author.referentTags', ModelAutocompleteFilter::class, [
                 'label' => 'Zones',
                 'show_filter' => true,
+                'field_options' => [
+                    'model_manager' => $this->getModelManager(),
+                    'admin_code' => $this->getCode(),
+                    'property' => 'name',
```
Remg: What about adding `'minimum_input_length' => 2,` to be able to search by department number ? 
**before:**
screenshot from 2019-01-31 12-48-20

**after:**
screenshot from 2019-01-31 12-49-12

src/Admin/JecouteSurveyAdmin.php
```diff
@@ -49,13 +50,21 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                ],
                 'show_filter' => true,
             ])
-            ->add('author.referentTags', null, [
+            ->add('author.referentTags', ModelAutocompleteFilter::class, [
                 'label' => 'Zones',
                 'show_filter' => true,
+                'field_options' => [
+                    'model_manager' => $this->getModelManager(),
+                    'admin_code' => $this->getCode(),
```
Remg: why are these 2 lines needed ?

src/Admin/JecouteSurveyAdmin.php
Raulnet: fixed

src/Admin/JecouteSurveyAdmin.php
Raulnet: removed

PR #3271, opened 2019-02-01
[ADI] Add comments cgu boolean

JeremieSamson:


features/api/adherents.feature
```diff
@@ -119,10 +119,12 @@ Feature:
       "emailAddress": "referent-75-77@en-marche-dev.fr",
       "firstName": "Referent75and77",
       "lastName": "Referent75and77",
-      "use_nickname": false
+      "use_nickname": false,
+      "comments_cgu_accepted": false
     }
     """
 
+  @wip
```
thomasbisignani: :man_cartwheeling:

JeremieSamson: :see_no_evil:

src/EventSubscriber/CommentsCguSubscriber.php
```diff
@@ -0,0 +1,43 @@
...
+        if (!in_array($route, ['api_threads_post_collection', 'api_thread_comments_post_collection'])) {
+            return;
+        }
+
+        $user = $this->security->getUser();
+
+        if ($user) {
+            $content = json_decode($event->getRequest()->getContent(), true);
+
+            $user->setCommentsCguAccepted(isset($content['comments_cgu_accepted']) && $content['comments_cgu_accepted']);
```
ottaviano:
```php
$user->setCommentsCguAccepted(filter_var(
  $content['comments_cgu_accepted'] ?? null, 
  FILTER_VALIDATE_BOOLEAN
));
```

src/Entity/IdeasWorkshop/BaseComment.php
```diff
@@ -126,6 +128,21 @@ public function enable(): void
         $this->enabled = true;
     }
 
+    /**
+     * @Assert\Callback(groups={"write"})
+     */
+    public function validateAuthorCgu(ExecutionContextInterface $context): void
+    {
+        if (!$this->getAuthor()->isCommentsCguAccepted() && $this->getAuthor()) {
```
ottaviano:
```php
!$this->getAuthor()->isCommentsCguAccepted() && $this->getAuthor()
```
here the second check is useless. If `$this->getAuthor()` returns null it will fail on first check because: `NULL->isCommentsCguAccepted()`.

this is a good order: 
```php
$this->getAuthor() && ...
```

src/Entity/IdeasWorkshop/BaseComment.php
```diff
@@ -126,6 +128,21 @@ public function enable(): void
         $this->enabled = true;
     }
 
+    /**
+     * @Assert\Callback(groups={"write"})
```
ottaviano: Sorry, but I think we can do better with 
```php
@Assert\Expression(
  "this.getAuthor().isCommentsCguAccepted()", 
  message="Vous devez cocher la case "CGU" afin de poster un commentaire."
)
```

PR #3277, opened 2019-02-04
[Test][Perf] Use DAMA\DoctrineTestBundle. (changes the way we use fixtures in tests)

Remg:
fyi: I introduced a new target :`make test-debug` that will run any test with `@group debug` in the testsuite.

Before:
- testsuite starts with an empty DB
- each scenario loads the specified fixtures (dropping the related table before, but not touching the others tables..)

Now:
- testsuite starts with all the fixtures loaded
- each scenario starts by starting a sql transaction
- each scenario ends by rollbacking the sql transaction

Note that:
- we do not need (and can't) load fixtures in phpunit tests anymore, since the DB is already loaded
- transactions cannot be open in phpunit tests because of how this bundle works.

Having all the fixtures loaded by default makes tests faster, but more importantly stronger: it showed up we had 2 bugs in our code, this PR also fixes them.

Maybe we should consider adopting the same strategy for our behat testsuite :)

### DIFF
--- a/src/Repository/AdherentRepository.php
+++ b/src/Repository/AdherentRepository.php
@@ -178,6 +178,7 @@ class AdherentRepository extends ServiceEntityRepository implements UserLoaderIn
             ->createQueryBuilder('a')
             ->select('COUNT(a.uuid)')
             ->where('a.status = :status')
+            ->andWhere('a.adherent = 1')
             ->setParameter('status', Adherent::ENABLED)
             ->getQuery()
         ;
--- a/src/Repository/EventRepository.php
+++ b/src/Repository/EventRepository.php
@@ -402,14 +402,22 @@ adherents.address_longitude AS adherent_address_longitude, adherents.position AS
 FROM events 
 LEFT JOIN adherents ON adherents.id = events.organizer_id
 LEFT JOIN committees ON committees.id = events.committee_id
-LEFT JOIN events_categories AS event_category ON event_category.id = events.category_id
+LEFT JOIN events_categories AS event_category
+    ON event_category.id = events.category_id
+    AND events.type = :event_type
+    AND event_category.status = :enabled
+LEFT JOIN citizen_action_categories AS citizen_action_category
+    ON citizen_action_category.id = events.category_id
+    AND events.type = :citizen_action_type
+    AND citizen_action_category.status = :enabled
 WHERE (events.address_latitude IS NOT NULL 
     AND events.address_longitude IS NOT NULL 
     AND (6371 * ACOS(COS(RADIANS(:latitude)) * COS(RADIANS(events.address_latitude)) * COS(RADIANS(events.address_longitude) - RADIANS(:longitude)) + SIN(RADIANS(:latitude)) * SIN(RADIANS(events.address_latitude)))) < :distance_max 
     AND events.begin_at > :today 
     AND events.published = :published
     AND events.status = :scheduled
-    AND event_category.status = :enabled)
+    AND (event_category.id IS NOT NULL OR citizen_action_category.id IS NOT NULL)
+    )
     __filter_query__ 
     __filter_category__
     __filter_type__ 
@@ -467,6 +475,8 @@ SQL;
         $query->setParameter('published', 1, \PDO::PARAM_INT);
         $query->setParameter('scheduled', BaseEvent::STATUS_SCHEDULED);
         $query->setParameter('enabled', BaseEventCategory::ENABLED);
+        $query->setParameter('event_type', BaseEvent::EVENT_TYPE);
+        $query->setParameter('citizen_action_type', BaseEvent::CITIZEN_ACTION_TYPE);
         $query->setParameter('first_result', $search->getOffset(), \PDO::PARAM_INT);
         $query->setParameter('max_results', $search->getMaxResults(), \PDO::PARAM_INT);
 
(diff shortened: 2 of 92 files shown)


src/Repository/AdherentRepository.php
```diff
@@ -178,6 +178,7 @@ public function countActiveAdherents(): int
             ->createQueryBuilder('a')
             ->select('COUNT(a.uuid)')
             ->where('a.status = :status')
+            ->andWhere('a.adherent = 1')
```
Remg: [This test](https://github.com/EnMarche/en-marche.fr/blob/testsuite/tests/Repository/AdherentRepositoryTest.php#L54) was passing because fixtures about simples users (`LoadUserData`) weren't loaded.

This new condition fixes this.

src/Repository/EventRepository.php
```diff
@@ -403,13 +403,15 @@ public function searchAllEvents(SearchParametersFilter $search): array
...
 LEFT JOIN events_categories AS event_category ON event_category.id = events.category_id
+LEFT JOIN citizen_action_categories AS citizen_action_category ON citizen_action_category.id = events.category_id
 WHERE (events.address_latitude IS NOT NULL 
     AND events.address_longitude IS NOT NULL 
     AND (6371 * ACOS(COS(RADIANS(:latitude)) * COS(RADIANS(events.address_latitude)) * COS(RADIANS(events.address_longitude) - RADIANS(:longitude)) + SIN(RADIANS(:latitude)) * SIN(RADIANS(events.address_latitude)))) < :distance_max 
     AND events.begin_at > :today 
     AND events.published = :published
     AND events.status = :scheduled
-    AND event_category.status = :enabled)
+    AND (event_category.status = :enabled OR citizen_action_category.status = :enabled)
```
Remg: This one was very tricky(long?) to debug :face_with_head_bandage: 

We aren't filtering citizen actions properly in production right now.

In short:
- the `events` table holds the `Event` and `CitizenAction` entity.
- `Event` entity has a ManyToOne relationship with `EventCategory`, through the `category` property.
- `CitizenAction` entity has a ManyToOne relationship with `CitizenActionCategory`, also through the `category` property.
- this method (`EventRepository::searchAllEvents()`) is meant to search through Events and CitizenActions at the same time.
- but the "category status" is filtered with a LEFT JOIN on EventCategories only.

Basically, our testsuite was passing because we were lucky that previous tests already loaded EventCategories with ids that are matching the CitizenActionCagories ids.


- cf [this test](https://github.com/EnMarche/en-marche.fr/blob/testsuite/tests/Repository/EventRepositoryTest.php#L50): 7 is including Citizen Actions.
- and cf [this test](https://github.com/EnMarche/en-marche.fr/blob/testsuite/tests/Controller/EnMarche/SearchControllerTest.php#L81): 6 is also including Citizen Actions.

tests/Controller/EnMarche/InteractiveControllerTest.php
```diff
@@ -50,11 +49,11 @@ public function testPurchasingPowerAction()
             'purchasing_power[friendFirstName]' => $purchasingPower->friendFirstName = 'Mylène',
             'purchasing_power[friendAge]' => '26',
             'purchasing_power[friendGender]' => $purchasingPower->friendGender = 'female',
-            'purchasing_power[friendPosition]' => '5',
+            'purchasing_power[friendPosition]' => '31',
```
Remg: Now all the fixtures are always loaded: this is normal that IDs of entities that share the same tables with other entities might change.

tests/Controller/EnMarche/AdherentControllerTest.php
```diff
@@ -73,12 +63,14 @@ public function testAuthenticatedAdherentCanSeeHisUpcomingAndPastEvents(): void
 
         $this->assertResponseStatusCode(Response::HTTP_OK, $this->client->getResponse());
 
-        $this->assertSame(4, $crawler->filter('.event-registration')->count());
+        $this->assertSame(6, $crawler->filter('.event-registration')->count());
```
Remg: related to the other fix !
Citizen actions were meant to appear here.

tests/Controller/EnMarche/ExplainerControllerTest.php
```diff
@@ -25,7 +23,7 @@ public function testSuccessfulActions(string $path)
         $crawler = $this->client->request(Request::METHOD_GET, $path);
 
         $this->assertResponseStatusCode(Response::HTTP_OK, $this->client->getResponse());
-        $this->assertSame(4, $crawler->filter('.explainer__articles li')->count());
+        $this->assertSame(4, $crawler->filter('.explainer__articles > ul > li')->count());
```
Remg: without this: we were just lucky that the total count of sub li elements were matching the final count.

This is the right way to count all the elements, now all king of articles are properly loaded.

Remg:
before:
screenshot from 2019-02-05 00-24-58

after:
screenshot from 2019-02-05 00-33-07

src/Repository/EventRepository.php
Remg: fixed :man_dancing:

ottaviano: :yeah: !

tests/Controller/EnMarche/ReferentControllerTest.php
```diff
@@ -464,14 +460,12 @@ protected function setUp()
...
-            LoadAdherentData::class,
-            LoadEventCategoryData::class,
-            LoadNewsletterSubscriptionData::class,
-            LoadReferentManagedUserData::class,
-        ]);
+        $this->init();
 
         $this->referentMessageRepository = $this->manager->getRepository(ReferentManagedUsersMessage::class);
+
+        // Disable republican silence for these tests
```
JeremieSamson: IMO `disableRepublicanSilence` is explicit enough

PR #3306, opened 2019-02-06
[Security] Stop saving target path for API request

ottaviano:


src/Security/Firewall/ExceptionListener.php
```diff
@@ -0,0 +1,22 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Http\Firewall\ExceptionListener as BaseExceptionListener;
+
+class ExceptionListener extends BaseExceptionListener
+{
+    protected function setTargetPath(Request $request)
+    {
+        if (
+            $request->isXmlHttpRequest()
+            || mb_strpos($request->getPathInfo(), '/api/') === 0
```
illusionOfParadise: What about using the parameter `%env(API_PATH_PREFIX)%`?
In this case if we change its value, we will have no need to modify this line.

ottaviano: YES! thanks

PR #3308, opened 2019-02-06
[ADI] Add position on themes

JeremieSamson:


app/migrations/Version20190205173350.php
```diff
@@ -0,0 +1,31 @@
...
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20190205173350 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE ideas_workshop_theme ADD position SMALLINT UNSIGNED NOT NULL');
+    }
+
+    public function postUp(Schema $schema)
```
JeremieSamson: I had to manually set logical position instead of an update because otherwise position can't be managed through the admin edition

src/Entity/IdeasWorkshop/Theme.php
```diff
@@ -86,11 +86,23 @@ class Theme implements EnabledInterface
...
+     *
+     * @Gedmo\SortablePosition
+     *
+     * @ORM\Column(type="smallint", options={"unsigned": true})
+     *
+     * @SymfonySerializer\Groups("theme_read")
+     */
+    private $position;
+
+    public function __construct(string $name = '', string $imageName = null, bool $enabled = false, $position = 1)
```
Remg: `int $position = 1`

src/Entity/IdeasWorkshop/Theme.php
```diff
@@ -86,11 +86,23 @@ class Theme implements EnabledInterface
      */
     protected $imageName;
 
-    public function __construct(string $name = '', string $imageName = null, bool $enabled = false)
+    /**
+     * @Assert\GreaterThanOrEqual(1)
+     *
+     * @Gedmo\SortablePosition
+     *
+     * @ORM\Column(type="smallint", options={"unsigned": true})
```
thomasbisignani: Can you declare this annotation first ? ( We declare `@ORM` annotations always first )

Remg: hue

src/Entity/IdeasWorkshop/Theme.php
```diff
@@ -86,11 +86,23 @@ class Theme implements EnabledInterface
...
 
-    public function __construct(string $name = '', string $imageName = null, bool $enabled = false)
+    /**
+     * @ORM\Column(type="smallint", options={"unsigned": true})
+     *
+     * @Assert\GreaterThanOrEqual(1)
+     *
+     * @Gedmo\SortablePosition
+     *
+     * @SymfonySerializer\Groups("theme_read")
```
ottaviano: do we really need to expose this value? 
Using it to sort the Theme collection is not enough?

JeremieSamson: I'll check with front dev, I guess you are right about this, will probably not be needed

PR #3321, opened 2019-02-07
Add a few more copy updates 🔥

marineb:


web/apps/atelier-des-idees/src/pages/Propose/PreProposal/index.jsx
```diff
@@ -27,6 +27,9 @@ class PreProposal extends React.PureComponent {
                         >
 							Suivre l'avancée du programme
                         </a>
+                        <p className="pre-proposal__container__text">
+              Ou <a href="/atelier-des-idees/contribuer">contribuez aux propositions en cours d'élaboration</a> par les autres marcheurs.
```
Remg: you should use the generated path, without hardcoding it
`href="{{ path('react_app_ideas_workshop_contribute') }}"`

PR #3342, opened 2019-02-08
[ADI] Update idea admin

JeremieSamson:


templates/admin/ideas_workshop/idea/_status.html.twig
```diff
@@ -3,7 +3,7 @@
 {% elseif object.isDraft %}
     <span class="label label-warning">{{ 'ideas_workshop.status.draft'|trans }}</span>
 {% elseif object.isPending %}
-    <span class="label label-primary">{{ 'ideas_workshop.status.pending'|trans }}</span>
+    <span class="label label-primary">{{ 'ideas_workshop.status.pending'|trans }}</span> {{ object.daysBeforeDeadline }} jour{{ 1 != object.daysBeforeDeadline ? 's' : '' }}
```
thomasbisignani: What about using a translation for the pluralization ?

PR #3343, opened 2019-02-08
refacto design admin referent organigramme

Raulnet:
refacto design admin referent organigramme

### DIFF
--- a/templates/admin/organizational_chart/admin_list.html.twig
+++ b/templates/admin/organizational_chart/admin_list.html.twig
@@ -1,11 +1,6 @@
 {% extends '@SonataAdmin/CRUD/base_list.html.twig' %}
 
-{% import 'referent/organizational_chart_macros.html.twig' as oc %}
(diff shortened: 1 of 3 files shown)


templates/admin/organizational_chart/admin_list.html.twig
```diff
@@ -1,10 +1,9 @@
 {% extends '@SonataAdmin/CRUD/base_list.html.twig' %}
 
-{% import 'referent/organizational_chart_macros.html.twig' as oc %}
+{% import 'admin/organizational_chart/chart_macros.html.twig' as oc %}
 
 {% block stylesheets %}
     {{ parent() }}
```
ottaviano: you can remove empty block 
```
{% block toto %}
    {{ parent() }}
{% endblock %}
```

Raulnet: removed

PR #3349, opened 2019-02-09
[AdherentMessage] Add Message action buttons if it was synchronized

ottaviano:
[ch5875]


front/services/api/ReqwestApiClient.js
```diff
@@ -208,18 +208,10 @@ export default class ReqwestApiClient {
         });
     }
 
-    getCitizenProjectLandingPage(city, callback) {
```
ottaviano: this function is unused

PR #3356, opened 2019-02-11
Replace Symfony/Symfony by Symfony/* packages

ottaviano:
- rename `/web` dir to `/public`
- replace `symfony/symfony` package by `symfony/*` independantes packages 
- add `symfony/flex` to the project ([doc](https://symfony.com/doc/current/setup/flex.html#upgrading-existing-applications-to-flex))
- apply new dir structure for using Flex
- remove `app/AppKernel.php` class
- remove `incenteev/composer-parameter-handler` (not need anymore, as it generates a `parameters.yml` file)
- replace `sensio/distribution-bundle` by `sensiolabs/security-checker` (newest version of the same lib)
- remove `sensio/generator-bundle` (we can add maker bundle in the future)

Note: you juste review the first and second commits, thrid commit rename `/web` to `/public`


bin/console
```diff
@@ -1,28 +1,41 @@
...
+    putenv('APP_DEBUG='.$_SERVER['APP_DEBUG'] = $_ENV['APP_DEBUG'] = '0');
+}
+
+require dirname(__DIR__).'/config/bootstrap.php';
+
+if ($_SERVER['APP_DEBUG']) {
+    // if you don't want to setup permissions the proper way, just uncomment the following PHP line
+    // read http://symfony.com/doc/current/setup.html#checking-symfony-application-configuration-and-setup
+    // for more information
+    //umask(0000);
```
Raulnet: is it important to keep this commentary ?

src/Kernel.php
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
+    {
+        // Feel free to remove the "container.autowiring.strict_mode" parameter
+        // if you are using symfony/dependency-injection 4.0+ as it's the default behavior
+
+        // $container->setParameter('container.autowiring.strict_mode', true);
+        $container->setParameter('container.dumper.inline_class_loader', true);
```
Raulnet: is it necessary to keep the commentaries ?

phpunit.xml.dist
```diff
@@ -2,14 +2,14 @@
...
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
          backupGlobals="false"
          colors="true"
-         bootstrap="vendor/autoload.php"
+         bootstrap="config/bootstrap.php"
 >
     <php>
         <ini name="error_reporting" value="-1" />
-        <server name="KERNEL_CLASS" value="AppKernel" />
+        <env name="APP_ENV" value="test" />
```
thomasbisignani: Is it really needed ? :thinking:

Remg: I would say yes since it isn't defined in the `.env.test` file

src/Entity/AbstractTranslatableEntity.php
```diff
@@ -2,20 +2,18 @@
 
 namespace AppBundle\Entity;
 
-use A2lix\I18nDoctrineBundle\Doctrine\ORM\Util\Translatable;
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Symfony\Component\Validator\Constraints as Assert;
 
-trait EntityTranslatableTrait
+abstract class AbstractTranslatableEntity
```
Remg: Why use an abstract class for this translatable behavior ?

I would prefer an interface (forcing implementation to defines several methods) + a trait for the `translate()` method (that would provide the method needed in the interface)

Why ? because using inheritance would force entities for this specific translation behavior.
(What if i want to translate CitizenActions and not Events?)

phpunit.xml.dist
ottaviano: Yes, it's needed, for leading .env.test file.

.env
```diff
@@ -0,0 +1,61 @@
...
+SSL_ENCRYPTION_KEY='PT/cLwGkQqwF5exN/CkkV5oUkBemNyhi2dXDHy2N+Nc='
+MAXMIND_LICENSE_KEY=
+MAXMIND_USER_ID=
+APP_EMAIL_SUBSCRIPTIONS_WEBHOOK_SECRET='mailchimp_secret'
+MAILCHIMP_API_KEY=
+MAILCHIMP_MEMBER_LIST_ID=
+MAILCHIMP_MEMBER_GROUP_INTEREST_GROUP_ID=
+MAILCHIMP_MEMBER_INTEREST_INTEREST_GROUP_ID=
+MAILCHIMP_CAMPAIGN_FOLDER_IDS='{}'
+MAILCHIMP_TEMPLATE_IDS='{}'
```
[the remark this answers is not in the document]
ottaviano: @illusionOfParadise, that is a really nice catch 👍

src/Entity/AbstractTranslatableEntity.php
ottaviano: @Remg , to able to replace `symfony/symfony` by independent packages `symfony/*` I should update 
`a2lix/I18nDoctrineBundle` to the release 0.2. The previous version (0.1) has `symfony/symfony` as the dependence and doesn't allow to remove our `symfony/symfony` package. Unfortunately, this new release (0.2) introduces a little BC break. 
I replaced by Abstract class, because our entity (Profile for example) that uses `EntityTranslatableTrait` which uses Translatable trait wasn't identified by the bundle's Listener as a Translatable Entity.

config/bootstrap.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+use Symfony\Component\Dotenv\Dotenv;
+
+require \dirname(__DIR__).'/vendor/autoload.php';
+
+// Load cached env vars if the .env.local.php file exists
+// Run "composer dump-env prod" to create it (requires symfony/flex >=1.2)
```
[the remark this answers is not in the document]
ottaviano: agree, removed 👍

src/Kernel.php
ottaviano: no, I removed it

bin/console
ottaviano: no, removed 👍

ottaviano:
Let's go into a new era 🚀

PR #3360, opened 2019-02-12
[ADI] Add idea.comments_count column

ottaviano:


src/Entity/IdeasWorkshop/Idea.php
```diff
@@ -639,6 +642,29 @@ public function decrementVotesCount(int $increment = 1): void
         $this->votesCount -= $increment;
     }
 
+    /**
+     * @SymfonySerializer\Groups({"idea_list_read", "idea_read"})
+     */
+    public function getNewCommentsCount(): int
```
illusionOfParadise: Why did you add this method that returns `$commentsCount` value as `getCommentsCount `?

src/EntityListener/IdeaThreadListener.php
```diff
@@ -0,0 +1,36 @@
...
+class IdeaThreadListener
+{
+    public function prePersist(Thread $thread): void
+    {
+        if ($thread->isEnabled()) {
+            $thread->getIdea()->incrementCommentsCount();
+        }
+    }
+
+//    public function preUpdate(Thread $thread, PreUpdateEventArgs $event): void
```
illusionOfParadise: Should be deleted if not used.

ottaviano: nop, it's used.

ottaviano: :trollface:

PR #3364, opened 2019-02-12
Set collection instance to Adherent tags property

ottaviano:
https://sentry.io/en-marche-i7/staging-en-marchefr/issues/884962976/?referrer=slack


thomasbisignani:
:1st_place_medal:

PR #3383, opened 2019-02-14
fix datetime notificationMessage

Raulnet:


tests/Mailer/Message/AbstractEventMessageTest.php
```diff
@@ -73,4 +90,24 @@ protected function createCommitteeFeedItemMock(Adherent $author, string $content
 
         return $mock;
     }
+
+    protected function createCitizenActionMock(
+        string $name,
+        string $beginAt,
+        string $street,
+        string $cityCode,
+        $timeZone = \AppBundle\Address\GeoCoder::DEFAULT_TIME_ZONE
```
JeremieSamson: Why not `use AppBundle\Address\GeoCoder;` and call `GeoCoder::DEFAULT_TIME_ZONE` here ?

tests/Mailer/Message/AbstractEventMessageTest.php
```diff
@@ -13,13 +14,22 @@
...
 {
-    protected function createEventMock(string $name, string $beginAt, string $street, string $cityCode, ?string $committeeName = null): Event
-    {
+    protected function createEventMock(
+        string $name,
+        string $beginAt,
+        string $street,
+        string $cityCode,
+        ?string $committeeName = null,
+        $timeZone = \AppBundle\Address\GeoCoder::DEFAULT_TIME_ZONE
```
Remg: typehint missing

tests/Mailer/Message/AbstractEventMessageTest.php
Remg: typehint missing

Remg: Why not hardcoding `'Europe/Paris'` directly here, instead of using a constant ?

This is a testsuite, we should test expected values also

Raulnet: harcoded

Raulnet: fixed

tests/Mailer/Message/AbstractEventMessageTest.php
Raulnet: fixed

PR #3388, opened 2019-02-14
add nationality on adherent entity

Raulnet:


src/Entity/Adherent.php
```diff
@@ -1412,4 +1420,9 @@ public function setDisplayMedia(bool $displayMedia): void
     {
         $this->displayMedia = $displayMedia;
     }
+
+    public function getNationality(): ?string
```
ottaviano: what about a setter?

src/Membership/MembershipRequest.php
```diff
@@ -91,6 +91,12 @@ class MembershipRequest implements MembershipInterface
      */
     public $conditions;
 
+    /**
```
ottaviano: `@var string|null`

templates/adherent/profile.html.twig
```diff
@@ -82,6 +82,13 @@
         <div class="space--10-0"></div>
 
         <div class="l__row l__row--top l__row--h-stretch l__mobile--col">
+
+            <div class="l__row l__row--h-stretch b__nudge--bottom-tiny form__radio-button">
+                {{ form_widget(form.nationality, { attr: { class: 'register__form__profession' } }) }}
```
ottaviano: why do you need `register__form__profession` class?

templates/adherent/profile.html.twig
```diff
@@ -82,6 +82,13 @@
         <div class="space--10-0"></div>
 
         <div class="l__row l__row--top l__row--h-stretch l__mobile--col">
+
+            <div class="l__row l__row--h-stretch b__nudge--bottom-tiny form__radio-button">
```
ottaviano: why do you need `form__radio-button`? it's not a radio, but select element, no?

templates/membership/join.html.twig
```diff
@@ -98,6 +98,10 @@
                                 {{ form_errors(form.password) }}
                             </div>
                         {% endif %}
+                        <div class="l__row l__row--h-stretch register__form--full">
+                            {{ form_errors(form.nationality) }}
```
ottaviano: in this template, the order of block is: `widget, errors`
but you display errors first and widget after. there is a real reason for that?

translations/validators.fr.yml
```diff
@@ -29,6 +29,9 @@ common.email.repeated: Les adresses email ne correspondent pas.
 common.country.not_blank: Cette valeur ne doit pas être vide.
 common.country.invalid: Ce pays n'est pas valide.
 
+common.nationality.not_blank: Cette valeur ne doit pas être vide.
```
ottaviano: Why do you need to custom the default NotBlank constrain's message?

src/Entity/Adherent.php
```diff
@@ -393,7 +398,8 @@ public static function create(
         string $registeredAt = 'now',
         ?array $tags = [],
         ?array $referentTags = [],
-        ?array $mandates = null
+        ?array $mandates = null,
```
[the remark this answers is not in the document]
Raulnet: edited

src/Entity/Adherent.php
Raulnet: not need in this case

src/Membership/MembershipRequest.php
Raulnet: added

templates/adherent/profile.html.twig
Raulnet: edited

templates/adherent/profile.html.twig
Raulnet: removed

translations/validators.fr.yml
Raulnet: All NotBlank's messages are override on object MembershipRequest, I think it's for the translation

templates/membership/join.html.twig
Raulnet: reversed

templates/membership/register.html.twig
```diff
@@ -60,6 +60,12 @@
                     {{ form_widget(form.password, {attr: {placeholder: 'Mot de passe', autocomplete: "new-password"} }) }}
                 </li>
 
+                <li>
+                    {{ form_label(form.nationality,'Nationalité',{ label_attr: {class: 'visually-hidden'} }) }}
```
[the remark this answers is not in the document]
Raulnet: all file is fixed

PR #3392, opened 2019-02-15
Manage mandates in the admin and 'elected' idea author category when user changes mandates itself

illusionOfParadise:


src/Repository/IdeaRepository.php
```diff
@@ -226,6 +226,7 @@ public function updateAuthorCategoryForIdeasOf(Adherent $adherent): void
             $qb->update()
                 ->set('idea.authorCategory', $qb->expr()->literal($categoryType))
                 ->where('idea.author = :adherent')
+                ->andWhere('idea.enabled = 0 OR idea.enabled = 1')
```
Raulnet: In this case it may be better to use 'IN' instead of 'OR' ?
`
$qb->update()
                ->set('idea.authorCategory', $qb->expr()->literal($categoryType))
                ->where('idea.author = :adherent')
                ->andWhere($qb->expr()->in(''idea.enabled', [0, 1]))
                ->setParameter('adherent', $adherent)
                ->getQuery()
                ->execute()
            ;`

illusionOfParadise: We have no need of this line, because we manage Doctrine 'enabled' filter here https://github.com/EnMarche/en-marche.fr/pull/3392/files#diff-bc12631958b44351bc32d009a69a0ca0R42

templates/adherent/profile.html.twig
```diff
@@ -21,6 +21,7 @@
                     return "{{ 'adherent.form.mandate.no_result'|trans }}";
                 },
             });
+            $('.multi_mandate').prop('required',false);
```
ottaviano: I think we need to fix this script instead:
https://github.com/EnMarche/en-marche.fr/blob/master/templates/membership/_mandates_script.html.twig#L11

templates/membership/_mandates_script.html.twig
```diff
@@ -3,12 +3,12 @@
         var formMandates = $('#form__mandates');
         var formElected = $('#{{ form.elected.vars.id }}');
         var selectMandates = $('#{{ form.mandates.vars.id }}');
+        selectMandates.attr('required', {{ form.mandates.vars.required }});
```
ottaviano: it's not correct, because now this field is always optional, but it should be required when `elected` checkbox is checked.

I think you can replace:

```js
selectMandates.attr('required', isChecked);
```
by
```js
selectMandates.prop('required', isChecked);
```

PR #3403, opened 2019-02-19
[AdherentMessage] Migrate Deputy message to Mailchimp

ottaviano:


app/config/parameters.yml.dist
```diff
@@ -53,8 +53,6 @@ parameters:
 
     env(LEGISLATIVES_PLATFORM): https://www.google.fr
 
-    env(ENABLE_SILENCE_RULE):   'false'
```
ottaviano: unused var env

PR #3407, opened 2019-02-20
Remove app dir

ottaviano:
Step 2: 
- remove `/app` directory
- move migration to `/src/Migrations` directory
- update `doctrine/migrations` lib to 2.0

to simplify the code review you could review each commit independently


MartinSogno:
Bon travail!

PR #3413, opened 2019-02-21
[Institutional events] Added invitations

thomasbisignani:
This PR adds  : 

- [x] Invitations emails fields and send email on create
- [x] Add BCC option for the emails
- [x] Edit an institutional event from referent space
- [x] Send email on edit
- [x] Delete an institutional event from referent space

### DIFF
--- a/src/Referent/ManagedInstitutionalEventsExporter.php
+++ b/src/Referent/ManagedInstitutionalEventsExporter.php
@@ -3,9 +3,17 @@
 namespace AppBundle\Referent;
 
 use AppBundle\Entity\InstitutionalEvent;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 class ManagedInstitutionalEventsExporter
 {
+    private $urlGenerator;
+
+    public function __construct(UrlGeneratorInterface $urlGenerator)
+    {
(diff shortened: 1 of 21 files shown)


src/Referent/ManagedInstitutionalEventsExporter.php
```diff
@@ -19,6 +27,21 @@ public function exportAsJson(array $managedInstitutionalEvents): string
                 'category' => $institutionalEvent->getCategoryName(),
                 'postalCode' => $institutionalEvent->getPostalCode(),
                 'organizer' => $institutionalEvent->getOrganizer()->getPartialName(),
+                'invitationsCount' => $institutionalEvent->getInvitationsCount(),
+                'edit' => [
+                    'label' => "<span class='btn btn--default'><i class='fa fa-edit'></i></span>",
+                    'url' => $this->urlGenerator->generate(
+                        'app_referent_institutional_events_edit',
+                        ['uuid' => $institutionalEvent->getUuid()]
```
ottaviano: do we need to generate absolute URL here (with the host)?

thomasbisignani: No, absolute path is enough :wink:

PR #3415, opened 2019-02-21
Add radio button to the proxy form

JeremieSamson:
Add radio button to the proxy form to let user add a number of available proxy. 

Had to add jquery in form to implement specific condition :
- If user vote in France, it's 2 max (one in France and the other elsewhere)
- if user vote elsewhere than France, it's 3 max 

When France is chosen

lieu_france

When another country than France is chosen

lieu_belgique

### DIFF
--- a/front/style/pages/_procuration.scss
+++ b/front/style/pages/_procuration.scss
@@ -175,6 +175,17 @@
             }
         }
     }
+
+    .form {
+        &__radio-button {
+            input[type='radio'] {
+                &:checked + label:before {
+                    border-color: $gray--dark;
+                    background: $gray--dark;
+                }
+            }
+        }
+    }
 }
 
 .procuration-manager {
--- a/src/Entity/ProcurationProxy.php
+++ b/src/Entity/ProcurationProxy.php
@@ -287,6 +287,26 @@ class ProcurationProxy
      */
     public $recaptcha = '';
 
+    /**
+     * @var int
+     *
+     * @Assert\Range(
+     *     min=1,
+     *     max=3,
+     *     minMessage="procuration.vote_country.conditions",
+     *     maxMessage="procuration.vote_country.conditions",
+     *     groups={"front"}
+     * )
+     * @Assert\Expression(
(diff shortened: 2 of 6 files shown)


front/style/pages/_procuration.scss
```diff
@@ -126,6 +126,13 @@
                     min-width: 0;
                 }
             }
+
+            #nbOfProxiesProposed {
+                label:before {
+                    content: none;
```
ottaviano: out of curiosity why we need to reset the content?

src/Entity/ProcurationProxy.php
```diff
@@ -287,6 +287,26 @@ class ProcurationProxy
...
+     *     maxMessage="procuration.vote_country.conditions",
+     *     groups={"front"}
+     * )
+     * @Assert\Expression(
+     *     "(this.getVoteCountry() == 'FR' and value <= 2) or (this.getVoteCountry() != 'FR' and value <= 3)",
+     *     message="procuration.vote_country.conditions",
+     *     groups={"front"}
+     * )
+     *
+     * @ORM\Column(type="integer",  options={"unsigned": true, "default": 1})
```
ottaviano: what about `smallint`?

src/Entity/ProcurationProxy.php
```diff
@@ -287,6 +287,26 @@ class ProcurationProxy
...
+     */
+    public $nbOfProxiesProposed = 1;
```
ottaviano: what about `(total|count)ProposedProxies` or just `(total|count)Proxies`?

front/style/pages/_procuration.scss
JeremieSamson: otherwise the global css apply for checkboxes and it looked like this : 

capture du 2019-02-22 11-27-02

but we worked on it with @thomasbisignani and now it looks better, just updated PR details with new css

src/Entity/ProcurationProxy.php
JeremieSamson: `proxies_count` it is

PR #3416, opened 2019-02-21
[ADI] Add endpoint to find ideas with searched key words

illusionOfParadise:


src/Filter/OrTextSearchFilter.php
```diff
@@ -0,0 +1,72 @@
...
+        string $property,
+        $value,
+        QueryBuilder $queryBuilder,
+        QueryNameGeneratorInterface $queryNameGenerator,
+        string $resourceClass,
+        string $operationName = null
+    ): void {
+        $property = str_replace(self::PROPERTY_SUFFIX, '', $property);
+
+        // to sure that filter is applied only to an existing string field
```
JeremieSamson:
```suggestion
        // to be sure that filter is applied only to an existing string field
```

src/Filter/OrTextSearchFilter.php
```diff
@@ -0,0 +1,72 @@
...
+use Doctrine\ORM\QueryBuilder;
+
+final class OrTextSearchFilter extends AbstractContextAwareFilter
+{
+    const PROPERTY_SUFFIX = '_contains';
+    const IGNORED_WORDS = ['avec', 'après', 'avant', 'depuis', 'jusque', 'jusqu’à', 'jusqu\'à', 'pendant', 'à côté',
+        'à droite', 'à gauche', 'au-delà', 'au-dessous', 'au-dessus', 'à travers', 'derrière', 'en dehors', 'contre',
+        'd\'après', 'd’après', 'en face', 'chez', 'hors', 'loin', 'par', 'près', 'sous', 'sur', 'vers', 'grâce à',
+        'malgré', 'pour', 'contre', 'entre', ' sauf', 'excepté', 'par', 'parmi', 'sans', 'selon', 'dans', 'en',
+        'le', 'la', 'les', 'au', 'aux', 'un', 'une', 'dès', 'des', 'du', 'de', 'a', 'à', ];
```
JeremieSamson: `, ]` at the end :thinking:  ?

illusionOfParadise: It's `php-cs-fixer` that makes this.

src/Filter/OrTextSearchFilter.php
```diff
@@ -0,0 +1,72 @@
...
+        if (!array_key_exists($property, $this->properties)
+            || !$this->getClassMetadata($resourceClass)->hasField($property)
+            || 'string' !== $this->getClassMetadata($resourceClass)->getTypeOfField($property)) {
+            return;
+        }
+
+        $value = str_replace(array_map(function (string $word) {
+            return ' '.$word.' ';
+        }, self::IGNORED_WORDS), ' ', ' '.$value.' ');
+        $values = array_filter(explode(' ', trim($value)));
```
ottaviano: an alternative: 
```php
array_filter(
  explode(' ', preg_replace('('.implode('|', self::INGORED_WORDS).')', '', $value))
);
```

illusionOfParadise: It will not give a desired result: the ignored words will be removed even from the middle of the word. For example, `Père Lachaise` will become `Père chaise`, that is not good.

PR #3431, opened 2019-02-27
[Front DX] Add new Symfony env dev_front

ottaviano:


bin/mac-docker-install.sh
```diff
@@ -0,0 +1,101 @@
...
+    fi
+fi
+
+# Remove virtual machine with `default` name
+VIRTUAL_MACHINES=`docker-machine ls --quiet`
+
+if [[ ${VIRTUAL_MACHINES} =~ "default" ]]
+then
+    echo "Default virtual matched was found"
+    echo ""
```
thomasbisignani: Instead of using`echo ""` instructions for a blank line, what do you think to add `\n` ?

Example : `echo "Default virtual matched was found\n"` (Maybe with the `-e` option)

docker/dev/nginx.conf
```diff
@@ -1,21 +1,17 @@
...
+    multi_accept on;
 }
 
 http {
-    log_format main '$http_x_forwarded_for ($remote_addr) - $remote_user [$time_local] '
-        '"$request" $status $body_bytes_sent "$http_referer" '
-        '"$http_user_agent"' ;
-
-    access_log /dev/stdout main;
+    access_log off;
```
thomasbisignani: Why turn off the access log for the dev configuration ? :thinking:

bin/mac-docker-install.sh
JeremieSamson: yes, you must add `-e` ;)

docker/dev/nginx.conf
ottaviano: @thomasbisignani, I try to optimize a bit nginx, because it isn't very fast on dev.
IMO this log is useless, have you ever had the need to see them?

bin/mac-docker-install.sh
ottaviano: done 👍

PR #3434, opened 2019-02-27
Add script to clean new migration

JeremieSamson:
PR closed, please check instead => #3448


scripts/new_migration_clean.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+MIGRATION_DIRECTORY='app/migrations'
+LAST_MIGRATION_FILE_NAME=$(ls -t $MIGRATION_DIRECTORY | head -n1)
+FILE_PATH=${MIGRATION_DIRECTORY}/${LAST_MIGRATION_FILE_NAME}
+
+if grep -q ' : void' $FILE_PATH; then
```
Remg: i would run the target `phpcsfix` before this script to avoid this CS step :)
wdyt?

thomasbisignani:
IMO we can check if the last migration file is a new file to avoid the useless checks. 

You can check if the file is present in the `git ls-files --others --exclude-standard` output.
WDYT ?

PR #3438, opened 2019-02-27
[ADMIN] Change report admin (status and actions)

illusionOfParadise:


src/Admin/ReportAdmin.php
```diff
@@ -255,10 +255,6 @@ protected function configureListFields(ListMapper $listMapper)
                 'label' => 'Date du signalement',
                 'format' => self::DATE_FORMAT,
             ])
-            ->add('details', null, [
```
ottaviano: why do you remove this column? it seems to me that this isn't what is requested, no?

src/Controller/Admin/AdminIdeaController.php
```diff
@@ -54,15 +58,21 @@ public function membersAction(Idea $idea, IdeaRepository $repository): Response
...
+        EntityManagerInterface $manager,
+        ReportRepository $reportRepository,
+        ReportManager $reportManager
+    ): Response {
         if ($idea->isEnabled()) {
-            $this->addFlash('warning', 'La proposition a déjà été démodéré');
+            $this->addFlash('warning', 'La proposition a déjà été démodérée.');
         } else {
             $idea->setEnabled(true);
+            $this->resolveReportsFor(IdeaReport::class, $idea, $reportRepository, $reportManager);
```
ottaviano: what about using event_dispatcher, or messageBus for doing the job?

src/Controller/Admin/AdminThreadCommentController.php
```diff
@@ -22,17 +25,25 @@
...
+        Request $request,
+        ThreadComment $comment,
+        ObjectManager $manager,
+        EventDispatcherInterface $dispatcher,
+        ReportRepository $reportRepository,
+        ReportManager $reportManager
+    ): Response {
         $comment->disable();
 
         $dispatcher->dispatch(Events::THREAD_COMMENT_DISABLE, new GenericEvent($comment));
```
ottaviano: for example, here you can listen to THREAD_COMMENT_DISABLE event to resolve related reports

templates/admin/report/show_details.html.twig
```diff
@@ -0,0 +1,11 @@
+{% extends '@SonataAdmin/CRUD/base_show_field.html.twig' %}
+
+{% block field_value %}
+    {% if object.subject is instanceof('AppBundle\\Entity\\EntitySoftDeletedInterface') and object.subject.getDeletedAt() is not null %}
```
ottaviano: what about including `partial/_details.html.twig` here to avoid duplicate code and fix the little bug with `(dés)approuvé` label?

src/Admin/ReportAdmin.php
illusionOfParadise: Actually, it was requested. We will show this information only on the show page and not on the list ont.

templates/admin/report/show_details.html.twig
illusionOfParadise: It's not a bug. On the report show page we need show information only if an reported object is approved and on the thread/comment page the both status should be shown.

PR #3446, opened 2019-03-04
[ADI] Add extend action for ideas

JeremieSamson:
Add extend action as `PUT` on the `Idea` class
This action should extends to 10 days a published or finalized idea
Idea will be back as `published` if `finalized`
Admin must be able to do it as many times he wants (no date update, no incrementation) 
An email is send when user extend his idea, but none when admin do it

### DIFF
--- a/app/migrations/Version20190304170145.php
+++ b/app/migrations/Version20190304170145.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20190304170145 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE 
+          ideas_workshop_idea 
+        ADD 
+          extensions_count SMALLINT UNSIGNED NOT NULL, 
+        ADD 
+          last_extension_date DATE DEFAULT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE ideas_workshop_idea DROP extensions_count, DROP last_extension_date');
+    }
+}
--- a/src/Controller/Admin/AdminIdeaController.php
+++ b/src/Controller/Admin/AdminIdeaController.php
@@ -106,4 +106,22 @@ class AdminIdeaController extends Controller
 
         return $this->redirectToRoute('admin_app_ideasworkshop_idea_list');
     }
+
+    /**
+     * @Route("/{id}/extend", name="app_admin_idea_extend", methods={"GET"})
+     */
+    public function extendAction(Idea $idea, ObjectManager $manager): Response
+    {
+        if ($idea->isDraft()) {
+            $this->addFlash('warning', 'La proposition est un brouillon');
+        } else {
+            $idea->publish();
+
+            $manager->flush();
+
+            $this->addFlash('success', 'La proposition a bien été prolongée');
+        }
+
+        return $this->redirectToRoute('admin_app_ideasworkshop_idea_list');
+    }
 }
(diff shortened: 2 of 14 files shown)


src/Controller/Admin/AdminIdeaController.php
```diff
@@ -106,4 +106,18 @@ public function finalizeAction(Idea $idea, ObjectManager $manager): Response
 
         return $this->redirectToRoute('admin_app_ideasworkshop_idea_list');
     }
+
+    /**
+     * @Route("/{id}/extend", name="app_admin_idea_extend", methods={"GET"})
+     */
+    public function extendAction(Idea $idea, ObjectManager $manager): Response
+    {
+        $idea->publish();
```
JeremieSamson: Admin can infinitly extends an pending or finalized idea, and should not increment the `$extendsCount`

src/Entity/IdeasWorkshop/Idea.php
```diff
@@ -202,6 +222,7 @@ class Idea implements AuthorInterface, ReportableInterface, EnabledInterface
     use EntityTimestampableTrait;
     use EntityNameSlugTrait;
 
+    public const EXTEND_COUNT_LIMIT = 2;
```
JeremieSamson: public to add detail in the test here https://github.com/EnMarche/en-marche.fr/pull/3446/files#diff-468229cd6b82697c76fb1bdc3a1159a4R21

src/Entity/IdeasWorkshop/Idea.php
```diff
@@ -706,6 +752,34 @@ public function getReportType(): string
         return ReportType::IDEAS_WORKSHOP_IDEA;
     }
 
+    /**
+     * @SymfonySerializer\Groups("idea_read")
+     */
+    public function isExtendable(): bool
+    {
+        return $this->extendsCount < self::EXTEND_COUNT_LIMIT;
```
ottaviano: this will return `true` for a draft idea too, what about adding the status check here?

src/Controller/Admin/AdminIdeaController.php
ottaviano: what about adding the status check here too?

src/Entity/IdeasWorkshop/Idea.php
JeremieSamson: I can add test here ut it will means that I will not be able to know if the 400 is from status or limitation ?

app/migrations/Version20190304170145.php
```diff
@@ -0,0 +1,24 @@
...
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20190304170145 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE 
+          ideas_workshop_idea 
+        ADD 
+          extends_count SMALLINT UNSIGNED NOT NULL,
```
illusionOfParadise: What about `extensions_count` ? 
`extension` is a correct noun.

src/Controller/Api/IdeasWorkshop/IdeaExtendController.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace AppBundle\Controller\Api\IdeasWorkshop;
+
+use AppBundle\Entity\IdeasWorkshop\Idea;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+class IdeaExtendController
```
illusionOfParadise: It will be better to transform `IdeaPublishController` to `IdeaController` and to have one controller per entity than one controller par entity action.

src/Controller/Admin/AdminIdeaController.php
```diff
@@ -106,4 +106,22 @@ public function finalizeAction(Idea $idea, ObjectManager $manager): Response
...
         return $this->redirectToRoute('admin_app_ideasworkshop_idea_list');
     }
+
+    /**
+     * @Route("/{id}/extend", name="app_admin_idea_extend", methods={"GET"})
+     */
+    public function extendAction(Idea $idea, ObjectManager $manager): Response
+    {
+        if ($idea->isDraft()) {
+            $this->addFlash('warning', 'La proposition est un brouillon');
```
ottaviano: redirect is missing after the message?

PR #3452, opened 2019-03-06
Add citizen action synchronization

illusionOfParadise:


features/sync_with_api.feature
```diff
@@ -70,7 +70,7 @@ Feature:
...
+      | routing_key   | body                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
thomasbisignani: Nice catch ! :+1:

thomasbisignani: Nice catch :yum:

PR #3461, opened 2019-03-08
set lowercase the first letter on loginEmail input

Raulnet:
set lowercase the first letter on loginEmail input


templates/security/adherent_login.html.twig
```diff
@@ -52,3 +52,13 @@
         Pas encore adhérent(e) ? <a href="{{ path('app_membership_join') }}">S’inscrire</a>.
     </section>
 {% endblock %}
+
+ {% block javascripts %}
+ <script>
```
ottaviano: what about `autocapitalize` attribute? https://developers.google.com/web/updates/2015/04/autocapitalize

Raulnet: edited

PR #3470, opened 2019-03-13
[Adherent Message] enable referent message

ottaviano:


src/Controller/EnMarche/AdherentMessage/ReferentMessageController.php
```diff
@@ -17,4 +17,9 @@ protected function getMessageType(): string
     {
         return AdherentMessageTypeEnum::REFERENT;
     }
+
+    public function disableInProduction(): void
+    {
+        // Override parent method for activating this feature in production
```
thomasbisignani: You cannot remove the `$this->disableInProduction()` calls instead ?

ottaviano: the call is in the parent class and I need to keep it for Deputy controller

thomasbisignani: Ok I see, thx :+1:

PR #3493, opened 2019-03-19
Generate QRCodes locally

Remg:
QRCodes can't be generated anymore since google charts api is deprecated:
```
Warning: This API is deprecated and is scheduled to be turned off on March 14, 2019. Please use the actively maintained Google Charts API instead.
```

Since using an API to generate a QRCode is a bad practice (sending the googleAuthenticatorSecret outside), let's use [endroid/qr-code-bundle](https://github.com/endroid/qr-code-bundle) to generate our QRCodes locally.

(also fixed this deprecation warning: `Deprecation warning: require.beberlei/DoctrineExtensions is invalid, it should not contain uppercase characters. Please use beberlei/doctrineextensions instead. Make sure you fix this as Composer 2.0 will error.` )


Remg:
I would like to thank @tony-tran for his amazing work and insights regardings these QRCodes.
He da real mvp.

@tony-tran president !

src/Security/QrCodeResponseFactory.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    private function getQrContent(TwoFactorInterface $user): string
+    {
+        return $this->googleAuthenticator->getQRContent($user);
+    }
+
+    private function getQrCode(TwoFactorInterface $user): QrCodeInterface
+    {
+        return $this->qrCodeFactory->create($this->getQrContent($user));
```
tony-tran: WDYT about : 

```return $this->qrCodeFactory->create($this->googleAuthenticator->getQRContent($user));```

PR #3494, opened 2019-03-19
Sync Committee memberships with Mailchimp

ottaviano:


app/config/services/messenger.yml
```diff
@@ -76,3 +76,9 @@ services:
     messenger.receiver_locator:
         tags: [container.service_locator]
         arguments: [[]]
+
```
JeremieSamson: can you remove this line please :) ?

src/Mailchimp/Driver.php
```diff
@@ -88,6 +88,31 @@ public function sendCampaign(string $externalId): bool
         return $this->sendRequest('POST', sprintf('/campaigns/%s/actions/send', $externalId));
     }
 
+    public function createStaticSegment(string $name): array
+    {
+        $response = $this->send('POST', sprintf('/lists/%s/segments', $this->listId), [
```
JeremieSamson: Shouldn't we have same indentation for those 3 functions ?

templates/profiler/messenger.html.twig
```diff
@@ -0,0 +1,187 @@
...
+                <span>{{ collector.exceptionsCount }}</span>
+            </span>
+        {% endif %}
+    </span>
+{% endblock %}
+
+{% block head %}
+    {{ parent() }}
+    <style>
+        .message-item thead th { position: relative; cursor: pointer; user-select: none; padding-right: 35px; }
```
thomasbisignani: Is it possible to move this style in a SCSS file ?

app/config/services/messenger.yml
ottaviano: done, thanks 👍

templates/profiler/messenger.html.twig
ottaviano: I prefer do not change this file, as it comes from the package Symfony/FrameworkBundle:4.1 and will be removed when we will up to sf4

src/Mailchimp/Synchronisation/EventListener/AdherentEventSubscriber.php
```diff
@@ -64,14 +66,31 @@ public function onAfterUpdate(UserEvent $event): void
...
+        $this->dispatchMessage($event->getUser()->getUuid(), $event->getUser()->getEmailAddress());
+    }
+
+    public function onCommitteePrivilegeChange(CommitteeEventInterface $event): void
+    {
+        $adherent = $event->getAdherent();
+
+        $this->dispatchMessage($adherent->getUuid(), $adherent->getEmailAddress());
+
+        if ($committee = $event->getCommittee()) {
```
Remg: what about:
```
if (!$committee = $event->getCommittee()) {
    return;
}
```

to remove an indentation level?
(not important)

tests/Admin/CommitteeAdminTest.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    use ControllerTestTrait;
+    use MessengerTestTrait;
+
+    /**
+     * @var CommitteeRepository
+     */
+    private $committeeRepository;
+
+    public function test committee static segment command is dispatched when committee is approved(): void
```
Remg: `public function dude_what?`

PR #3501, opened 2019-03-22
[AdherentMessage] Add check of subscription type for each message

ottaviano:


src/Form/AdherentMessageReferentZoneFilterType.php
```diff
@@ -11,7 +11,10 @@ class AdherentMessageReferentZoneFilterType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('referentTags', ReferentTagChoiceType::class);
+        $builder->add('referentTag', ReferentTagChoiceType::class, [
+            'multiple' => false,
```
thomasbisignani: This default value is really needed ?

ottaviano: nop, removed, thanks

PR #3521, opened 2019-03-27
Add search only of referent events

illusionOfParadise:


src/Repository/EventRepository.php
```diff
@@ -437,15 +442,21 @@ public function searchAllEvents(SearchParametersFilter $search): array
             $filterCategory = '';
         }
 
+        if ($search->getReferentEvents()) {
```
JeremieSamson: Why not a ternary condition here :thinking: ?

illusionOfParadise: With simple `if` it's more readable.

PR #3523, opened 2019-04-01
Add referent tags to CA and synchronize them with API

illusionOfParadise:


src/Entity/CitizenAction.php
```diff
@@ -5,14 +5,15 @@
...
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\Mapping as ORM;
 use JMS\Serializer\Annotation as JMS;
 use Ramsey\Uuid\UuidInterface;
 
 /**
  * @ORM\Entity(repositoryClass="AppBundle\Repository\CitizenActionRepository")
  */
-class CitizenAction extends BaseEvent
+class CitizenAction extends BaseEvent implements ReferentTaggableEntity
```
ottaviano: no needed as BaseEvent already implements this interface

src/Entity/EntityReferentTagTrait.php
```diff
@@ -3,7 +3,6 @@
 namespace AppBundle\Entity;
 
 use Doctrine\Common\Collections\Collection;
-use Doctrine\ORM\Mapping as ORM;
```
ottaviano: hm why?

src/Admin/CitizenActionAdmin.php
```diff
@@ -29,6 +33,17 @@ class CitizenActionAdmin extends AbstractAdmin
...
 
+    /** @var EventDispatcherInterface */
+    private $eventDispatcher;
+    private $referentTagManager;
+
+    public function __construct($code, $class, $baseControllerName, ReferentTagManager $referentTagManager)
+    {
+        parent::__construct($code, $class, $baseControllerName);
+
+        $this->referentTagManager = $referentTagManager;
```
thomasbisignani: To avoid the override of the construct, what do you think to inject also by setter injection the `referentTagManager` like the `eventDispatcher` ?

src/Entity/EntityReferentTagTrait.php
illusionOfParadise: Why not 😜

PR #3534, opened 2019-04-04
[Procuration] Add fields region, state and reachable on procuration proxy and request

JeremieSamson:


front/services/form/changeStateFieldVisibility.js
```diff
@@ -0,0 +1,7 @@
+export function changeStateFieldVisibility(country, state) {
```
ottaviano: I think we can move here the dom element selection from request and proxy js files:
```js
export default function (countryFieldSelector, stateFieldSelector) {
    const countryElement = dom(countryFieldSelector);
    const stateElement = dom(stateFieldSelector);

    changeStateFieldVisibility(countryElement, stateElement);

    on(countryElement, 'change', () => changeStateFieldVisibility(countryElement, stateElement));
}

function changeStateFieldVisibility(country, state) {
    if ('FR' === country.value) {
        state.classList.add('hidden');
    } else {
        state.classList.remove('hidden');
    }
}
```

and after:
```js
// request.js file
import changeStateFieldVisibility from '../services/form/changeStateFieldVisibility';

changeStateFieldVisibility(countrySelector, stateSelector);
```

PR #3558, opened 2019-04-10
Split symfony monolit package

ottaviano:
vendors diffs, left side = master, right side this branch:
1. 
image
2.
image
3.
image

### DIFF
--- a/src/Entity/AbstractTranslatableEntity.php
+++ b/src/Entity/AbstractTranslatableEntity.php
@@ -2,20 +2,18 @@
 
 namespace AppBundle\Entity;
 
-use A2lix\I18nDoctrineBundle\Doctrine\ORM\Util\Translatable;
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Symfony\Component\Validator\Constraints as Assert;
 
-trait EntityTranslatableTrait
+abstract class AbstractTranslatableEntity
 {
-    use Translatable;
-
     /**
      * @var EntityTranslationInterface[]|Collection
      *
      * @Assert\Valid
      */
-    private $translations;
+    protected $translations;
 
     public function translate(string $locale = 'fr'): ?EntityTranslationInterface
     {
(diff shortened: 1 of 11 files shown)


src/Entity/AbstractTranslatableEntity.php
```diff
@@ -46,4 +44,13 @@ private function removeTranslationIfEmpty(string $locale): void
             $this->removeTranslation($translation);
         }
     }
+
+    /** @return Collection */
+    abstract public function getTranslations();
```
Remg: can't we typehint the return with `Collection` ?

src/Entity/AbstractTranslatableEntity.php
```diff
@@ -46,4 +44,13 @@ private function removeTranslationIfEmpty(string $locale): void
...
+
+    abstract public function setTranslations(ArrayCollection $translations);
```
Remg: `: void`

src/Entity/AbstractTranslatableEntity.php
```diff
@@ -46,4 +44,13 @@ private function removeTranslationIfEmpty(string $locale): void
...
+
+    abstract public function addTranslation($translation);
```
Remg: `: void`

src/Entity/AbstractTranslatableEntity.php
```diff
@@ -46,4 +44,13 @@ private function removeTranslationIfEmpty(string $locale): void
...
+
+    abstract public function removeTranslation($translation);
```
Remg: `: void`

src/Entity/AbstractTranslatableEntity.php
ottaviano: @Remg, I can't do it, because they must respect the vendor Trait method signatures 
```
Compile Error: Declaration of A2lix\I18nDoctrineBundle\Doctrine\ORM\Util\Translatable::getTranslations() must be compatible with AppBundle\Entity
  \AbstractTranslatableEntity::getTranslations(): Doctrine\Common\Collections\Collection
```

PR #3560, opened 2019-04-11
[REFERENT SPACE] Change interests filter to multi select

illusionOfParadise:


templates/message/filter.html.twig
```diff
@@ -10,11 +10,7 @@
     <span class="text--bold text--black">2</span> / 2
 {% endblock %}
 
-{% block message_content %}
-    {% include('message/filter/'~ message.type ~'.html.twig') %}
-
-    <div id="message-actions-block" class="l__wrapper--slim background--lighter b__nudge--top-40 text--center"></div>
```
ottaviano: can you keep this div here, because it's common div for all filter page and it's going with `App.runMessageFilters(...)` javascript code.
you can declare a new block `filter_content` for example

templates/message/filter.html.twig
```diff
@@ -25,7 +21,7 @@
                         '{{ message.uuid|e('js') }}',
```
ottaviano: it's not your code but can you replace `() => {` by `function () {` please?

templates/message/filter/referent.html.twig
```diff
@@ -11,5 +14,24 @@
...
+{% endblock %}
+
+{% block javascripts %}
+    {{ parent() }}
+    <script type="text/javascript" src={{ asset('bundles/sonataadmin/vendor/jquery/dist/jquery.min.js') }}></script>
+    <script type="text/javascript" src={{ asset('select2/select2.min.js') }}></script>
+    <script type="text/javascript" src={{ asset('select2/fr.js') }}></script>
+    <script type="text/javascript">
+        Kernel.onLoad(function() {
+            $('.multi_interests').select2({placeholder: '{{ 'referent.filter.interests'|trans|e('js') }}'});
```
ottaviano: I think we need to check if `multi_interests` exists before call `.select2` function

PR #3562, opened 2019-04-12
[REFERENT SPACE] Show email subscription status in the adherents list and add its filter

illusionOfParadise:


src/Referent/ManagedUsersFilter.php
```diff
@@ -112,6 +114,11 @@ public function handleRequest(Request $request): self
         $this->queryAgeMaximum = $query->getInt(self::PARAMETER_AGE_MAX);
         $this->queryInterests = (array) $query->get(self::PARAMETER_INTEREST, []);
         $this->includeCP = $query->getBoolean(self::PARAMETER_INCLUDE_CP);
+        if ('' !== $query->get(self::PARAMETER_EMAIL_SUBSCRIPTION) && null !== $query->get(self::PARAMETER_EMAIL_SUBSCRIPTION)) {
```
ottaviano: we can replace this IF with 
```php
$this->onlyEmailSubscribers = $query->getBoolean(self::PARAMETER_EMAIL_SUBSCRIPTION, null);
```
no?

illusionOfParadise: No, in this case when the `'Tous'` select value is selected and an empty string `''` is passed as a parameter value, `false` will be set to `onlyEmailSubscribers`, when we need it to be `null`.

PR #3569, opened 2019-04-16
[REFERENT SPACE] Show previous connection date in the referent space

illusionOfParadise:


src/Twig/AdherentRuntime.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace AppBundle\Twig;
+
+use AppBundle\Entity\Adherent;
+use AppBundle\Entity\ReferentSpaceAccessInformation;
+use AppBundle\Repository\ReferentSpaceAccessInformationRepository;
+
+class AdherentRuntime
```
ottaviano: `implements RuntimeExtensionInterface`

src/Twig/AdherentRuntime.php
```diff
@@ -0,0 +1,66 @@
...
+
+        if ($adherent->isHost()) {
+            return $adherent->isFemale() ? 'Co-animatrice 🏅' : 'Co-animateur 🏅';
+        }
+
+        // It means the user is an adherent
+        return $adherent->isFemale() ? 'Adhérente 😍' : 'Adhérent 😍';
+    }
+
+    public function getReferentPreviousConnection(Adherent $adherent, string $format = 'd/m/Y'): string
```
ottaviano: `getReferentLastVisitDate` wdyt?

src/Twig/AdherentRuntime.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        /** @var ReferentSpaceAccessInformation $accessInformation */
+        $accessInformation = $this->accessInformationRepository->findByAdherent($adherent);
```
ottaviano: maybe we can store in a cache (for 30min for example) the result of a query to avoid executing the query when a user navigates:
`function findByAdherent(Adherent $adherent, int $resultTtl = -1)` by default without caching

src/Twig/AdherentRuntime.php
illusionOfParadise: I'd rather change to `getReferentPreviousVisitDate`, as we get `previousDate` field value.

PR #3597, opened 2019-04-23
[PHP CS Fixer] Add ordered imports rule

JeremieSamson:


thomasbisignani:
Very nice rule :ok_hand:

PR #3598, opened 2019-04-23
Remove google plus column

Remg:


src/Committee/CommitteeUpdateCommandHandler.php
```diff
@@ -48,8 +48,7 @@ public function handle(CommitteeCommand $command)
 
         $committee->setSocialNetworks(
             $command->facebookPageUrl,
-            $command->twitterNickname,
-            $command->googlePlusPageUrl
+            $command->twitterNickname
```
thomasbisignani: Can be inlined

PR #3613, opened 2019-04-26
[REFERENT SPACE] Add sending test mail

illusionOfParadise:


src/Mailchimp/Driver.php
```diff
@@ -88,6 +88,14 @@ public function sendCampaign(string $externalId): bool
         return $this->sendRequest('POST', sprintf('/campaigns/%s/actions/send', $externalId));
     }
 
+    public function testCampaign(string $externalId, array $emails): bool
```
ottaviano: what about `sendTestCampaign`?

illusionOfParadise: Renaming done.

PR #3619, opened 2019-04-29
[Assessor] Update wording

JeremieSamson:


src/Assessor/AssessorRequestCommand.php
```diff
@@ -48,6 +48,7 @@ class AssessorRequestCommand
     private $firstName;
 
     /**
+     * @Assert\NotBlank(message="assessor.birth_name.not_blank", groups={"fill_personal_info"})
```
thomasbisignani: The message `Votre nom de naissance est requis.` is generic, can you use instead : `common.birth_name.not_blank` ?

Like this new PR : https://github.com/EnMarche/en-marche.fr/pull/3610/files#diff-8f79de8cc8dbed0031a212b6338a458eR26

thomasbisignani: Also, can you add the assert on the entity ?

JeremieSamson: done, thanks :ok_hand:

tests/Controller/EnMarche/AssessorControllerTest.php
```diff
@@ -34,6 +34,7 @@ public function testAssessorRequest()
                 'gender' => 'male',
                 'firstName' => 'Ernestino',
                 'lastName' => 'Bonsoirini',
+                'birthName' => 'Boujourini',
```
illusionOfParadise: Perhaps `Bonjourini`? 😄

translations/validators.fr.yml
```diff
@@ -23,6 +23,10 @@ common.email.invalid: Ceci n'est pas une adresse e-mail valide.
 common.email.max_length: L'adresse e-mail est trop longue, 255 caractères maximum.
 common.email.repeated: Les adresses email ne correspondent pas.
 
+common.birth_name.note_blank: Votre nom de naissance est requis.
```
illusionOfParadise: What about `Le nom de naissance est requis`?

thomasbisignani:
You should also remove the `nullable` behavior of this field : 

- in the `AssessorRequestCommand` DTO (setter / getter ) 
- In the `AssessorRequestFactory::createFromArray` method
- In the `AssessorRequest` entity (create static method, setter / getter, ORM annotation)

PR #3625, opened 2019-04-30
[REFERENT SPACE] Add managed zone filter

illusionOfParadise:


templates/referent/users_list.html.twig
```diff
@@ -36,6 +36,18 @@
                 <div class="manager__filters__row">
                     <div class="manager__filters__section">
                         <div class="manager__filters__group">
+                            {% dump  managedArea.tags.count %}
```
ottaviano: :)

app/migrations/Version20190430130242.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20190430130242 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE referent_managed_users_message ADD managed_area LONGTEXT NOT NULL');
```
ottaviano: should be `default null`

src/Referent/ManagedUsersFilter.php
```diff
@@ -59,6 +60,7 @@ class ManagedUsersFilter
      * @Assert\NotNull
      */
     protected $offset = 0;
+    protected $managedArea = '';
```
ottaviano: what about prefixing by `query` as other query parameters? 
maybe just `queryZone`? (same remark for Entity parameter)

src/Entity/ReferentManagedUsersMessage.php
```diff
@@ -35,6 +35,11 @@ class ReferentManagedUsersMessage extends ManagedUsersMessage
      */
     private $includeSupervisors;
 
+    /**
+     * @ORM\Column(type="text")
```
ottaviano: why not just `string` as type if it's only for persisting the tag's code?

templates/referent/users_list.html.twig
```diff
@@ -36,6 +36,18 @@
...
+                            {% if managedArea.tags.count > 1 %}
```
ottaviano: what about 
```twig 
{% set tags = app.user.managedArea.tags %}
{% if tags.count > 1 %}
...
```

illusionOfParadise: Not possible, because this page is also accessible by a co-referent and `app.user.managedArea.tags` can not be used.

PR #3628, opened 2019-04-30
fix pagination - array sort are commented

Charlesdoiron:


front/components/DataGrid.js
```diff
@@ -197,6 +207,33 @@ export default class DataGrid extends React.Component {
         return pagesList;
     }
 
+    // Work in Progress
```
Remg: we shouldnt merge commantated code, and we shouldnt merge `@wip` or `@todo`  tags

either we put it in another branch and work on it, either we remove this code

Charlesdoiron: 💅

PR #3631, opened 2019-05-02
[REFERENT SPACE] Update style #5

enzozccr:


templates/citizen_project/dashboard.html.twig
```diff
@@ -50,6 +50,9 @@
             {% if citizen_project.isApproved() %}
                 <span class="b__nudge--left-small">{{ citizen_project.membersCount }} membres</span>
             {% endif %}
+            {% if not citizen_project.isApproved() %}
```
ottaviano: you can do this with `if else`
```twig
{% if citizen_project.isApproved() %}
  <span class="b__nudge--left-small">{{ citizen_project.membersCount }} membres</span>
{% else %}
  <span class="text--small approval__status approval__status--pending">En attente de validation</span>
{% endif %}
```

PR #3634, opened 2019-05-03
Import assessors command

Remg:


src/Command/ImportAssessorsCommand.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        $this->em = $em;
+        $this->repository = $repository;
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->addArgument('file', InputArgument::REQUIRED, 'CSV file of all districts to load');
```
ottaviano: `districts` -> `assessors`

src/Command/ImportAssessorsCommand.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        $this->io = new SymfonyStyle($input, $output);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->io->text('Start importing assessors');
+
+        $assessors = $this->parseCSV($input->getArgument('file'));
+        $progress = new ProgressBar($output, \count($assessors));
```
ottaviano: you could use `$this->io->progress*()` methodes

src/Command/ImportAssessorsCommand.php
Remg: copy pasted with the ass.

PR #3643, opened 2019-05-06
[REFERENT SPACE] Update style #6

enzozccr:
https://app.clubhouse.io/enmarche/story/8039/espace-r%C3%A9f%C3%A9rent-am%C3%A9liorations-int%C3%A9gration

### DIFF
--- a/templates/message/list.html.twig
+++ b/templates/message/list.html.twig
@@ -95,7 +95,10 @@
                     </td>
                 </tr>
             {% else %}
-                <tr><td colspan="6" class="text--center">Vous n'avez aucun message</td></tr>
+                <tr><td colspan="6" class="text--center">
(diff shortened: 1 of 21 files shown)


templates/message/list.html.twig
```diff
@@ -95,7 +95,10 @@
                     </td>
                 </tr>
             {% else %}
-                <tr><td colspan="6" class="text--center">Vous n'avez aucun message</td></tr>
+                <tr><td colspan="6" class="text--center">
+                    <img src="/images/icons/icn_no-result.svg" class="icn--no-result" width="30" />
```
ottaviano: `.. src="{{ asset('/images/icons/icn_no-result.svg') }}" ..`

PR #3648, opened 2019-05-07
Improve referent person link edition process

ottaviano:


src/Form/ReferentPersonLinkType.php
```diff
@@ -98,4 +58,27 @@ public function configureOptions(OptionsResolver $resolver)
             'data_class' => ReferentPersonLink::class,
         ]);
     }
+
+    public function transform($value)
```
thomasbisignani: Instead of using a `FormType` who implements `DataTransformerInterface`, IMO, it will be logic to : 
- Create a new data transformer (but IMO this data transformer is too specific to re-use it)
- Or use directly a `CallbackTransformer`

ottaviano: ok, done with CallbackTransformer

PR #3675, opened 2019-05-14
[Mailchimp] add webhook controller

ottaviano:


app/config/parameters.yml.dist
```diff
@@ -75,6 +75,7 @@ parameters:
     env(APP_EMAIL_SUBSCRIPTIONS_WEBHOOK_SECRET): 'mailchimp_secret'
 
     env(MAILCHIMP_API_KEY): ''
+    env(MAILCHIMP_WEBHOOK_KEY): 'abc'
```
thomasbisignani: Not a big deal and not a request change, but WDYT to use `webhook_key` for the default dist parameter ?

PR #3696, opened 2019-05-24
[Jecoute] Added transactional email on survey answered

thomasbisignani:


src/Jecoute/SurveyEvents.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace AppBundle\Jecoute;
+
+final class SurveyEvents
+{
+    public const DATA_SURVEY_ANSWERED = 'data_survey.answered';
+
+    public function __construct()
```
ottaviano: maybe remove it, no?

src/Repository/Jecoute/DataSurveyRepository.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        return $this
+            ->createQueryBuilder('dataSurvey')
+            ->select('COUNT(dataSurvey.id)')
+            ->andWhere('dataSurvey.emailAddress = :email')
+            ->andWhere('dataSurvey.postedAt >= :from')
+            ->andWhere('dataSurvey.postedAt < :until')
+            ->setParameter('email', $email)
+            ->setParameter('from', $date->modify('-1 month'))
+            ->setParameter('until', $date)
```
ottaviano: if `$date` is not immutable date here you will have the same date as for `:from` parameter. I think you should clone the date, wdyt?

PR #3699, opened 2019-05-27
Feature/#8264 admin technical skill

ktherage:


app/config/services/admin.xml
```diff
@@ -976,5 +976,20 @@
...
             <tag name="kernel.event_listener" event="sonata.admin.event.configure.menu.sidebar" method="addMenuItems" />
         </service>
+
+        <!--
+        ######################################################
+        # Municipales
+        ######################################################
+         -->
+
+        <service id="app.admin.applicationrequest_technicalskill" class="AppBundle\Admin\TechnicalSkillAdmin">
```
ottaviano: what about just `app.admin.technicalskill`?

src/Admin/TechnicalSkillAdmin.php
```diff
@@ -0,0 +1,51 @@
...
+            ->add('_action', null, [
+                'actions' => [
+                    'edit' => [],
+                    'delete' => [],
+                ],
+            ])
+        ;
+    }
+
+    protected function configureShowFields(ShowMapper $showMapper)
```
ottaviano: you should avoid overriding this method as you don't have `show` action

src/Admin/ThemeAdmin.php
```diff
@@ -0,0 +1,51 @@
...
+            ->add('_action', null, [
+                'actions' => [
+                    'edit' => [],
+                    'delete' => [],
+                ],
+            ])
+        ;
+    }
+
+    protected function configureShowFields(ShowMapper $showMapper)
```
ottaviano: same for this method

PR #3715, opened 2019-06-04
[Mailchimp] Add command for updating adherent data

ottaviano:


src/Mailchimp/Webhook/Handler/UnsubscribeHandler.php
```diff
@@ -17,6 +17,7 @@ public function __construct(SubscriptionHandler $handler)
     public function handle(array $data): void
     {
         if ($adherent = $this->getAdherent($data['email'])) {
+            dump($adherent);
```
ktherage: Forgot dump :smile:

PR #3720, opened 2019-06-06
Adding new space for Municipal Chief

ktherage:


src/Intl/FranceCitiesBundle.php
```diff
@@ -84,6 +104,36 @@ public static function searchCities(string $search): array
...
+    }
+
+    /**
+     * Returns the a list of city for a given list of INSEE code.
+     *
+     * @return array|null
+     */
+    public static function searchCitiesByInseeCodes(array $inseeCodes)
+    {
+        $citiesByInsee = [];
```
ottaviano:
```php
return array_intersect_key(
  self::getCityByInseeCode(), 
  array_fill_keys($inseeCodes, 1)
);
```

templates/municipal_leader/_layout.html.twig
```diff
@@ -0,0 +1,51 @@
...
+                            <a href="https://t.me/joinchat/EmY0e1J2fyTv4Fd-cHEMHg" target="_blank" class="text--blue--dark link--no-decor">Écrivez-nous !</a>
+                        </p>
+                    </div>
+                    <div class="second-section">
+                        <h1 class="page-title text--large b__nudge--bottom">
+                            {{ block('page_title') }}
+                        </h1>
+                    </div>
+
+                    {% if is_granted('ROLE_MUNICIPAL_LEADER') %}
```
ottaviano: do we really need to do this check in this template?

src/Admin/AdherentAdmin.php
```diff
@@ -580,6 +589,12 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                         $qb->setParameter('cp_privilege', CitizenProjectMembership::CITIZEN_PROJECT_ADMINISTRATOR);
                     }
 
+                    // municipal chief
```
thomasbisignani: IMO this comment is useless, the `if` statement is already explicit

src/Controller/EnMarche/MunicipalChiefController.php
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * @Route("/", name="app_municipalchief_index")
+     */
+    public function indexAction(UserInterface $municipalChief): Response
+    {
+        $municipalChiefManagedArea = $municipalChief->getMunicipalChiefManagedArea();
+
+        return $this->render('municipal_chief/index.html.twig', [
+            'managedCities' => FranceCitiesBundle::searchCitiesByInseeCodes(
+                $municipalChiefManagedArea->getCodes()
```
thomasbisignani: You can write directly : 
```php 
$municipalChief->getMunicipalChiefManagedArea()->getCodes()
```

src/Entity/MunicipalChiefManagedArea.php
```diff
@@ -0,0 +1,25 @@
...
+ * @ORM\Entity
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class MunicipalChiefManagedArea extends ManagedArea
+{
+    public function setCodes(array $codes): void
+    {
+        $codes = array_filter($codes, function ($value) {
+            return \in_array($value, array_keys(FranceCitiesBundle::getCityByInseeCode()));
```
ottaviano: I think we should do it with form validation process and not here

src/Adherent/AdherentRoleEnum.php
```diff
@@ -29,4 +29,6 @@ class AdherentRoleEnum extends Enum
     public const LAREM = 'gq_ideas';
 
     public const ELECTED = 'elected';
+
+    public const MUNICIPAL_CHIEF = 'Chef municipal';
```
Remg: this should be a value in snake_case like the others

src/Intl/FranceCitiesBundle.php
```diff
@@ -36,6 +36,24 @@ public static function getCity(string $postalCode, string $inseeCode)
         return self::$cities[$postalCode][$inseeCode] ?? null;
     }
 
+    /**
+     * Returns the cities indexed by INSEE code.
+     */
+    public static function getCityByInseeCode(): array
+    {
+        static $citiesByInsee = [];
```
Remg: why `static` ?

src/Intl/FranceCitiesBundle.php
```diff
@@ -36,6 +36,24 @@ public static function getCity(string $postalCode, string $inseeCode)
...
+
+        if (empty($citiesByInsee)) {
```
Remg: i don't understand this check. this array will always be empty here ?

src/Intl/FranceCitiesBundle.php
ktherage: this is to keep the values between multiple calls. this works like a class property scoped to the function. see : https://www.php.net/manual/fr/language.variables.scope.php#language.variables.scope.static

src/Intl/FranceCitiesBundle.php
ktherage: at the first function call `$citiesByInsee` is empty and so it is processed. But on the second (or more) function it is not processed anymore _Home made cache_ :wink:.

src/Adherent/AdherentRoleEnum.php
ktherage: Wouldn't it disturb admin users to see snake_case string in a label?

PR #3721, opened 2019-06-07
[Admin] Add board member form to adherent page

ottaviano:
[ch8527]

### DIFF
--- a/tests/Admin/AdherentAdminTest.php
+++ b/tests/Admin/AdherentAdminTest.php
@@ -6,6 +6,7 @@ use AppBundle\DataFixtures\ORM\LoadAdherentData;
 use AppBundle\Entity\Adherent;
 use Liip\FunctionalTestBundle\Test\WebTestCase;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


tests/Admin/AdherentAdminTest.php
```diff
@@ -35,6 +36,78 @@ public function testAnAdminCantBanAnAdherent()
         $this->assertResponseStatusCode(404, $this->client->getResponse());
     }
 
+    public function testEditBoardMemberInformations()
+    {
+        $adherent = $this->adherentRepository->findOneByUuid(LoadAdherentData::ADHERENT_2_UUID);
+        $this->assertTrue($adherent->isBoardMember());
+        $this->authenticateAsAdmin($this->client);
+        $editUrl = sprintf('/admin/app/adherent/%s/edit', $adherent->getId());
```
ktherage: The URI is used at least twice, i would place her in a class constante. WDYT?

ottaviano: why not, done

PR #3731, opened 2019-06-11
update training homepage & add faq page

nicoweb:
Capture d’écran 2019-06-11 à 14 52 58


templates/formation/_navbar.html.twig
```diff
@@ -0,0 +1,20 @@
+<nav>
+    <ul class="list__links list__links--row list__links--no-decor list__links--nav">
+        {% set currentRoute = app.request.attributes.get('_route') %}
+        <li>
+            <a href="{{ path('app_formation_home') }}" class="{{ 'app_formation_home' == currentRoute ? 'active' }}">
```
ottaviano: you could use a twig helper 
```
... class="{{ is_active_route(app.request, 'app_referent_events*') ? 'active' : '' }}">
```

tests/Controller/EnMarche/FormationControllerTest.php
```diff
@@ -0,0 +1,75 @@
...
+namespace Tests\AppBundle\Controller\EnMarche;
+
+use Liip\FunctionalTestBundle\Test\WebTestCase;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Tests\AppBundle\Controller\ControllerTestTrait;
+
+/**
+ * @group functional
+ * @group formation
```
thomasbisignani: We can remove this `@group`, now we don't use this annotation to split the job. 
We use only the `@group functional` to run the `make tf` target

nicoweb: done

PR #3737, opened 2019-06-12
Feature/ch8556 : upload files (pdfs) in training section

nicoweb:


src/Entity/Formation/File.php
```diff
@@ -0,0 +1,58 @@
...
+class File extends BaseFile
+{
+    const PREFIX_PATH = 'files/formations';
+
+    /**
+     * @var UploadedFile|null
+     *
+     * @Assert\File(
+     *     maxSize="5M",
+     *     mimeTypes={"application/pdf"}
```
Remg: Can you please also allow the mimeType `application/x-pdf` ?

src/Entity/Formation/Article.php
```diff
@@ -79,6 +81,19 @@ class Article implements EntityMediaInterface
      */
     private $axe;
 
+    /**
+     * @var Collection|File[]
+     *
+     * @ORM\OneToMany(targetEntity="AppBundle\Entity\Formation\File", cascade={"persist"}, mappedBy="article", orphanRemoval=true)
```
Remg: `orphanRemoval=true` => The file will be deleted if it is unlinked from this Article.

You also need to set a DELETE cascade => The file will be deleted is the Article is deleted.
( `cascade={"persist", "remove"}` or `cascade={"all"}` might be great here )

src/Entity/Formation/Article.php
```diff
@@ -133,4 +148,40 @@ public function setAxe(Axe $axe): void
...
+
+    /**
+     * @return Collection|File[]
+     */
+    public function getFiles(): Collection
+    {
+        return $this->files;
+    }
+
+    public function setFiles(array $files): self
```
Remg: i still find weird that this method is needed.

src/Entity/Formation/Article.php
```diff
@@ -133,4 +148,40 @@ public function setAxe(Axe $axe): void
...
+    {
+        if (\count($files) > 0) {
+            foreach ($files as $file) {
+                $this->addFile($file);
+            }
+        }
+
+        return $this;
```
Remg: We do not use fluent pattern on this project.

Please remove all of these `return $this;` from your setters. ( setters return types should be `: void` )

src/Entity/Formation/Article.php
Remg: i confirm it's ok for me if i remove this method

PR #3741, opened 2019-06-13
Ch8508 application request tags

ktherage:


templates/admin/application_request/show_favorite_cities.html.twig
```diff
@@ -0,0 +1,3 @@
+<td>
```
ottaviano: could you reuse Sonata template for the field value:
```twig

{% block field %}
    {{ object.getFavoriteCitiesNames()|join(', ') }}
{% endblock %}
```

templates/admin/application_request/show_favorite_cities.html.twig
```diff
@@ -1,5 +1,5 @@
 {% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
 
 {% block field %}
-    {{ object.favoriteCities|map(code => get_city_name_from_insee_code(code))|join(', ') }}
+    {{ object.getFavoriteCitiesNames()|join(', ') }}
```
Remg: It would be nice for admin to display cities like this:
```
{{ city.name }} ({{ city.postal_code }})
```
it can be done with [this](https://github.com/EnMarche/en-marche.fr/blob/master/src/Twig/AppIntlExtension.php#L20) method

PR #3744, opened 2019-06-13
Application request - fixes 2.0

Remg:


src/Entity/ApplicationRequest/ApplicationRequest.php
```diff
@@ -30,12 +30,13 @@ abstract class ApplicationRequest implements ReferentTaggableEntity
      *
      * @ORM\Column
      *
-     * @Assert\NotBlank(message="application_request.first_name.not_blank")
+     * @Assert\NotBlank(message="application_request.first_name.not_blank", groups={"application_request"})
```
ottaviano: I think you should use `Default` group too for keeping these constraints on Admin form validation (or just keep only one group for this entity - Default?)

src/Entity/ApplicationRequest/Theme.php
```diff
@@ -56,4 +63,14 @@ public function setName(string $name): void
     {
         $this->name = $name;
     }
+
+    public function getDisplay(): bool
```
nicoweb: maybe `isDisplay`?

PR #3749, opened 2019-06-14
Application requests confirmation email

Remg:


src/Entity/ApplicationRequest/ApplicationRequest.php
```diff
@@ -199,6 +199,11 @@ public function setLastName(?string $lastName): void
         $this->lastName = $lastName;
     }
 
+    public function getFullName(): string
+    {
+        return "$firstName $lastName";
```
nicoweb: maybe:
```
return "$this->firstName $this->lastName";
```

PR #3751, opened 2019-06-14
[ApplicationRequest] Some improvements & Twig update

ottaviano:


src/Entity/Adherent.php
```diff
@@ -1102,8 +1102,7 @@ public function setReferent(array $tags, string $markerLatitude = null, string $
 
     public function isReferent(): bool
     {
-        return $this->managedArea instanceof ReferentManagedArea
-            && !$this->managedArea->getTags()->isEmpty();
+        return $this->managedArea instanceof ReferentManagedArea;
```
Remg: i remember, a while ago, this extra check missing made the application send an email to the whole database :)
are we sure we want to remove this check ?
is it for performance?
what if we forget to apply this eventlistener on a form one day?

templates/admin/application_request/_favorite_cities.html.twig
```diff
@@ -1,10 +1,5 @@
 {% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
 
 {% block field %}
-    {% if object.favoriteCities %}
-        {% for city in object.favoriteCities %}
-            {{ get_city_name_from_insee_code(city) }} ({{ city }})
-            {% if not loop.last %}, {% endif %}
-        {% endfor %}
-    {% endif %}
+    {{ object.favoriteCities|map(code => get_city_name_from_insee_code(code))|join(', ') }}
```
Remg: 👍

src/Entity/Adherent.php
Remg: same modification/questions would apply to : 
https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/Adherent.php#L1149
https://github.com/EnMarche/en-marche.fr/blob/master/src/Entity/Adherent.php#L1151

etc etc

i mostly see these checks as a security if we ever have a corrupted entry in the database.

ottaviano: agree, I prefer to rollback this part

PR #3763, opened 2019-06-18
Add paths in training section

nicoweb:


src/Admin/Formation/PathAdmin.php
```diff
@@ -0,0 +1,57 @@
...
+                    'label' => 'description',
+                ])
+            ->end()
+        ;
+    }
+
+    protected function configureListFields(ListMapper $listMapper)
+    {
+        $listMapper
+            ->addIdentifier('title', TextType::class, [
```
ottaviano: you should not use `*Type::class` of form component for configuring list page.
you could keep it `null` and sonata will guess the type of displayed value or string `array`, '...'

src/Entity/Formation/Axe.php
```diff
@@ -77,20 +77,30 @@ class Axe implements EntityMediaInterface
...
     public function __construct()
     {
-        $this->articles = new ArrayCollection();
+        $this->modules = new ArrayCollection();
     }
 
     public function __toString()
     {
-        return (string) $this->title;
+        return (string) $this->path->getTitle().', '.$this->title;
```
ottaviano: we can remove cast here `(string)` as the concatenation operation returns always a string:
`1 . ' hey ' . 2 => string`

src/Repository/Formation/AxeRepository.php
```diff
@@ -16,11 +16,11 @@ public function __construct(RegistryInterface $registry)
...
      * @return Axe[]
      */
-    public function findAllWithArticles(): array
+    public function findAllWithModules(): array
     {
         return $this->createQueryBuilder('a')
-            ->addSelect('articles')
-            ->leftJoin('a.articles', 'articles')
+            ->addSelect('modules')
+            ->leftJoin('a.modules', 'modules')
```
ottaviano: if the method name is `...WithModules` it should be `innerJoin` here, no?

src/Repository/Formation/PathRepository.php
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * @return Path[]
+     */
+    public function findAllWithAxesAndModules(): array
+    {
+        return $this->createQueryBuilder('p')
+            ->addSelect('axes', 'modules')
+            ->leftJoin('p.axes', 'axes')
```
ottaviano: same remark here for `innerJoin`

templates/formation/home.html.twig
```diff
@@ -23,18 +23,22 @@
...
+        {% for path in paths %}
+            <h2>{{ path.title }}</h2>
+            <p>{{ path.description }}</p>
+            <ol>
+                {% for keyAxe, axe in path.axes %}
+                <li>
+                    <h2 class="text--medium-small" style="padding:80px 0 0 30px;">{{ axe.title }}</h2>
+                </li>
+                    {% for keyModule, module in axe.modules %}
+                        <a href="{{ path('app_formation_module', { 'slug': module.slug }) }}">
```
ottaviano: it seems that we can have this html structure: 
```html
<ol>
  <li>...</li>
  <a>...</a>
  <a>...</a>
  <a>...</a>
  <li>...</li>
</ol>
```
so it's not very correct as an OL element can have only LI element as children

src/Repository/Formation/AxeRepository.php
nicoweb: yes, could be better to not show empty `axes` and empty `paths` 👍

PR #3771, opened 2019-06-19
Change cookies's bar style

enzozccr:
https://app.clubhouse.io/enmarche/story/8699/int%C3%A9gration-de-la-banni%C3%A8re-des-cookies

### DIFF
--- a/features/surveys.feature
+++ b/features/surveys.feature
@@ -18,8 +18,8 @@ Feature:
 
     # Create
     Given I am on "/espace-responsable-jecoute/questionnaire/creer"
-    Then I press "OK"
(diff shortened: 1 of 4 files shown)


features/surveys.feature
```diff
@@ -18,11 +18,12 @@ Feature:
 
     # Create
     Given I am on "/espace-responsable-jecoute/questionnaire/creer"
-    When I wait 10 seconds until I see "Retour à la liste des questionnaires locaux"
     Then I press "OK"
+    When I wait 10 seconds until I see "Retour à la liste des questionnaires locaux"
```
Remg: It should be "When" before "Then"

PR #3783, opened 2019-06-21
[ApplicationRequest] add gender field

ottaviano:
[ch8752]

### DIFF
--- a/src/DataFixtures/ORM/LoadApplicationRequestRunningMateRequestData.php
+++ b/src/DataFixtures/ORM/LoadApplicationRequestRunningMateRequestData.php
@@ -28,6 +28,7 @@ class LoadApplicationRequestRunningMateRequestData extends Fixture
             $municipal3->municipalChiefManagedArea()->getCodes()[1],
         ]);
         $runningMateRequest2->setEmailAddress('damien.schmidt@example.ch');
+        $runningMateRequest2->setAdherent($this->getReference('adherent-14'));
 
         $runningMateRequest3 = new RunningMateRequest();
         $runningMateRequest3->setFavoriteCities([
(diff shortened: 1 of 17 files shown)


src/DataFixtures/ORM/LoadApplicationRequestRunningMateRequestData.php
```diff
@@ -41,6 +42,7 @@ public function load(ObjectManager $manager)
             $municipal1->municipalChiefManagedArea()->getCodes()[2],
         ]);
         $runningMateRequest4->setEmailAddress('damien.schmidt@example.ch');
+        $runningMateRequest2->setAdherent($this->getReference('adherent-14'));
```
ktherage: Is that what you planed to do when using `runningMateRequest2` instead of `runningMateRequest4`?

templates/assessor_manager/_layout.html.twig
```diff
@@ -59,7 +59,7 @@
 
                 <ul class="assessor-manager__tabs">
                     {% for page in pages %}
-                        <li class="{{ app.request.attributes.get('_route') == page.route and filters.status == page.status ? 'active' : '' }}">
+                        <li class="{{ is_active_route(app.request, page.route) and filters.status == page.status ? 'active' }}">
```
ktherage: Just asking, the `: ''` is not a mandatory ?

src/DataFixtures/ORM/LoadApplicationRequestRunningMateRequestData.php
ottaviano: nope, fixed 👍 👀

templates/assessor_manager/_layout.html.twig
ottaviano: yes, it's optional, I discovered this yesterday 😄

PR #3785, opened 2019-06-21
[TIMELINE] Introduce Manifesto entity (& translatable)

Remg:


src/Entity/Timeline/Manifesto.php
```diff
@@ -0,0 +1,130 @@
...
+        return $this->id;
+    }
+
+    /**
+     * @Algolia\Attribute
+     */
+    public function titles(): array
+    {
+        /** @var ProfileTranslation $french */
+        if (!$french = $this->translate('fr')) {
```
nicoweb: maybe you can put this code in `__construct()` method and create properties `$french` and `$english` in order to no repeat it in `slugs()` method

Remg: made a generic method in the abstract class :+1:

src/Admin/TimelineProfileAdmin.php
```diff
@@ -53,8 +54,10 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
     protected function configureListFields(ListMapper $listMapper)
     {
         $listMapper
-            ->addIdentifier('title', null, [
+            ->addIdentifier('title', TextType::class, [
```
ottaviano: not need this change for ListAction

src/Entity/Timeline/Measure.php
```diff
@@ -312,22 +332,17 @@ public function profileIds(): array
     /**
      * @Algolia\Attribute
      */
-    public function titles(): array
+    public function manifestoId(): ?int
```
ottaviano: maybe `getManifestoId()` with `@Algolia\Attribute(algoliaName="...")` if you want change the attribute name?

src/Entity/Timeline/Theme.php
```diff
@@ -164,64 +164,45 @@ public function profileIds(): array
     /**
      * @Algolia\Attribute
      */
-    public function titles(): array
+    public function manifestoIds(): array
```
ottaviano: same for the method name

src/Entity/Timeline/Theme.php
```diff
@@ -164,64 +164,45 @@ public function profileIds(): array
...
-        /** @var ThemeTranslation $french */
-        if (!$french = $this->translate('fr')) {
-            return [];
-        }
+        $manifestos = new ArrayCollection();
 
-        /** @var ThemeTranslation $english */
-        if (!$english = $this->translate('en')) {
-            $english = $french;
+        foreach ($this->measures as $measure) {
```
ottaviano: what about: 
```php

return array_unique(
    array_map(
        function (Measure $m) { return $m->getManifesto()->getId();}, 
        $measures
    )
);
```

src/Admin/TimelineManifestoAdmin.php
```diff
@@ -0,0 +1,81 @@
...
+                'label' => 'Description',
+                'show_filter' => true,
+            ])
+        ;
+    }
+
+    protected function configureListFields(ListMapper $listMapper)
+    {
+        $listMapper
+            ->addIdentifier('title', TextType::class, [
```
ottaviano: `TextType` => `null`

src/Admin/TimelineManifestoAdmin.php
```diff
@@ -0,0 +1,81 @@
...
+                'label' => 'Titre',
+                'virtual_field' => true,
+                'template' => 'admin/timeline/theme/list_title.html.twig',
```
ottaviano: could you try to use `translate.title` as the path of property to display and avoid using the custom template?

Remg: doesnt work as it shows up the english translations

src/Entity/Timeline/Manifesto.php
```diff
@@ -0,0 +1,101 @@
...
+     * @Algolia\Attribute(algoliaName="descriptions")
+     */
+    public function getDescriptions(): array
+    {
+        return $this->getFieldTranslations('description');
+    }
+
+    public function exportTitles(): string
+    {
+        return join(', ', $this->getTitles());
```
ktherage: What about using `implode` directly as `join` is his alias (https://www.php.net/manual/en/function.join.php) ?

PR #3788, opened 2019-06-21
refactor subnav

nicoweb:


templates/components/_nav_desktop.html.twig
```diff
@@ -208,71 +208,47 @@
     </div>
 </nav>
 {% if is_granted('ROLE_SUPERVISOR') or is_granted('ROLE_HOST') or is_granted('ROLE_CITIZEN_PROJECT_ADMINISTRATOR') or is_granted('ROLE_REFERENT') or is_granted('ROLE_COREFERENT') or is_granted('ROLE_DEPUTY') or is_granted('ROLE_BOARD_MEMBER') or is_granted('ROLE_COORDINATOR') or is_granted('ROLE_PROCURATION_MANAGER') or is_granted('ROLE_LEGISLATIVE_CANDIDATE') or is_granted('ROLE_ASSESSOR_MANAGER') or is_granted('ROLE_MUNICIPAL_CHIEF') or is_granted('ROLE_JECOUTE_MANAGER') %}
```
Remg: Because of this line : 
https://github.com/EnMarche/en-marche.fr/pull/3788/files#diff-f79be8b0cbb80997908729868e744321R42

the button "Espace chef municipal" will only appear if :
- user is a municipal chief
- canary is enabled (globally or for this user)

so, if the user is ONLY a municipal chief, the subnav should only appear if canary is indeed enabled.

So in this line it should be : 
```
- or is_granted('ROLE_MUNICIPAL_CHIEF')
+ or (is_granted('ROLE_MUNICIPAL_CHIEF') and is_canary_enabled())
```

PR #3789, opened 2019-06-21
Adjust autocomplete style

enzozccr:


templates/javascript.js.twig
```diff
@@ -304,8 +307,11 @@ $(document).ready(function() {
             }
 
             return false;
+        },
+        open: function() {
+            $("ul.ui-menu").width($(this).innerWidth());
         }
-    });
+    }).autocomplete('widget').addClass('em-form__autocomplete');;
```
ottaviano: CS: double `;`

PR #3799, opened 2019-06-26
fix: disable administrator

nicoweb:


migrations/Version20190626102815.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20190626102815 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE administrators ADD activated TINYINT(1) DEFAULT NULL');
+        $this->addSql('UPDATE administrators SET activated = 1 WHERE activated IS NULL');
```
ottaviano: it can be without WHERE statement as it's a freshly added column ;)

src/Admin/AdministratorFactory.php
```diff
@@ -20,6 +20,9 @@ public function createFromArray(array $data): Administrator
         $admin->setEmailAddress($data['email']);
         $admin->setPassword($this->encodePassword($data['password']));
         $admin->setGoogleAuthenticatorSecret($data['secret'] ?? null);
+        if (isset($data['activated'])) {
```
ottaviano: as an alternative: 
```php
$admin->setActivated(!empty($data['activated']));
```

PR #3801, opened 2019-06-26
[ApplicationRequest] improve datagrid

ottaviano:
image

PR #3802, opened 2019-06-26
Wrapping tags into span

ktherage:


ottaviano:
could you close this PR, please? as discussed I do it [here](https://github.com/EnMarche/en-marche.fr/pull/3801/files#diff-aebab153b9580a34544eaa87136375dfR7)
```html
<td class="municipal-candidate-tags">
  <span>tag 1</span>
  <span>tag 2</span>
</td>
```

PR #3803, opened 2019-06-26
[AL Journey] Add "Tous politiques!" section

enzozccr:


src/Entity/Formation/File.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @var Module
+     *
+     * @ORM\ManyToOne(targetEntity="Module", inversedBy="files")
+     * @ORM\OrderBy({"id": "ASC"})
+     */
+    private $module;
+
+    public function getPrefixPath(): string
```
ktherage: Is that necessary somewhere? IMO, this can be remove in favor of a constant call.

PR #3807, opened 2019-06-27
Update committee section

nicoweb:


templates/committee_manager/edit.html.twig
```diff
@@ -14,6 +14,16 @@
 {% block committee_content %}
     <main>
         <section class="l__wrapper--slim">
+            {% if committee.isPending %}
```
ottaviano: `if not approved` will be more correct as a committee can be in `pending` or in `pre-pending` status.

src/Controller/EnMarche/CommitteeController.php
```diff
@@ -28,14 +28,20 @@ class CommitteeController extends Controller
...
     {
         if ($this->isGranted('IS_ANONYMOUS')
             && $authenticate = $this->get(AnonymousFollowerSession::class)->start($request)
         ) {
             return $authenticate;
         }
 
+        if ($committee->isPending()) {
+            return $this->redirectToRoute('app_committee_manager_edit', [
+                'slug' => $slug,
```
ottaviano: or just `'slug' => $committee->getSlug()`?

PR #3813, opened 2019-06-28
Increase readability in PHPUnit classes

nicoweb:


thomasbisignani:
NOTE : 

The return type `: void` for `setUp` and `tearDown` are useless, the signatures don't have return types.

nicoweb:
The return type for `setUp` and `tearDown` is for compatibility with phpUnit 8

ottaviano:
already done (10 years later xD), but thank you @nicoweb

PR #3815, opened 2019-07-01
[ADH Space] Hotfixes

enzozccr:


front/style/manager-space/_tables.scss
```diff
@@ -292,3 +292,17 @@
         }
     }
 }
+
+td.municipal-candidate-tags {
```
ottaviano: `tbody tr td.municipal-candidate-tags span` will not be enough as the selector for avoiding add a new class to each span?

enzozccr: hmm, yes it might be enough !

enzozccr: up, https://github.com/EnMarche/en-marche.fr/pull/3815/commits/9f771eb1e58a98169fd4479fd100b6fa1c872fc6

ottaviano: nice, thanks! already approved ;)

PR #3831, opened 2019-07-04
[ADMIN] keep track of exports

Remg:


src/EventListener/AdministratorExportHistoryListener.php
```diff
@@ -0,0 +1,51 @@
...
+        $administrator = $this->security->getUser();
+
+        if (!$administrator instanceof Administrator) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        $routeName = $request->get('_route');
+
+        if (!preg_match('/^admin_(.)+_export$/', $routeName)) {
```
ottaviano: what about checking 200 response status code too?

Remg: why ?
i want to know if someone tried to export things.
i dont care if it succeeded or not?

Remg: yes.
the reporting is about the export requests. not the export responses

PR #3896, opened 2019-07-25
Refactor storage adapters

Remg:
1. cleans the local structure:
Screenshot from 2019-07-25 05-46-22

it is now clear this directory is holding the local storage, and this local storage is no more poluted by other internal files.

[ch9358]


thomasbisignani:
:1st_place_medal:

PR #3930, opened 2019-08-27
[Jecoute] Admin : Added surveys tags

thomasbisignani:


src/Admin/JecouteLocalSurveyAdmin.php
```diff
@@ -127,7 +134,7 @@ protected function configureListFields(ListMapper $listMapper)
             ->add('getQuestionsCount', null, [
                 'label' => 'Nombre de questions',
             ])
-            ->add('author.referentTags', null, [
+            ->add('getTagsAsString', null, [
```
ottaviano: could you try to use `array` type without `getTagsAsString`, ex: 
```
add('tags', 'array')
```

thomasbisignani: Nice idea :+1:

PR #3937, opened 2019-08-29
[Donation] Added donators synchronization command

thomasbisignani:


src/Command/DonatorSyncronizeCommand.php
```diff
@@ -0,0 +1,153 @@
...
+                if (!$donator) {
+                    $donator = $this->createDonatorByDonation($donation);
+
+                    $this->manager->persist($donator);
+                }
+            }
+
+            $donator->addDonation($donation);
+
+            if (!\in_array($donation->getStatus(), [Donation::STATUS_REFUNDED, Donation::STATUS_WAITING_CONFIRMATION])) {
```
ottaviano: I think we need to check that:
```
&& $donator->getLastDonationAt() < $donation->getCreatedAt()
```

src/Command/DonatorSyncronizeCommand.php
```diff
@@ -0,0 +1,153 @@
...
+                $donator->setLastDonationAt($donation->getCreatedAt());
+            }
+
+            $cacheDonators[$cacheKey] = $donator;
```
ottaviano: we can move this in the `else` block lines: 67-77, wdyt?

src/Command/DonatorSyncronizeCommand.php
```diff
@@ -0,0 +1,153 @@
...
+            ->setName('app:donator:synchronize')
+            ->setDescription('Synchronize donators with the donations.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $output->writeln(['', 'Starting synchronization.']);
+
+        $progressBar = new ProgressBar($output, $this->getCount());
```
ottaviano: why do not use SymfonyStyle for that and for writeln?

src/Entity/Donator.php
```diff
@@ -13,6 +13,7 @@
  *     name="donators",
  *     uniqueConstraints={
  *         @ORM\UniqueConstraint(name="donator_identifier_unique", columns="identifier"),
+ *     *   @ORM\UniqueConstraint(name="donator_unique_matching", columns={"last_name", "first_name", "email_address"}),
```
ottaviano: you should respect the order that it'll be used: 
for ex, you have these conditions: 
```
->andWhere('donator.emailAddress = :emailAddress')
->andWhere('donator.firstName = :firstName')
->andWhere('donator.lastName = :lastName')
```

your index should be in this order: `emailAddress, firstName, lastName`

PR #3954, opened 2019-09-04
[ChezVous] Introduce MeasureType entity

Remg:


src/Repository/Projection/ReferentManagedUserRepository.php
```diff
@@ -43,7 +42,7 @@ public function search(
         ));
     }
 
-    public function createDispatcherIterator(Adherent $referent, ManagedUsersFilter $filter = null): IterableResult
+    public function createDispatcherIterator(Adherent $referent, ManagedUsersFilter $filter = null): ²
```
ottaviano: :D

Remg: ?
problem ?!

PR #3968, opened 2019-09-18
Import donations command

Remg:


src/Entity/Donation.php
```diff
@@ -55,7 +55,7 @@ class Donation implements GeoPointInterface
     private $gender;
 
     /**
-     * @ORM\Column
+     * @ORM\Column(nullable=true)
```
ottaviano: need to change typehint of return type: 
```php
function getEmailAddress(): ?string
```

migrations/Version20190919134908.php
```diff
@@ -0,0 +1,27 @@
...
+        $this->addSql('ALTER TABLE donators DROP INDEX donator_unique_matching, ADD INDEX IDX_A902FDD7B08E074EA9D1C132C808BA5A (email_address, first_name, last_name)');
+        $this->addSql('ALTER TABLE donators CHANGE email_address email_address VARCHAR(255) DEFAULT NULL');
+        $this->addSql('ALTER TABLE donations ADD type VARCHAR(255) DEFAULT NULL, CHANGE email_address email_address VARCHAR(255) DEFAULT NULL');
+        $this->addSQL('UPDATE donations SET type = \'cb\'');
+        $this->addSql('ALTER TABLE donations CHANGE type type VARCHAR(255) NOT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE donation_transactions ADD type VARCHAR(255) NOT NULL COLLATE utf8_unicode_ci');
```
ottaviano: `ADD type VARCHAR(255) NOT NULL DEFAULT 'cb' `?

src/Entity/Transaction.php
```diff
@@ -99,22 +88,24 @@ class Transaction
      */
     private $donation;
 
-    public function __construct(Donation $donation, array $payboxPayload)
+    public function __construct(Donation $donation, ?array $payboxPayload, ?\DateTimeInterface $donatedAt = null)
```
ottaviano: can we avoid changing the construct of Transaction if you don't use anymore Transaction for Transfer/Check payment?

migrations/Version20190919134908.php
Remg: i dont want a "default" value in the end.

PR #4155, opened 2020-01-06
[SOCLE] Wording fixes & city as a choice

Remg:


front/components/ProgrammaticFoundation/ProgrammaticFoundation.js
```diff
@@ -148,7 +148,25 @@ export default class ProgrammaticFoundation extends React.Component {
                     });
                 });
             });
-        })).sort((a, b) => a.localeCompare(b));
+        })).sort((a, b) => a.localeCompare(b)).sort(function(a,b) {
+            var important_results = {
```
ottaviano: CS: could you use camelCase syntax?

PR #4164, opened 2020-01-10
Improvements on donation backend

Remg:
Screenshot from 2020-01-13 10-50-53

To execute AFTER deploying this PR:
```SQL
UPDATE donations AS donation
SET donation.last_success_date = IF(
	donation.type = 'cb',
	(
		SELECT MAX(transaction.paybox_date_time)
		FROM donation_transactions transaction
		WHERE transaction.donation_id = donation.id
		AND transaction.paybox_result_code = '00000'
	),
	donation.donated_at
)
;
```


ottaviano:
💅

migrations/Version20200114210855.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+final class Version20200114210855 extends AbstractMigration
```
ottaviano: could we merge this migration into the first one?

PR #4208, opened 2020-01-27
Refactor & fix procurations

Remg:


src/Entity/ProcurationProxy.php
```diff
@@ -737,63 +744,47 @@ public function isFrenchRequestAvailable(): bool
...
     {
         return $this->foreignRequestAvailable;
     }
 
-    private function setForeignRequestAvailable(bool $foreignRequestAvailable): void
+    public function process(ProcurationRequest $request): void
     {
-        $this->foreignRequestAvailable = $foreignRequestAvailable;
+        $this->addFoundRequest($request);
+
```
tony-tran: Useless whitespace

Remg: `useless` is a big word :p

i put a whitespace here for readability. 2 separate steps are done in this method.

Remg: but well. ok. you're right, as always Ô @tony-tran

PR #4212, opened 2020-01-29
Use another class for the email issue banner

Remg:


JeremieSamson:
Sorry but what is a `banenr` ?

Remg:
@JeremieSamson it's most likely a banner but without big fingers.

PR #4221, opened 2020-02-04
Move nationality to donators, add donations beneficiary

Remg:


src/Admin/DonationAdmin.php
```diff
@@ -363,6 +386,7 @@ public function getExportFields()
             'ID' => 'id',
             'Montant' => 'amountInEuros',
             'Code don' => 'code',
+            'Bénéficiarie' => 'beneficiary',
```
tony-tran: Bénéficiaire

PR #4231, opened 2020-02-10
Fix USER_CAN_SEND_MESSAGE voter

Remg:


ottaviano:
GG, nice catch @Remg )

PR #4416, opened 2020-04-20
Adherent certification

Remg:


src/Controller/EnMarche/CertificationRequestController.php
```diff
@@ -0,0 +1,64 @@
...
+     * @Route("/demande", name="form", methods={"GET", "POST"})
+     */
+    public function requestAction(Request $request, CertificationManager $certificationManager): Response
+    {
+        $this->disableInProduction();
+
+        /** @var Adherent $adherent */
+        $adherent = $this->getUser();
+
+        if (!$certificationManager->canRequest($adherent)) {
```
ottaviano, line 41: what about moving this logic into a Voter 
```php
if ($this->isGranted(Voter::ABLE_FOR_CERTIFICATION, $adherent)) {
```

src/Entity/CertificationRequest.php
```diff
@@ -0,0 +1,179 @@
...
+    {
+        return $this->status;
+    }
+
+    public function getDocumentName(): ?string
+    {
+        return $this->documentName;
+    }
+
+    public function getDocumentExtension(): ?string
```
ottaviano, line 94: It will be nice to save mimeType directly from `UploadedFile::getMimeType()` for reusing it with `<embed mimetype="..." />` tag on twig, wdyt?

templates/adherent/_profile_menu.html.twig
```diff
@@ -6,6 +6,12 @@
             <a href="{{ path('app_user_profile') }}">Mon compte</a>
         </li>
 
+        {% if is_canary_enabled() %}
+            <li class="{{ route starts with 'app_certification_request_' ? 'active' }}">
```
ottaviano, line 10: you can use the extension too 
```twig
{{ is_active_route(app.request, 'app_certification_request_*') ? 'active' }}
```

PR #4426, opened 2020-04-24
Fix multiple certification requests bug & improvments

Remg:


templates/components/_certification_logo.html.twig
```diff
@@ -0,0 +1 @@
+<img class="b__nudge--left-nano" src="{{ asset('/images/icons/certification/certified-sm.svg') }}" />
```
enzozccr, line 1: Can we keep the b__nudge class out of the component ?

PR #4430, opened 2020-04-29
Voting platform

ottaviano:
tests incoming


ottaviano:
[ch11360]

config/packages/workflows.yaml
```diff
@@ -92,3 +92,36 @@ framework:
...
+            type: 'state_machine'
+            marking_store:
+                type: 'single_state'
+                arguments:
+                    - 'state'
+            initial_place: !php/const AppBundle\VotingPlatform\Election\VoteCommandStateEnum::INITIALISE
+            supports:
+                - AppBundle\VotingPlatform\Election\VoteCommand\VoteCommand
+            places:
+                - !php/const AppBundle\VotingPlatform\Election\VoteCommandStateEnum::INITIALISE
```
kevin-verschaeve, line 105: should be
```suggestion
                - !php/const AppBundle\VotingPlatform\Election\VoteCommandStateEnum::INITIALIZE
```

src/Controller/EnMarche/VotingPlatform/VotingPlatformController.php
```diff
@@ -0,0 +1,157 @@
...
+            return $this->redirectToElectionRoute('app_voting_platform_vote_step', $election);
+        }
+
+        $form = $this
+            ->createForm(ConfirmActionType::class)
+            ->remove('deny')
+            ->handleRequest($request)
+        ;
+
+        if ($form->isSubmitted() && $form->isValid() && $form->get('allow')->isClicked()) {
```
kevin-verschaeve, line 115:
```suggestion
        if ($form->isSubmitted() && $form->isValid()) {
```
don't seems to be needed as you remove the deny button before

src/VotingPlatform/Election/Listener/FinishVoteCommandListener.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private function createVoteResult(Election $election, VoteCommand $command, string $voterKey): VoteResult
+    {
+        $voteResult = new VoteResult($election, $voterKey);
+
+        foreach ($command->getCandidateGroups() as $choice) {
+            $voteChoice = new VoteChoice();
+
+            if (VoteChoice::BLANK_VOTE_VALUE == $choice) {
```
kevin-verschaeve, line 98: === ?

templates/voting_platform/vote_step/_candidate_box.html.twig
```diff
@@ -0,0 +1,44 @@
+<div class="candidate__box l__row  b__nudge--bottom-medium">
+    {{ form_widget(form) }}
+    {{ form_label(form, ' ') }}
```
kevin-verschaeve, line 3: should be false no ? otherwise Symfyony will try to translate `' '`, and it willl be missing

enzozccr: we use label as a clickable area, is there is another way to do it ?

ottaviano: we need to have an invisible `<label>` tag to wrap the box content: 
image

with `false`, `null` value, form does not display `<label>` tag

src/VotingPlatform/Election/Listener/FinishVoteCommandListener.php
ottaviano: I have int type into const and string type coming from POST

src/Controller/EnMarche/VotingPlatform/VotingPlatformController.php
ottaviano: fixed, tnx

config/packages/workflows.yaml
ottaviano: done

templates/voting_platform/finish.html.twig
```diff
@@ -0,0 +1,14 @@
+{% extends base_layout %}
+
+{% block page_title %}
+    À voté!
```
illusionOfParadise, line 4:
```suggestion
    A voté!
```

PR #4433, opened 2020-04-30
Fix adherent filters for non subscribed users

kevin-verschaeve:


tony-tran:
Missing some tests but ok for me

Remg:
:trophy:

PR #4440, opened 2020-05-04
Change role display in message adherent filter

kevin-verschaeve:
[ch11418]

image

In the "exclude" select, the first row is disabled because it is selected in the "include" select

### DIFF
--- a/src/Entity/AdherentMessage/Filter/BasicUserFiltersTrait.php
+++ b/src/Entity/AdherentMessage/Filter/BasicUserFiltersTrait.php
@@ -26,21 +26,21 @@ trait BasicUserFiltersTrait
      *
      * @ORM\Column(type="boolean")
      */
-    private $includeCommitteeSupervisors = true;
+    private $includeCommitteeSupervisors;
 
     /**
      * @var bool
      *
      * @ORM\Column(type="boolean")
      */
-    private $includeCommitteeHosts = true;
+    private $includeCommitteeHosts;
 
     /**
      * @var bool
      *
      * @ORM\Column(type="boolean")
      */
-    private $includeCitizenProjectHosts = true;
+    private $includeCitizenProjectHosts;
 
     /**
      * @var Committee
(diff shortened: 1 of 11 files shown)


src/Entity/AdherentMessage/Filter/BasicUserFiltersTrait.php
```diff
@@ -26,21 +26,21 @@ trait BasicUserFiltersTrait
      *
      * @ORM\Column(type="boolean")
      */
-    private $includeCommitteeSupervisors = true;
+    private $includeCommitteeSupervisors;
```
kevin-verschaeve, line 29: the idea behind this change is : if true, add the filter, if false, add the filter but negationned, if null, just skip the filter. It allows us to not rely on a new field. WDYT ?

ottaviano: yes, this is a good approach for me

src/Form/FilterRoleType.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class FilterRoleType extends AbstractType
+{
+    public const ROLES = [
+        'adherent_message.filter.role.committeeHosts' => 'CommitteeHosts',
+        'adherent_message.filter.role.committeeSupervisors' => 'CommitteeSupervisors',
+        'adherent_message.filter.role.citizenProjectHosts' => 'CitizenProjectHosts',
+    ];
```
Remg, line 15: what about keeping this array "clean" with only values,
and add this in the `setDefaults` method:
```
'choice_label' => function (string $choice) {
    return "adherent_message.filter.role.$choice";
},
```

kevin-verschaeve: done

PR #4445, opened 2020-05-05
[ELECTED REPRESENTATIVE] Add referent tag to elected representative zone

illusionOfParadise:


illusionOfParadise:
[ch11414]

src/Entity/ElectedRepresentative/Zone.php
```diff
@@ -78,4 +101,24 @@ public function __toString(): string
...
+     */
+    public function getReferentTags(): Collection
+    {
+        return $this->referentTags;
+    }
+
+    public function addReferentTag(ReferentTag $referentTag): void
+    {
+        if (!$this->referentTags->contains($referentTag)) {
+            $this->referentTags->add($referentTag);
```
kevin-verschaeve, line 116: as it is a ManyToMany, don't we need to `$referentTag->addZone($this);` ?

PR #4461, opened 2020-05-11
[VotingPlatform] add configuring scripts

ottaviano:


src/Command/VotingPlatformConfigureCommand.php
```diff
@@ -0,0 +1,237 @@
...
+
+        foreach ($memberships as $membership) {
+            $adherent = $membership->getAdherent();
+
+            $list->addVoter($this->voterRepository->findForAdherent($adherent) ?? new Voter($adherent));
+        }
+
+        $this->entityManager->persist($list);
+        $this->entityManager->persist($election);
+        $this->entityManager->flush();
```
kevin-verschaeve, line 142: maybe you should flush just before the clear in the `configureCommitteeElections` method ?

src/Command/VotingPlatformInitializeCommitteeElectionCommand.php
```diff
@@ -0,0 +1,93 @@
...
+    /** @var SymfonyStyle */
+    private $io;
+    /** @var EntityManagerInterface */
+    private $entityManager;
+    /** @var CommitteeRepository */
+    private $committeeRepository;
+
+    protected function configure()
+    {
+        $this->setDescription('Configure Voting Platform');
```
kevin-verschaeve, line 31: it is the same description of the other command, is it normal ?

src/Entity/CommitteeMembership.php
```diff
@@ -230,7 +235,7 @@ public function getCommitteeCandidacy(): ?CommitteeCandidacy
         return $this->committeeCandidacy;
     }
 
-    public function setCommitteeCandidacy(CommitteeCandidacy $committeeCandidacy): void
+    public function setCommitteeCandidacy(?CommitteeCandidacy $committeeCandidacy): void
     {
         $this->committeeCandidacy = $committeeCandidacy;
```
kevin-verschaeve, line 240: no need to the reverse association like here https://github.com/EnMarche/en-marche.fr/pull/4461/files#diff-c2c51603d1b5293687e7848efdc5994dR266 ?

src/Repository/CommitteeMembershipRepository.php
```diff
@@ -606,4 +609,18 @@ public function enableVoteStatusForAdherents(Committee $committee, array $adhere
             ->execute()
         ;
     }
+
+    public function committeeHasVoters(Committee $committee): bool
+    {
+        return (bool) $this->createQueryBuilder('cm')
+            ->select('COUNT(1)')
+            ->where('cm.committee = :committee AND cm.enableVote = :true')
```
kevin-verschaeve, line 617: you can directly use :
```suggestion
            ->where('cm.committee = :committee AND cm.enableVote = 1')
```
no ?

src/Entity/CommitteeMembership.php
ottaviano: no need here, as we are on the owner side of the relation

src/Command/VotingPlatformConfigureCommand.php
ottaviano: here I prefer flush after each committee instead of doing it by batch. If the process fails, it fails for one committee and not for many committees.

src/Command/VotingPlatformInitializeCommitteeElectionCommand.php
ottaviano: fixed, thanks :+1:

ottaviano:
[ch11333]

PR #4473, opened 2020-05-13
[Referent] Stylise edit member form

enzozccr:


src/Form/ReferentPersonLinkType.php
```diff
@@ -126,7 +126,6 @@ function ($coReferentChoices) {
             ))
         ;
 
-        $builder->get('restrictedCommittees')->addModelTransformer($this->committeeTransformer);
```
kevin-verschaeve, line 129: this can also be removed from the __construct

PR #4474, opened 2020-05-13
Access to voting platform

ottaviano:


templates/voting_platform/finish.html.twig
```diff
@@ -18,23 +18,38 @@
...
             </div>
         </div>
     </div>
 {% endblock %}
+
+{% block final_javascripts %}
+    {{ parent() }}
+
+    <script type="text/javascript">
+        Kernel.onLoad(function () { addClass(dom('.envelope'), 'active'); });
```
kevin-verschaeve, line 53: if there no condition to add the class, why not adding it directly in twig ?

ottaviano: oh yes, it's for starting CSS animation only after the page is loaded.

PR #4479, opened 2020-05-14
[REFERENT] Get all ElectedRepresentatives for Paris tag

illusionOfParadise:


features/api/adherents.feature
```diff
@@ -50,6 +50,7 @@ Feature:
     And I am on "/api/statistics/adherents/count-by-referent-area"
     Then the response status code should be 401
 
+    @rol
```
ottaviano, line 53: 🙈

src/Repository/ElectedRepresentative/ElectedRepresentativeRepository.php
```diff
@@ -82,11 +83,31 @@ private function withZoneCondition(QueryBuilder $qb, array $referentTags, string
             $qb->leftJoin($alias.'.mandates', 'mandate');
         }
 
-        return $qb
+        $hasParis = false;
+        foreach ($referentTags as $tag) {
+            if (0 === mb_strpos($tag->getName(), 'Paris')) {
```
ottaviano, line 88: could we check the code instead here? the label can change but the code no

src/Repository/ElectedRepresentative/ElectedRepresentativeRepository.php
```diff
@@ -82,11 +83,31 @@ private function withZoneCondition(QueryBuilder $qb, array $referentTags, string
...
+        }
+
+        $zoneCondition = new Orx();
+        $zoneCondition->add('tag IN (:tags)');
+        $qb->setParameter('tags', $referentTags);
+        // if referent has some Paris tag, we should return elected representatives of all Paris zones
+        if ($hasParis) {
+            $zoneCondition->add('tag.code LIKE :paris_arr OR tag.code LIKE :paris_circo');
+            $qb->setParameter('paris_arr', '750%');
+            $qb->setParameter('paris_circo', 'CIRCO_750%');
```
kevin-verschaeve, line 102: this can lead to unexpected behavior I think because `_` in a `LIKE` means `one char no matter which`. I think we need to escape it here

PR #4482, opened 2020-05-15
[ElectedRepresentative][ADMIN] Add UserListDefinition and its filter

illusionOfParadise:


illusionOfParadise:
[ch11497]

illusionOfParadise:
[ch11498]

src/Entity/ElectedRepresentative/ElectedRepresentative.php
```diff
@@ -118,16 +118,23 @@ class ElectedRepresentative
...
+    private $isSupportingLaREM = false;
 
     /**
      * @var bool|null
      *
-     * @ORM\Column(type="boolean", nullable=true)
+     * @ORM\Column(type="boolean", options={"default": false})
      */
-    private $hasFollowedTraining;
+    private $isInstancesMember = false;
```
kevin-verschaeve, line 130: can we rename it `$instanceMember`, so the getter become `isInstanceMember()` ? (If so, don't forget to update the migration, or the column name here)

illusionOfParadise: There is no more `instanceMember`, the structure has been completely changed.

illusionOfParadise:
[ch11514]

src/Admin/ElectedRepresentative/ElectedRepresentativeAdmin.php
```diff
@@ -451,6 +449,16 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                     }
                 },
             ])
+            ->add('userListDefinitions', ModelAutocompleteFilter::class, [
```
ottaviano, line 452: do we need to check here the specific UserList type for avoiding to display referent user lists?

src/Entity/ElectedRepresentative/ElectedRepresentative.php
```diff
@@ -23,6 +24,7 @@
  */
 class ElectedRepresentative
 {
+    use EntityUserListDefinitionTrait;
```
ottaviano, line 27: an empty line is missing after `use`

src/Entity/UserListDefinition.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->type = $type;
+        $this->label = $label;
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getType(): ?string
```
ottaviano, line 58: `: string`

src/Entity/UserListDefinition.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return $this->type;
+    }
+
+    public function setType(string $type): void
+    {
+        $this->type = $type;
+    }
+
+    public function getLabel(): ?string
```
ottaviano, line 68: `: string`

src/Entity/UserListDefinition.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function setLabel(string $label): void
+    {
+        $this->label = $label;
+    }
+
+    public function __toString(): string
+    {
+        return (string) $this->label;
```
ottaviano, line 80: `return $this->label;`
as $label is required construct string argument

src/Entity/UserListDefinitionEnum.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace App\Entity;
+
+use MyCLabs\Enum\Enum;
+
+final class UserListDefinitionEnum extends Enum
+{
+    public const TYPE_ELECTED_REPRESENTATIVE = 'elected_representative';
+
+    public const LABEL_ELECTED_REPRESENTATIVE_SUPPORTING_LA_REM = 'Sympathisant(e) LaREM';
```
ottaviano, line 11: I think we don't need to have a const for the label as each label will be managed by admin

PR #4491, opened 2020-05-20
[ElectedRepresentative][ADMIN] Add a zone filter

illusionOfParadise:


illusionOfParadise:
[ch11499]

src/Admin/ElectedRepresentative/ElectedRepresentativeAdmin.php
```diff
@@ -426,6 +429,84 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                     return true;
                 },
             ])
+            ->add('mandates.zone', CallbackFilter::class, [
+                'label' => 'Périmètres géographiques',
+                'show_filter' => true,
+                'field_type' => ModelAutocompleteType::class,
```
ottaviano, line 435: maybe we could use `ModelAutocompleteFilter` class instead?

src/Admin/ElectedRepresentative/ElectedRepresentativeAdmin.php
```diff
@@ -426,6 +429,84 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
...
+                            case ZoneCategory::DEPARTMENT:
+                                if (!\in_array('category', $qb->getAllAliases(), true)) {
+                                    $qb->leftJoin('zone.category', 'category');
+                                }
+
+                                if (!\in_array('referentTag', $qb->getAllAliases(), true)) {
+                                    $qb->leftJoin('zone.referentTags', 'referentTag');
+                                }
+
+                                foreach ($zone->getReferentTags() as $key_t => $referentTag) {
```
ottaviano, line 487: will it be enough? 
```php
->andWhere('tags IN (tags) AND category != region')
``` 
?

src/Admin/ElectedRepresentative/ElectedRepresentativeAdmin.php
illusionOfParadise: It's not possible, the class should implement `FormTypeInterface`, that is not the case of `ModelAutocompleteFilter`.

PR #4499, opened 2020-05-22
Liip test query count

ottaviano:


ottaviano:
Liip query count is a nice check, but as mentioned [here](https://github.com/liip/LiipFunctionalTestBundle/blob/master/doc/query.md#caveats) it breaks PHPUnit annotation like `@dataProvider`

PR #4517, opened 2020-06-03
Replace a2lix/i18n-doctrine-bundle by knplabs/doctrine-behaviors

ottaviano:
I18nDoctrineBundle is not maintained anymore 
https://github.com/a2lix/I18nDoctrineBundle


ottaviano:
[ch12734]

PR #4539, opened 2020-06-09
Add validation for required non-mapped checkboxes

Remg:


kevin-verschaeve:
I think we should change the default message for these constraints, as `This value should be true` does not really means "This checkbox should be checked". WDYT ?

PR #4541, opened 2020-06-09
[My Team] Allow senator to delegate their access

kevin-verschaeve:


src/Repository/GeoFilterTrait.php
```diff
@@ -26,6 +26,15 @@ private function applyDeputyGeoFilter(QueryBuilder $qb, Adherent $deputy, string
         $this->applyGeoFilter($qb, [$deputy->getManagedDistrict()->getReferentTag()], $alias);
     }
 
+    private function applySenatorGeoFilter(QueryBuilder $qb, Adherent $senator, string $alias): void
+    {
+        if (!$senator->isDeputy()) {
```
ottaviano: `isSenator`? 
it seems that `applySenatorGeoFilter` is unused

templates/senator/_layout.html.twig
```diff
@@ -27,18 +27,43 @@
                         </h1>
                     </div>
                     <nav class="manager-header__menu">
+                        {% set isSenator = is_granted('ROLE_SENATOR') %}
```
ottaviano, line 30: I think we could avoid this check as we already are in `senator/_layout.html.twig`.
what about doing something like this: 

```twig
<ul>
  {% if is_granted('HAS_DELEGATED_ACCESS_...', 'senator') %}
    <li delegated_route>
  {% else %}
    <li senator_route>
  {% endif %}

  ...

</ul>
```

src/Repository/GeoFilterTrait.php
```diff
@@ -26,6 +26,15 @@ private function applyDeputyGeoFilter(QueryBuilder $qb, Adherent $deputy, string
         $this->applyGeoFilter($qb, [$deputy->getManagedDistrict()->getReferentTag()], $alias);
     }
 
+    private function applySenatorGeoFilter(QueryBuilder $qb, Adherent $senator, string $alias): void
```
ottaviano, line 29: it seems to be unused

templates/senator/_layout.html.twig
kevin-verschaeve: ok done. I have still let it on the last tab, as "my team" should be displayed only to senator

src/Repository/GeoFilterTrait.php
kevin-verschaeve: wow,  good catch ! I indeed forgot to use it ! Fixed

PR #4543, opened 2020-06-09
Change UserListDefinition managing system display

illusionOfParadise:
Updates:
- if labels list is displayed, hide it clicking on `Ajouter un label`.
- can apply labels modification if any, no more need to wait that all labels have defined status.


illusionOfParadise:
[ch11513]

PR #4553, opened 2020-06-11
[My Team] Allow to have 2 delegated accesses of same type (eg. deputy)

kevin-verschaeve:


templates/adherent/_list_my_sections.html.twig
```diff
@@ -14,9 +14,11 @@
     </li>
 {% endif %}
 {% if is_canary_enabled() and is_granted('IS_DELEGATED_REFERENT') %}
-    <li>
-        <a href="{{ path(get_first_access_route(app.request.attributes.get('delegated_accesses'), 'referent')) }}">Espace référent délégué</a>
-    </li>
+    {% for delegated_access in delegated_accesses('referent') %}
```
ottaviano, line 17: an alternative without extension for this:
```twig
{% for x in app.user.receivedDelegatedAccesses if x.isDeputy() %}
  // ...
{% endfor %}
```

to if you want to keep the extension, wdyt: `get_delegated_accesses`?

kevin-verschaeve: what do you think to instead, move this to the adherent entity ?

kevin-verschaeve:
comments fixed

PR #4558, opened 2020-06-15
[Newsletter] Mise à jour des styles

enzozccr:


front/style/layout/_layout.scss
```diff
@@ -67,7 +67,8 @@
     &__wrapper--500 {
         width: 90%;
         max-width: 500px;
-        margin: 0 auto;
+        margin-left: auto;
+        margin-right: auto;
```
kevin-verschaeve, line 71: just asking : what is the difference with before ?

enzozccr: leave the possibility to add a margin top or bottom with another class on the same markeup

PR #4559, opened 2020-06-16
[Certification] translate reason in email

ottaviano:
[ch11609]

### DIFF
--- a/src/Adherent/Certification/CertificationRequestNotifier.php
+++ b/src/Adherent/Certification/CertificationRequestNotifier.php
@@ -41,7 +41,7 @@ class CertificationRequestNotifier
     {
         $refusalReason = $certificationRequest->isRefusedWithOtherReason()
             ? $certificationRequest->getCustomRefusalReason()
-            : $this->translator->trans($certificationRequest->getRefusalReason())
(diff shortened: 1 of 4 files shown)


src/Adherent/Certification/CertificationRequestNotifier.php
```diff
@@ -41,7 +41,7 @@ public function sendRefusalMessage(CertificationRequest $certificationRequest):
     {
         $refusalReason = $certificationRequest->isRefusedWithOtherReason()
             ? $certificationRequest->getCustomRefusalReason()
-            : $this->translator->trans($certificationRequest->getRefusalReason())
+            : $this->translator->trans($certificationRequest->getRefusalReasonKey())
```
kevin-verschaeve, line 44: why not just add the key prefix here ? I dont really like to add this into the entity, would be the last file I check to edit it if needed

ottaviano: for keeping the translation key in the one place (`getRefusalReasonKey`) and avoiding 3 concatenations with this prefix `certification_request.refusal_reason.`

PR #4563, opened 2020-06-18
Delegated/small fixes

kevin-verschaeve:


src/Entity/MyTeam/DelegatedAccess.php
```diff
@@ -70,9 +71,7 @@ class DelegatedAccess
     /**
      * @var array
      *
-     * @ORM\Column(type="simple_array")
-     *
-     * @Assert\Count(min=1)
+     * @ORM\Column(type="simple_array", nullable=true)
```
ottaviano, line 74: why do we need to have nullable access?

kevin-verschaeve: this is a request from Victor. We want to be able to create a member of a team without any accesses

PR #4570, opened 2020-06-24
Fix user can be a string

kevin-verschaeve:


src/Entity/AdherentMessage/AbstractAdherentMessage.php
```diff
@@ -303,8 +303,12 @@ public function setMailchimpCampaigns(array $campaigns): void
         }
     }
 
-    public function isDelegatedAuthor(Adherent $user): bool
+    public function isDelegatedAuthor($user): bool
```
ottaviano, line 306: what about updating this statement and adding `is_granted(ROLE_ADHERENT) and (a or b)`?
with is_granted check before, the method `isDelegatedAuthor` will always receive an Adherent instance 

https://github.com/EnMarche/en-marche.fr/blob/1e7cc7d91a89ce2ab27618f0abdf43b75828e0d2/src/Entity/AdherentMessage/AbstractAdherentMessage.php#L40

kevin-verschaeve: indeed, updated.

PR #4574, opened 2020-06-24
Keep track of UserListDefinitions movements

Remg:
[ch11731]

### DIFF
--- a/src/ElectedRepresentative/UserListDefinitionHistoryManager.php
+++ b/src/ElectedRepresentative/UserListDefinitionHistoryManager.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace App\ElectedRepresentative;
+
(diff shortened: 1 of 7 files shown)


src/ElectedRepresentative/UserListDefinitionHistoryManager.php
```diff
@@ -0,0 +1,64 @@
...
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class UserListDefinitionHistoryManager
+{
+    private $tokenStorage;
+    private $entityManager;
+
+    public function __construct(TokenStorageInterface $tokenStorage, EntityManagerInterface $entityManager)
+    {
+        $this->tokenStorage = $tokenStorage;
```
ottaviano, line 19: what about using `Security` service instead?

PR #4584, opened 2020-06-30
French geo entities

renanbr:
This PR create basis entities for new zone engine.

Tables:

- `geo_country`
- `geo_region`
- `geo_department`
- `geo_canton`
- `geo_district`
- `geo_city_community`
- `geo_city`
    - `geo_city_canton`

Relations are declared in acendenting way, because:

- Populating will run in descending: country, then region, and so on
- Flat zones will be catched from the zone node to its parents, and so on

### DIFF
--- a/src/Entity/Geo/Canton.php
+++ b/src/Entity/Geo/Canton.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace App\Entity\Geo;
+
+use Algolia\AlgoliaSearchBundle\Mapping\Annotation as Algolia;
+use App\Entity\EntityTimestampableTrait;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="geo_canton")
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class Canton implements CollectivityInterface
+{
+    use GeoTrait;
+    use EntityTimestampableTrait;
+
+    /**
+     * @var Department
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Geo\Department")
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private $department;
+
+    /**
+     * @var City[]|Collection
+     *
+     * @ORM\ManyToMany(targetEntity="App\Entity\Geo\City", mappedBy="cantons")
+     */
+    private $cities;
+
+    public function __construct(string $code, string $name, Department $department)
+    {
+        $this->code = $code;
+        $this->name = $name;
+        $this->department = $department;
+        $this->cities = new ArrayCollection();
+    }
+
+    public function getDepartment(): Department
+    {
+        return $this->department;
+    }
+
+    public function setDepartment(Department $department): void
+    {
+        $this->department = $department;
+    }
+
+    /**
+     * @return City[]|Collection
+     */
+    public function getCities(): Collection
+    {
+        return $this->cities;
+    }
+
+    public function getParents(): array
+    {
+        $parents = [];
+
+        $parents[] = $department = $this->getDepartment();
+        if ($department) {
+            $parents = array_merge($parents, $department->getParents());
+        }
+
+        return $this->sanitizeEntityList($parents);
+    }
+}
--- a/src/Entity/Geo/City.php
+++ b/src/Entity/Geo/City.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace App\Entity\Geo;
+
+use Algolia\AlgoliaSearchBundle\Mapping\Annotation as Algolia;
+use App\Entity\EntityTimestampableTrait;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="geo_city")
+ *
+ * @Algolia\Index(autoIndex=false)
+ */
+class City implements CollectivityInterface
+{
+    use GeoTrait;
+    use EntityTimestampableTrait;
+
+    /**
+     * @var string[]|null
+     *
+     * @ORM\Column(type="simple_array", nullable=true)
+     */
+    private $postalCode;
+
+    /**
+     * @var int|null
+     *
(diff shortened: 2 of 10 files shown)


src/Entity/Geo/Canton.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * @var Department
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Geo\Department")
+     * @ORM\JoinColumn(nullable=false)
+     */
+    public $department;
+
+    public function __construct(string $id, string $name, Department $department)
```
ottaviano, line 38: we don't pass the id, we keep Doctrine manage it. (same for other constructs)

src/Entity/Geo/Community.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Algolia\AlgoliaSearchBundle\Mapping\Annotation as Algolia;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="geo_community")
+ * @Algolia\Index(autoIndex=false)
+ */
+class Community
```
ottaviano, line 13: `CityCommunity`?

src/Entity/Geo/Canton.php
```diff
@@ -0,0 +1,44 @@
...
+ * @ORM\Table(name="geo_canton")
+ * @Algolia\Index(autoIndex=false)
+ */
+class Canton
+{
+    /**
+     * @var string
+     *
+     * @ORM\Id
+     * @ORM\Column
```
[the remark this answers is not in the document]
renanbr: I was assuming we'll external code (that may starts with `0`) as identifier.

migrations/Version20200630153958.php
```diff
@@ -0,0 +1,124 @@
+<?php
+
+declare(strict_types=1);
```
kevin-verschaeve, line 3: you can remove this

src/Entity/Geo/Canton.php
```diff
@@ -0,0 +1,44 @@
...
+    public $name;
+
+    /**
+     * @var Department
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Geo\Department")
+     * @ORM\JoinColumn(nullable=false)
+     */
+    public $department;
+
```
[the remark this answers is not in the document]
renanbr: https://github.com/EnMarche/en-marche.fr/pull/4584#discussion_r447715094

src/Entity/Geo/Canton.php
renanbr: https://github.com/EnMarche/en-marche.fr/pull/4584#discussion_r447715094

src/Entity/Geo/Canton.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+class Canton
+{
+    /**
+     * @var string
+     *
+     * @ORM\Id
+     * @ORM\Column
+     */
+    public $id;
```
kevin-verschaeve, line 21: I think in the rest of the project, we are not using public properties for entities, maybe we should set them private and use getter/setter ?

src/Entity/Geo/Canton.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class Canton
+{
+    use GeoTrait;
+
+    /**
+     * @var Department|null
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Geo\Department")
+     * @ORM\JoinColumn(nullable=false)
```
ottaviano, line 27: if the relatoin is `nullable=false`:
- `@var Department|null` => `@var Department`
- we should have `__construct(Department $department)`
- `getDepartment(): ?Department` => `getDepartment(): Department`

(same for other properties in the same case)

src/Entity/Geo/City.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    private $canton;
+
+    /**
+     * @var CityCommunity|null
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Geo\CityCommunity")
+     * @ORM\JoinColumn(nullable=true)
+     */
+    private $community;
```
ottaviano, line 45: $cityCommunity

src/Entity/Geo/GeoTrait.php
```diff
@@ -0,0 +1,56 @@
...
+     * @ORM\Id
+     * @ORM\Column(type="integer", options={"unsigned": true})
+     * @ORM\GeneratedValue
+     */
+    private $id;
+
+    /**
+     * @var string|null
+     *
+     * @ORM\Column
```
ottaviano, line 21: what about making this column `unique` here and avoid all `@ORM\UniqueConstraint(name="...", columns="code")`
```
* @ORM\Column(unique=true)
```

src/Entity/Geo/Canton.php
renanbr: for all classes that belongs to department (1) I created a `UnderDepartmentTrait` to handle this attribute and its getter and setter, and (2) declared it in constructor

renanbr:
There are some orphaned cities (no department nor region).

```json
{
    "nom": "Thio",
    "code": "98829",
    "codesPostaux": []
}
```
Source: https://geo.api.gouv.fr/communes/98829

We need some decision about them before merging it

renanbr:
Update from 94930e6a9083c42f717f3f9c48ef2f8f500fb348:

- Mark `$departement` as optional in `City` entity
- Add `$population` to `City`
- Added `$createdAt` and `$updatedAt` attributes to all geo entities

src/Entity/Geo/GeoTrait.php
```diff
@@ -0,0 +1,85 @@
...
+
+    public function activate(bool $active = true): void
+    {
+        $this->active = $active;
+    }
+
+    private function sanitizeEntityList(array $entities): array
+    {
+        $unique = [];
+        foreach (array_filter($entities) as $entity) {
```
ottaviano, line 77: what about using `array_unique` instead of `foreach`:
```php
return array_unique(array_filter($entities));
```

ottaviano (approved):

src/Entity/Geo/GeoTrait.php
renanbr: `array_unique()` converts objets into string to detect duplications, it generates errors during my tests

https://3v4l.org/uUT1H

PR #4594, opened 2020-07-06
Add delegated access role labels for referents

kevin-verschaeve:


src/Controller/AccessDelegatorTrait.php
```diff
@@ -31,4 +33,21 @@ protected function getDelegatedAccess(Request $request): DelegatedAccess
 
         return $delegatedAccess;
     }
+
+    protected function getType(): ?string
+    {
+        if (\method_exists($this, 'getSpaceType')) {
```
ottaviano, line 39: we could refactor this after for using interface instead

kevin-verschaeve: Yes, this is not the same everywhere :'( ^^

PR #4599, opened 2020-07-07
Add elected representative tabs to senators

kevin-verschaeve:
[ch11801]

### DIFF
--- a/templates/senator/_layout.html.twig
+++ b/templates/senator/_layout.html.twig
@@ -74,9 +74,12 @@
                             {% endif %}
 
                             {% if is_granted('ROLE_SENATOR') %}
-                            <li class="{{ is_active_route(app.request, ['app_senator_my_team_list', 'app_senator_my_team_delegate_access']) ? 'active' }}">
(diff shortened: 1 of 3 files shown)


templates/senator/_layout.html.twig
```diff
@@ -73,6 +73,12 @@
                             </li>
                             {% endif %}
 
+                            {% if is_granted('ROLE_SENATOR') %}
```
ottaviano, line 76: I see we check `is_granted('ROLE_SENATOR')` many times in this template. 
Could you refactor this for checking once this role, please?

kevin-verschaeve: fixed :)

ottaviano: the best will be: 
```twig
{% if is_senator %}
  <li>ITEM 1</li>
{% endif %}

{% if is_senator %}
  <li>ITEM 2</li>
{% endif %}
```
=> 
```twig
{% if is_senator %}
  <li>ITEM 1</li>
  <li>ITEM 2</li>
{% endif %}
```

kevin-verschaeve: indeed, re fixed

PR #4601, opened 2020-07-08
Remove canary mode on delegated access

kevin-verschaeve:


src/Controller/AccessDelegatorTrait.php
```diff
@@ -10,10 +10,6 @@ trait AccessDelegatorTrait
 {
     protected function getMainUser(Request $request): UserInterface
     {
-        if ('referent' !== $this->getType()) {
-            $this->disableInProduction();
-        }
-
         return $this->getDelegatedAccess($request)->getDelegator();
```
renanbr, line 13: `DelegatedAccess->getDelegator()` may returns `null`

PR #4606, opened 2020-07-09
[LRE] Add LRE space

kevin-verschaeve:
"La république ensemble"

[ch11802]


templates/adherent/_list_my_sections.html.twig
```diff
@@ -102,3 +102,10 @@
         </a>
     </li>
 {% endif %}
+{% if app.user.isLre %}
```
renanbr, line 105: Should we check for `ROLE_LRE` instead?

kevin-verschaeve: I did the same as the other checks, anyway, in this case, this is the same, as ROLE_LRE is present when isLre returns true :)

src/Collection/AdherentCharterCollection.php
```diff
@@ -48,6 +49,13 @@ public function hasSenatorialCandidateCharterAccepted(): bool
         });
     }
 
+    public function hasLRECharterAccepter()
```
ottaviano, line 52: typehint please

src/Collection/AdherentCharterCollection.php
```diff
@@ -65,6 +73,9 @@ public function hasCharterAcceptedForType(string $type): bool
 
             case AdherentCharterTypeEnum::TYPE_SENATORIAL_CANDIDATE:
                 return $this->hasSenatorialCandidateCharterAccepted();
+
+            case AdherentCharterTypeEnum::TYPE_LRE:
+                return $this->hasLRECharterAccepter();
```
ottaviano, line 78: `hasLRECharterAccepted`

src/Entity/Adherent.php
```diff
@@ -456,6 +456,14 @@ class Adherent implements UserInterface, UserEntityInterface, GeoPointInterface,
      */
     private $senatorialCandidateManagedArea;
 
+    /**
+     * @var LREArea|null
+     *
+     * @Assert\Valid
```
ottaviano, line 462: It seems that `LREArea` entity does not have any `@Assert`, is it an oversight?

src/Entity/LREArea.php
```diff
@@ -0,0 +1,45 @@
...
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     */
+    private $id;
+
+    /**
+     * @var ReferentTag|null
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\ReferentTag")
```
ottaviano, line 27: do we really need to keep it nullable? if not, I propose to do: 
- `@JoinColumn(nullable=false)`
- `__construct(ReferentTag $tag)`

src/Entity/LreArea.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->departmentTag = $tag;
+    }
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    public function getDepartmentTag(): ?ReferentTag
```
ottaviano, line 42: `?ReferentTag` => `ReferentTag`

src/Entity/LreArea.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return $this->departmentTag;
+    }
+
+    public function setDepartmentTag(?ReferentTag $departmentTag): void
```
ottaviano, line 47: `?ReferentTag` => `ReferentTag`

PR #4641, opened 2020-07-29
Add TerritorialCouncil (entity + admin + command)

illusionOfParadise:


src/Command/ImportTerritorialCouncilFromCSVCommand.php
```diff
@@ -0,0 +1,124 @@
...
+            if ($territorialCouncil = $this->em->getRepository(TerritorialCouncil::class)->findOneBy(['name' => $data['name']])) {
+                $territorialCouncil->clearReferentTags();
+            } else {
+                $territorialCouncil = new TerritorialCouncil($data['name'], $data['codes']);
+            }
+
+            $tags = explode(',', $data['tag_code']);
+            foreach ($tags as $tag) {
+                $tag = trim($tag);
+                if (!($referentTag = $this->referentTagRepository->findOneByCode($tag))) {
```
kevin-verschaeve, line 74: maybe you can find all with codes : `findBy(['code' => explode(',', $data['tag_code'])` and browse results, so you have one query instead of one per code ?

src/Command/ImportTerritorialCouncilFromCSVCommand.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        $firstline = true;
+
+        while (false !== ($data = fgetcsv($handle, 10000, ','))) {
+            if ($firstline) {
+                $firstline = false;
+
+                continue;
+            }
```
kevin-verschaeve, line 111: If you want, to remove this if and the `$firstline` variable, you can call `fgetcsv` before the foreach, and do nothin with it. This will read the first line, and skip to the next one.

src/Command/ImportTerritorialCouncilFromCSVCommand.php
illusionOfParadise: Yes, I can find all codes like this, but in this case I don't know if all passed referent tags have been found.
If I do one by one, I can display errors if there are in CSV file.

illusionOfParadise:
[ch11632]

src/Entity/TerritorialCouncil/TerritorialCouncilMembership.php
```diff
@@ -0,0 +1,133 @@
...
+    ];
+
+    /**
+     * @var int|null
+     *
+     * @ORM\Id
+     * @ORM\Column(type="integer", options={"unsigned": true})
+     * @ORM\GeneratedValue
+     */
+    private $id;
```
Remg, line 33: can we just use EntityIdentityTrait here too ?

i am sure we will need to display pages, in frontend, of memberships (in a future manager controller or idk)

src/Entity/TerritorialCouncil/TerritorialCouncilMembership.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->roles[] = $role;
+        $this->joinedAt = $joinedAt ?? new \DateTimeImmutable('now');
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getTerritorialCouncil(): ?TerritorialCouncil
```
Remg, line 84: return type is not nullable here

src/Entity/TerritorialCouncil/TerritorialCouncilMembership.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return $this->territorialCouncil;
+    }
+
+    public function getAdherent(): ?Adherent
```
Remg, line 89: return type is not nullable here

src/Entity/TerritorialCouncil/TerritorialCouncilMembership.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return $this->adherent;
+    }
+
+    public function getRoles()
```
Remg, line 94: `: array` ?

PR #4657, opened 2020-08-05
Fix display lre labels only in LRE space

kevin-verschaeve:


src/Collection/UserListDefinitionCollection.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace App\Collection;
+
+use App\Entity\UserListDefinition;
+use App\Entity\UserListDefinitionEnum;
+use Doctrine\Common\Collections\ArrayCollection;
+
+class UserListDefinitionCollection extends ArrayCollection
+{
+    public function forType(string $type)
```
ottaviano, line 11: `: self`

PR #4663, opened 2020-08-07
[LRE] Allow to disable zone restriction and see all elected represent…

kevin-verschaeve:
…atives

### DIFF
--- a/src/Entity/LreArea.php
+++ b/src/Entity/LreArea.php
@@ -4,6 +4,7 @@ namespace App\Entity;
 
 use Algolia\AlgoliaSearchBundle\Mapping\Annotation as Algolia;
 use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * @ORM\Entity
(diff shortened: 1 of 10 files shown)


src/Entity/LreArea.php
```diff
@@ -48,4 +57,35 @@ public function setReferentTag(?ReferentTag $referentTag): void
...
+
+    /**
+     * @Assert\Callback
+     */
+    public function validate(ExecutionContextInterface $context, $payload)
+    {
+        /** @var LreArea $lreArea */
+        $lreArea = $context->getValue();
+
+        if (!$lreArea->isAllTags()) {
```
ottaviano, line 79: potentially we can have `LreArea` without `refTag` and without the option `AllTags` and it will be valid 🤔

kevin-verschaeve: no, if this is the case, so the listener [here](https://github.com/EnMarche/en-marche.fr/pull/4663/files/9162aab36799d8945d1cd2bc8a77cae63a06b996#diff-e8cde86e76cde95a03e74a240541a646R33) will remove the lreArea to the adherent, like before

ottaviano: I meen this: 
```php
$o = new LreArea();
$errors = $validator->validate($o);

// will output `int(0)`
var_dump(count($errors));
```

PR #4640, opened 2020-07-24
Refacto/profile

kevin-verschaeve:

PR #4667, opened 2020-08-07
Activity/instances

kevin-verschaeve:
based on #4640

### DIFF
--- a/templates/adherent_profile/activity/_instances.html.twig
+++ b/templates/adherent_profile/activity/_instances.html.twig
@@ -1,0 +1,29 @@
+<h1 class="text--medium-large">Mes instances</h1>
+
+<div class="l__col">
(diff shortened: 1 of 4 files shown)


templates/adherent_profile/activity/_instances.html.twig
```diff
@@ -0,0 +1,31 @@
+<h1 class="text--medium-large">Mes instances</h1>
+
+<div class="l__col">
+    {% if app.user.territorialCouncilMembership %}
```
ottaviano, line 4: IMO we can avoid this check as we already test it in `activity.html.twig` template

ottaviano (approved):

templates/adherent_profile/activity/_instances.html.twig
kevin-verschaeve: indeed, removed

PR #4720, opened 2020-08-19
Revoke delegated access when delegator loose his role

kevin-verschaeve:


src/EntityListener/RevokeReferentTeamMemberRolesListener.php
```diff
@@ -49,6 +50,8 @@ public function preFlush(Adherent $adherent): void
             $personLink->setIsMunicipalManagerSupervisor(false);
         }
 
+        $this->entityManager->getRepository(DelegatedAccess::class)->removeFromDelegator($adherent);
```
ottaviano, line 53: we should handle this change in `src/EntityListener/RevokeDelegatedAccessListener.php` too as this listener is part of the old system and will be removed

ottaviano (approved):

src/EntityListener/RevokeReferentTeamMemberRolesListener.php
kevin-verschaeve: updated

PR #4729, opened 2020-08-21
Add adherent readonly role on admin

kevin-verschaeve:


ottaviano (approved):

templates/admin/adherent/_summary.html.twig
```diff
@@ -2,6 +2,10 @@
     <a href="{{ path('admin_app_adherent_edit', { id: adherent.id }) }}">
         {{ adherent.fullName }}
     </a>
+{% elseif is_granted('ROLE_APP_ADMIN_ADHERENT_SHOW') %}
```
illusionOfParadise, line 5: It seams that `ROLE_APP_ADMIN_ADHERENT_SHOW` role does not exist and it should be `ROLE_APP_ADMIN_ADHERENT_VIEW` instead🤔

kevin-verschaeve: wow indeed, good catch

illusionOfParadise (approved):

PR #4770, opened 2020-09-04
[VotingPlatform] fix return back button

ottaviano:


templates/voting_platform/confirmation.html.twig
```diff
@@ -30,7 +32,8 @@
                                     <h3 class="text--medium">Bulletin blanc</h3>
                                 {% endif %}
                             </div>
-                            <a href="{{ path('app_voting_platform_vote_step', {uuid: election.uuid}) }}" class="text--body text--center link--no-decor link--blue--dark">
+
+                            <a href="{{ path('app_voting_platform_vote_step', {uuid: election.uuid, s: loop.index}) }}" class="text--body text--center link--no-decor link--blue--dark">
```
kevin-verschaeve, line 36: you can use `loop.index0`, so in your controller you don't need to `-1` : `$currentPool = $pools[$step];`

kevin-verschaeve (approved):

templates/voting_platform/confirmation.html.twig
ottaviano: as `$request->query->getInt(..)` can return `0` for an unexisting argument, I prefer to pass the value `>= 1` to distinguish the case when the argument is omitted

PR #4807, opened 2020-09-10
Add borough councilor

renanbr:


translations/messages.fr.yml
```diff
@@ -125,6 +125,7 @@ adherent.mandate.city_councilor: Conseiller municipal
 adherent.mandate.parliamentary: Parlementaire
 adherent.mandate.european_deputy: Député européen
 adherent.mandate.consular_conselor: Conseiller consulaire
+adherent.mandate.borough_councilor: Conseiller(ère) d'arrondissement
```
ottaviano, line 128: I think we could rollback this change too, as it is the old mandate system.

ottaviano (commented):
do we need to validate the mandate zone when this new type is used?

ottaviano (approved):

illusionOfParadise (approved):

PR #4810, opened 2020-09-10
[Committee] Display designed adherents in election box

ottaviano:
[ch11782]

### DIFF
--- a/src/DataFixtures/ORM/LoadVotingPlatformElectionData.php
+++ b/src/DataFixtures/ORM/LoadVotingPlatformElectionData.php
@@ -224,8 +224,12 @@ class LoadVotingPlatformElectionData extends Fixture implements DependentFixture
 
         /** @var CommitteeCandidacy $committeeCandidacy */
         $committeeCandidacy = $this->getReference('committee-candidacy-1');
-        $candidate1->setImagePath($committeeCandidacy->getImagePath());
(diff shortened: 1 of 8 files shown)


src/DataFixtures/ORM/LoadVotingPlatformElectionData.php
```diff
@@ -224,8 +224,12 @@ private function getCandidates(): array
 
         /** @var CommitteeCandidacy $committeeCandidacy */
         $committeeCandidacy = $this->getReference('committee-candidacy-1');
-        $candidate1->setImagePath($committeeCandidacy->getImagePath());
-        $candidate2->setImagePath($committeeCandidacy->getImagePath());
+
+        array_map(function (Candidate $candidate) use ($committeeCandidacy) {
+            if (rand(0, 1)) {
```
renanbr, line 229: to make sure we'll have both cases we could check something like `0 === $count++ % 2` instead of random bool

kevin-verschaeve (approved):

renanbr (approved):
just a miner comment

src/DataFixtures/ORM/LoadVotingPlatformElectionData.php
ottaviano: good idea, I will change it in the next PR 👍

PR #4818, opened 2020-09-14
[VotingPlatform] remove title field from ElectionPool and use code in…

ottaviano:
…stead


templates/committee_designation/votes_list.html.twig
```diff
@@ -24,9 +24,9 @@
 
                 <tr>
                     <td>{{ vote.voterKey }}</td>
-                    {% for pool_title in ['Femme', 'Homme'] %}
+                    {% for pool_code in ['female', 'male'] %}
```
renanbr, line 27: Could we use `constant('App\ValueObject\Genders::*')` in all twig files?

renanbr (changes requested):

templates/committee_designation/votes_list.html.twig
ottaviano: @renanbr yes, fixed 👍 
```twig
{% for pool_code in constant('App\\ValueObject\\Genders::MALE_FEMALE') %}
```

renanbr (approved):

PR #4845, opened 2020-09-22
[Thematic Communities] Set chief from adherent admin

kevin-verschaeve:


ottaviano (approved):

src/Entity/Adherent.php
```diff
@@ -591,6 +592,13 @@ class Adherent implements UserInterface, UserEntityInterface, GeoPointInterface,
      */
     private $commitment;
 
+    /**
+     * @var ThematicCommunity[]|Collection
+     *
+     * @ORM\ManyToMany(targetEntity="App\Entity\ThematicCommunity\ThematicCommunity")
+     */
+    private $thematicCommunities;
```
ottaviano, line 600: What about using `thematicCommunitiesChiefArea` as a name to distinguish from CT memberships?

kevin-verschaeve: this is not really linked to areas. maybe something like `$handledThematicCommunity` ?

PR #4855, opened 2020-09-25
Update flysysteme service

ottaviano:


composer.json
```diff
@@ -135,6 +134,7 @@
         "fabpot/goutte": "^3.2",
         "friendsofphp/php-cs-fixer": "^2.16",
         "fzaninotto/faker": "^1.9",
+        "league/flysystem-memory": "^1.0",
```
kevin-verschaeve, line 137: why did you moved this ine require-dev part ? It seems used in prod env too

ottaviano: @kevin-verschaeve, it seems to be used only on test, no?
image

kevin-verschaeve: my bad, misread the package name !

kevin-verschaeve (approved):

ottaviano:
[ch12734]

PR #4856, opened 2020-09-25
Fix admin can not impersonate if he has readonly right

kevin-verschaeve:


templates/admin/adherent/action_button_switch_user.html.twig
```diff
@@ -1,4 +1,4 @@
-{% if object.enabled %}
+{% if object.enabled and admin.hasAccess('edit') %}
```
ottaviano, line 1: `is_granted('ROLE_ALLOWED_TO_SWITCH')` will be better

kevin-verschaeve: no actually, I want to check if he has the role `ROLE_ADMIN_ADHERENTS` or only `ROLE_ADMIN_ADHERENTS_READONLY` (if he has readonly, he has not access on edit for adherents)

kevin-verschaeve:
[ch12254]

ottaviano (approved):

illusionOfParadise (approved):

PR #4861, opened 2020-09-29
Export national survey data

kevin-verschaeve:
[ch12440]

### DIFF
--- a/src/Repository/Jecoute/DataSurveyRepository.php
+++ b/src/Repository/Jecoute/DataSurveyRepository.php
@@ -3,7 +3,9 @@
 namespace App\Repository\Jecoute;
 
 use App\Entity\Jecoute\DataSurvey;
+use App\Entity\Jecoute\Survey;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\ORM\Internal\Hydration\IterableResult;
 use Symfony\Bridge\Doctrine\RegistryInterface;
 
 class DataSurveyRepository extends ServiceEntityRepository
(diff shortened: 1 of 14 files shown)


src/Repository/Jecoute/DataSurveyRepository.php
```diff
@@ -30,4 +31,14 @@ public function countByEmailAnsweredForOneMonth(string $email, \DateTime $posted
             ->getSingleScalarResult()
         ;
     }
+
+    public function iterateForSurvey(Survey $survey)
```
ottaviano, line 35: typehint

templates/admin/jecoute/exports.html.twig
```diff
@@ -0,0 +1,14 @@
+{% extends 'SonataAdminBundle:CRUD:base_list_field.html.twig' %}
+
+{% block field %}
+    {% if admin.hasAccess('show') %}
```
ottaviano, line 4: do we need to do the same check in the controller?

src/Exporter/NationalSurveyExporter.php
```diff
@@ -0,0 +1,87 @@
...
+                (new Slugify())->slugify($survey->getName()),
+                $survey->getId(),
+                (new \DateTime())->format('YmdHis'),
+                $format
+            ),
+            new IteratorCallbackSourceIterator($this->dataSurveyRepository->iterateForSurvey($survey), function (array $data) use ($survey, $allowedOnly) {
+                /** @var DataSurvey $dataSurvey */
+                $dataSurvey = $data[0];
+
+                $allowPersonalData = ($allowedOnly && $dataSurvey->getAgreedToTreatPersonalData()) || !$allowedOnly;
```
ottaviano, line 49: is it the same? 
```php
$allowPersonalData = !$allowedOnly || $dataSurvey->getAgreedToTreatPersonalData();
```

templates/admin/jecoute/exports.html.twig
kevin-verschaeve: yes indeed, and I actually moved this check into the Admin class, so the column is not added if we don't have the access

ottaviano (approved):

PR #4872, opened 2020-10-01
Do not export all data from front and fixes from questions not answered by everyone

kevin-verschaeve:
questions not answered by everyone questions where not added in the header row

### DIFF
--- a/src/Controller/EnMarche/Jecoute/AbstractJecouteController.php
+++ b/src/Controller/EnMarche/Jecoute/AbstractJecouteController.php
@@ -166,7 +166,7 @@ abstract class AbstractJecouteController extends Controller
         SurveyExporter $exporter
     ): Response {
         if ($format = $request->query->get('export')) {
-            return $exporter->export($survey, $format, true);
+            return $exporter->export($survey, $format, false);
         }
 
         return $this->renderTemplate('jecoute/stats.html.twig', ['data' => $provider->getStatsBySurvey($survey)]);
--- a/src/Exporter/SurveyExporter.php
+++ b/src/Exporter/SurveyExporter.php
@@ -39,7 +39,7 @@ class SurveyExporter
         $this->translator = $translator;
     }
 
-    public function export(Survey $survey, string $format, bool $allowedOnly = true): StreamedResponse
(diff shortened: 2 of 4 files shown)


src/Controller/EnMarche/Jecoute/AbstractJecouteController.php
```diff
@@ -166,7 +166,7 @@ public function jecouteSurveyStatsAction(
         SurveyExporter $exporter
     ): Response {
         if ($format = $request->query->get('export')) {
-            return $exporter->export($survey, $format, true);
+            return $exporter->export($survey, $format, false);
```
illusionOfParadise, line 169: You can not to specify `false`:
```suggestion
            return $exporter->export($survey, $format);
```

src/Exporter/SurveyExporter.php
```diff
@@ -50,41 +51,55 @@ public function export(Survey $survey, string $format, bool $allowedOnly = true)
...
+                if ($fromAdmin) {
+                    $row['Accepte que ses données soient traitées'] = (int) $dataSurvey->getAgreedToTreatPersonalData();
+                }
+
+                $row['Profession'] = $allowPersonalData && $dataSurvey->getProfession() ? $dataSurvey->getProfession() : null;
 
                 /** @var SurveyQuestion $surveyQuestion */
                 foreach ($survey->getQuestions() as $surveyQuestion) {
+                    $questionName = $surveyQuestion->getQuestion()->getContent();
+                    $row[$surveyQuestion->getQuestion()->getContent()] = '';
```
illusionOfParadise, line 91: What about:
```suggestion
                    $row[$questionName] = '';
```
?

src/Controller/EnMarche/Jecoute/AbstractJecouteController.php
kevin-verschaeve: yes but, in some cases, like that, I prefer to precise it, so we know that it is intentional and that we really want this behaviour

ottaviano (approved):

src/Exporter/SurveyExporter.php
kevin-verschaeve: yes good catch, I created a variable for that, and forgot to use it !

PR #4890, opened 2020-10-06
[Certification] convert a pdf to an image before send to vision

kevin-verschaeve:


src/Vision/ImageAnnotations.php
```diff
@@ -8,7 +8,11 @@ class ImageAnnotations
 {
     private const IDENTITY_DOCUMENT_LABEL = 'Identity document';
     private const NATIONAL_IDENTITY_CARD_LABEL = 'National identity card';
-    private const FRENCH_IDENTITY_CARD_LABEL = 'carte d identité française';
+    private const FRENCH_LANGUAGE = 'French language';
+    private const FRENCH_IDENTITY_CARD_LABELS = [
+        'carte d identité française',
+        'carte d identité verso',
```
kevin-verschaeve, line 14: on my test I had this, so I added it here because I think it is valid too. Do you agree ?

src/Vision/ImageAnnotations.php
```diff
@@ -71,12 +75,15 @@ public function isIdentityDocument(): bool
 
     public function isFrenchNationalIdentityCard(): bool
     {
-        return $this->isIdentityDocument() && \in_array(self::FRENCH_IDENTITY_CARD_LABEL, $this->labels, true);
+        return $this->isIdentityDocument() && (
+            \count(\array_intersect(self::FRENCH_IDENTITY_CARD_LABELS, $this->labels)) > 0 ||
+            \in_array(self::FRENCH_LANGUAGE, $this->webEntities, true)
```
kevin-verschaeve, line 80: on some test, my `labels` only contains `[document]` which is not helpful. However my idendity card was a good one, but Google seems to not parse correctly. So If it is a n identity document, this check allows now to pass if it find any of the french authorized labels or if it find the `French language` webEntity; What do you think ?

src/Vision/ImageAnnotations.php
```diff
@@ -71,12 +75,15 @@ public function isIdentityDocument(): bool
...
+        );
     }
 
     public function getFirstNames(): array
     {
-        preg_match('/Prénom( )?\(s\)( )?:( )?(?<first_names>.+)\\n/', $this->text, $matches);
+        preg_match('/Prénom\s?((\(s\))?|s)\s?:\s?(?<first_names>.+)\\n/', $this->text, $matches);
```
kevin-verschaeve, line 86: on my tests, I did not have the `(s)` part on `Prénom(s)` but,  `Prénoms`; So this regex allow this too now

src/Image/PdfToImageConverter.php
```diff
@@ -19,7 +25,7 @@ public function toImage(string $pdfPath, string $format = 'jpg')
 
         $imagePath = \str_replace('pdf', $format, $pdfPath);
 
-        file_put_contents($this->dataDirectory.$imagePath, $imagick);
+        $this->filesystem->dumpFile($this->dataDirectory.$imagePath, $imagick);
```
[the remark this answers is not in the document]
kevin-verschaeve: yes, I was using the Symfony one. Changed to the League one

src/Vision/ImageAnnotations.php
Remg: i am not so sure.
we specifically ask for the first side of the card, not the other side.

src/Vision/ImageAnnotations.php
Remg: label `French language` is not mandatory, no

src/Entity/CertificationRequest.php
```diff
@@ -67,6 +67,7 @@ class CertificationRequest
      *     mimeTypes={
      *         "image/jpeg",
      *         "image/png",
+     *         "application/pdf",
```
Remg, line 70: we should have one array of these valid mime types and use it everywhere

src/Vision/ImageAnnotations.php
kevin-verschaeve: reverted

src/Vision/ImageAnnotations.php
kevin-verschaeve: reverted

translations/validators.fr.yml
```diff
@@ -453,7 +453,7 @@ elected_representative.invalid_adherent: Cet e-mail correspond à un compte adh
 #
 # CertificationRequest
 #
-certification_request.document.mime_type: Seuls les fichiers jpg et png sont acceptés.
+certification_request.document.mime_type: Seuls les fichiers jpg, png et pdf sont acceptés.
```
ottaviano, line 456: I think we should update this template too: 
https://github.com/EnMarche/en-marche.fr/blob/8c78625fbe725bfba31011fa8d7e2f32ff496c8c/templates/certification_request/form.html.twig#L32

ottaviano (approved):

translations/validators.fr.yml
kevin-verschaeve: yes, I can do it in my next PR, as I will work mainly on this page

kevin-verschaeve:
[ch12483]

PR #4905, opened 2020-10-12
Add borough level to the geo mapping

renanbr:
~Enhance `app:geo:update-france` to handle boroughs.~

Create `app:geo:update-boroughs` to handle french boroughs.

Zones are updated as well.

No official data is used at this time.

### DIFF
--- a/src/Entity/Geo/Borough.php
+++ b/src/Entity/Geo/Borough.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace App\Entity\Geo;
+
+use App\Entity\EntityTimestampableTrait;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="geo_borough")
+ */
+class Borough implements ZoneableInterface
+{
+    use GeoTrait;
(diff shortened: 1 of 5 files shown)


renanbr:
By merging this PR we can solve almost 100% of ongoing mandates references to the new mapping.

src/Entity/Geo/Borough.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public function getParents(): array
+    {
+        $toMerge = [];
+
+        $toMerge[] = [$this->city];
+        $toMerge[] = $this->city->getParents();
+
+        return $toMerge ? array_values(array_unique(array_merge(...$toMerge))) : [];
```
kevin-verschaeve, line 84: I think this condition is useless as it will always pass because, at least the array will look like `[ [ ], [ ] ]` at minimum

renanbr: done via a4f9789be926999b2e62e3e5b146878f7ba00edc

ottaviano (approved):

kevin-verschaeve (approved):

PR #4909, opened 2020-10-12
add a slug on thematic community

kevin-verschaeve:


migrations/Version20201012163616.php
```diff
@@ -0,0 +1,24 @@
...
+final class Version20201012163616 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE 
+          thematic_community 
+        ADD 
+          canonical_name VARCHAR(255) NOT NULL, 
+        ADD 
+          slug VARCHAR(255) NOT NULL');
```
ottaviano, line 17: do we need to fill these fields for existing thematics?

ottaviano (approved):

migrations/Version20201012163616.php
kevin-verschaeve: yes but, no communities exists yet

PR #4914, opened 2020-10-13
Add FDE level as an entity in the new geo mapping

renanbr:
The "Abroad" level is the parent of "Foreign Districts", therefore "Consular Districts" and "Countries".

This change allows us to group these levels into the FDE one as "Zones" entities.

### DIFF
--- a/migrations/Version20201013145634.php
+++ b/migrations/Version20201013145634.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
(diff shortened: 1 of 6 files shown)


renanbr:
Here is where we create the `FDE` entity: https://github.com/EnMarche/en-marche.fr/pull/4914/files#diff-12dcea49cb321fb296038e406c80502697cf13dc44be52cf0d4ce940830b34a1R24

kevin-verschaeve (approved):

ottaviano (approved):

PR #4919, opened 2020-10-13
Fix matching for OCR responses

Remg:


kevin-verschaeve (approved):

src/Vision/ImageAnnotations.php
```diff
@@ -69,12 +69,12 @@ public function setText(?string $text): void
 
     public function isIdentityDocument(): bool
     {
-        return 0 < array_intersect($this->labels, [
+        return 0 < \count(array_intersect($this->webEntities, [
```
ottaviano, line 72: is it normal that the method `isIdentityDocument` uses `$this->webEntities` and `isFrenchPassport` uses `$this->labels`

ottaviano (approved):

src/Vision/ImageAnnotations.php
Remg: yes it is ! 
i look at both of these arrays to make sure its a document

PR #4921, opened 2020-10-13
[Admin] Change adherent list page: add filters + fix column display

illusionOfParadise:


illusionOfParadise:
[ch12556]

src/Entity/Adherent.php
```diff
@@ -1721,6 +1721,22 @@ public function getMemberships(): CommitteeMembershipCollection
         return $this->memberships;
     }
 
+    public function hasVotingCommitteeMembership(): bool
+    {
+        return null !== $this->getVotingCommitteeMembership();
```
ottaviano, line 1726: you can use the existing method: 
```php
return null !== $this->memberships->getVotingCommitteeMembership()
```

ottaviano (approved):

PR #4922, opened 2020-10-14
Fix getParents() in Country

renanbr:
The foreign district is not always present, so `getParents()` is broken.

This PR fixes it.

### DIFF
--- a/src/Entity/Geo/Country.php
+++ b/src/Entity/Geo/Country.php
@@ -39,12 +39,14 @@ class Country implements ZoneableInterface
 
     public function getParents(): array
     {
-        $toMerge = [
-            [$this->foreignDistrict],
-            $this->foreignDistrict->getParents(),
-        ];
+        $toMerge = [];
 
-        return array_values(array_unique(array_merge(...$toMerge)));
+        if ($this->foreignDistrict) {
+            $toMerge[] = [$this->foreignDistrict];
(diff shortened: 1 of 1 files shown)


kevin-verschaeve (approved):

src/Entity/Geo/Country.php
```diff
@@ -39,12 +39,14 @@ public function setForeignDistrict(?ForeignDistrict $foreignDistrict): void
...
-        ];
+        $toMerge = [];
 
-        return array_values(array_unique(array_merge(...$toMerge)));
+        if ($this->foreignDistrict) {
+            $toMerge[] = [$this->foreignDistrict];
+            $toMerge[] = $this->foreignDistrict->getParents();
+        }
+
+        return $toMerge ? array_values(array_unique(array_merge(...$toMerge))) : [];
```
ottaviano, line 49: as an alternative:
```php
if (!$this->foreignDistrict) {
    return [];
}

return array_values(array_unique(array_merge(
    [$this->foreignDistrict],
    $this->foreignDistrict->getParents()
)));
```

ottaviano (approved):

src/Entity/Geo/Country.php
renanbr: Thanks for the feedback.
I'll keep the filling strategy to be coherent with other implementations.

PR #4927, opened 2020-10-15
Introduce geo zone into mandate entity

renanbr:
Geo zones will migrate to a new schema.

This PR allows us to keep both relationships in the `Mandate` entity.

- `zone` already exists and will remain until the complete migration
- `geoZone` points to the new mapping

Once this PR is merged:

1. we can populate the new field with the new corresponding id, then
2. make a new PR to turn the new field as `NOT NULL`

### DIFF
--- a/src/Entity/ElectedRepresentative/Mandate.php
+++ b/src/Entity/ElectedRepresentative/Mandate.php
@@ -2,6 +2,7 @@
 
 namespace App\Entity\ElectedRepresentative;
 
+use App\Entity\Geo\Zone as GeoZone;
 use App\Exception\BadMandateTypeException;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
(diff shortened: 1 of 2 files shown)


src/Entity/ElectedRepresentative/Mandate.php
```diff
@@ -49,9 +50,20 @@ class Mandate
...
+     * @deprecated Will be replace by $geoZone
      */
     private $zone;
 
+    /**
+     * @var GeoZone|null
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Geo\Zone")
+     *
+     * @Assert\NotBlank
```
ottaviano, line 63: it will block updates on admin?

ottaviano (approved):

src/Entity/ElectedRepresentative/Mandate.php
renanbr: Good question, I don't know how the admin form is configured.
If the form gathers all fields automatically, yes we have a problem.
If this new field isn't mapped, I think it may pass.

To prevent any problems I'll remove this assert, and place this in the subsequent PRs

renanbr: done via 927a1f351d716effd5646ae478fe3f2d9b0789ab

kevin-verschaeve (approved):

PR #4936, opened 2020-10-19
Add a banner field on CT and move motivations to an array

kevin-verschaeve:


src/ThematicCommunity/ThematicCommunityMembershipFilter.php
```diff
@@ -362,7 +362,7 @@ public function toArray(): array
             }, $this->thematicCommunities),
             'joinedSince' => $this->joinedSince ? $this->joinedSince->format('Y-m-d') : null,
             'joinedUntil' => $this->joinedUntil ? $this->joinedUntil->format('Y-m-d') : null,
-            'motivation' => $this->motivation,
+            'motivation' => $this->motivations,
```
ottaviano, line 365: I think it should be `'motivations'`?

ottaviano (approved):

PR #4948, opened 2020-10-22
Update relationship between zone and referent tag

renanbr:
Change the relation between  Zone and ReferentTag from `ManyToOne` to unidirectional `OneToOne`.

Changes:
- Remove ReferentTag reference in Zone entity
- Create a `OneToOne` reference to Zone in ReferentTag entity
    - Zone entity is intentionally not aware of its relationship with ReferentTag
- Mark ReferentTag as its repository as deprecated

As consequence, when migrating from ReferentTag to Zone there will be at least one more query to execute, e.g. `TerritorialCouncilRepository::findByMandates()`

---

Note that we need to fill up the new `referent_tags.zone_id` field once it gets deployed.

### DIFF
--- a/src/Entity/ReferentTag.php
+++ b/src/Entity/ReferentTag.php
@@ -2,6 +2,7 @@
 
 namespace App\Entity;
 
+use App\Entity\Geo\Zone;
 use Doctrine\ORM\Mapping as ORM;
 use JMS\Serializer\Annotation as JMSSerializer;
 use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
@@ -20,6 +21,8 @@ use Symfony\Component\Validator\Constraints as Assert;
  *
  * @UniqueEntity("name")
  * @UniqueEntity("code")
+ *
+ * @deprecated
  */
 class ReferentTag
 {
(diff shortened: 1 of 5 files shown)


src/Entity/ReferentTag.php
```diff
@@ -165,4 +175,9 @@ public function getDepartmentCodeFromCirconscriptionName(): ?string
     {
         return $this->isDistrictTag() ? \substr($this->code, 6, 2) : null;
     }
+
+    public function getZone(): Zone
```
ottaviano, line 179: do we need to allow `null`?
`: ?Zone`

src/Repository/TerritorialCouncil/TerritorialCouncilRepository.php
```diff
@@ -56,16 +56,24 @@ public function findAllWithoutStartedElection(Designation $designation, int $off
...
+                $zones[] = $mandate->getGeoZone();
             }
         });
 
+        $tagsFromZone = $this->getEntityManager()
+            ->getRepository(ReferentTag::class)
+            ->findByZones($zones)
+        ;
+
+        $tags = array_merge($tags, $tagsFromZone);
```
ottaviano, line 75: what do you think about modifying the query condition instead: 
```php
if ($zones) {
  $qb->leftJoin('tag.zone', 'zone')
    ->orWhere('zone IN (:zones)');
}
```

src/Entity/ReferentTag.php
renanbr: the idea is to make sure there are no orphaned `ReferentTags`, so they *MUST* point to some `Zone`

ottaviano (approved):

PR #4949, opened 2020-10-23
[CoPol][OfficialReport] Not allow to create an official report if no president + change the download link

illusionOfParadise:


kevin-verschaeve (approved):

src/Controller/EnMarche/TerritorialCouncil/OfficialReportManagerController.php
```diff
@@ -44,6 +44,13 @@ public function createAction(Request $request, OfficialReportManager $manager):
         ;
 
         if ($form->isSubmitted() && $form->isValid()) {
+            if (!$report->getAuthor()
```
ottaviano, line 47: why do not move it on the top of the action to not allow to fill the form as it's not possible to submit it?

src/Controller/EnMarche/TerritorialCouncil/PoliticalCommitteeController.php
```diff
@@ -116,7 +116,7 @@ public function listOfficialReportsAction(
     }
 
     /**
-     * @Route("/proces-verbaux/{uuid}/telecharger", name="official_report_download", methods={"GET"})
+     * @Route("/proces-verbaux/{uuid}", name="official_report_download", methods={"GET"})
```
ottaviano, line 119: do we need to change the page title too?

src/Controller/EnMarche/TerritorialCouncil/OfficialReportManagerController.php
illusionOfParadise: We cannot, because we have no necessary information to make a check like political committee, that is accessible only after submitting a form, because its value can be chose by in the form.

src/Controller/EnMarche/TerritorialCouncil/PoliticalCommitteeController.php
illusionOfParadise: Which title?

src/Controller/EnMarche/TerritorialCouncil/OfficialReportManagerController.php
ottaviano: oh my bad 👍

ottaviano (approved):

PR #4957, opened 2020-10-26
[Thematic Community] Various small fixes

kevin-verschaeve:


templates/admin/thematic_community/member/community.html.twig
```diff
@@ -1,21 +1,22 @@
...
-        &nbsp;
-    {% endfor %}
 
-    {% if object.expert %}
-        <br>
-        <span class="label label-danger">
-        Expert
-    </span>
-    {% endif %}
+        {% for motivation in object.motivations|filter(m => m is not empty) %}
```
ottaviano, line 9: is it normal that the array motivations can have an empty element? maybe we should improve setter and not twig?

ottaviano (approved):

templates/admin/thematic_community/member/community.html.twig
kevin-verschaeve: normally, it's not. But from the import, it may be empty (that's how I found this actually)

ottaviano: so we can improve the import command to not allow to pass an empty element?

kevin-verschaeve: yes, will do that

kevin-verschaeve: fixed

PR #4968, opened 2020-10-28
Add geo zone to adherents

illusionOfParadise:
SQL to execute:
```sql
INSERT INTO adherent_zone (adherent_id, zone_id)
SELECT a.id, z.id FROM adherents a, geo_zone z
WHERE a.address_country = 'FR' AND z.code = SUBSTRING(a.address_city_insee, 7, 5)
;

INSERT INTO adherent_zone (adherent_id, zone_id)
SELECT a.id, z.id FROM adherents a, geo_zone z
WHERE a.address_country != 'FR' AND z.code = a.address_country
;
```

### DIFF
--- a/src/Membership/MembershipRequestHandler.php
+++ b/src/Membership/MembershipRequestHandler.php
@@ -16,6 +16,7 @@ use App\Mailer\Message\AdherentMembershipReminderMessage;
 use App\Mailer\Message\AdherentTerminateMembershipMessage;
 use App\OAuth\CallbackManager;
 use App\Referent\ReferentTagManager;
+use App\Referent\ReferentZoneManager;
 use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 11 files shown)


illusionOfParadise:
[ch12609]

src/Membership/MembershipRequestHandler.php
```diff
@@ -52,6 +55,7 @@ public function __construct(
         $this->mailer = $mailer;
         $this->manager = $manager;
         $this->referentTagManager = $referentTagManager;
+        $this->referentZoneManager = $referentZoneManager;
```
ottaviano, line 58: what about creating a listener for `UserEvents::USER_UPDATED` event instead and moving this logic into?

illusionOfParadise: As discussed in PM, for now we left it like it is.

ottaviano (approved):

PR #4973, opened 2020-10-30
[Admin] Add new candidate roles

illusionOfParadise:


src/Form/Admin/CandidateManagedAreaType.php
```diff
@@ -0,0 +1,49 @@
...
+class CandidateManagedAreaType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('zone', EntityType::class, [
+                'label' => false,
+                'required' => false,
+                'class' => Zone::class,
+                'query_builder' => function (ZoneRepository $repository) {
```
ottaviano, line 24: it will load 2K entities for filling the <select>, maybe we can use an autocomplete?

ottaviano (approved):

PR #4978, opened 2020-11-02
Dimitri/ch12673/programmer les emails transacs

ottaviano:


ottaviano:
[ch12637]

src/Security/Voter/Committee/CommitteeCandidacyVoter.php
```diff
@@ -3,28 +3,65 @@
...
+            if (
+                $adherent->isReferent()
+                || $adherent->isSupervisor()
+                || $adherent->isDeputy()
+                || $adherent->isSenator()
+            ) {
+                return false;
+            }
+
+            return true;
```
renanbr, line 33:
```suggestion
            return
                !$adherent->isReferent()
                && !$adherent->isSupervisor()
                && !$adherent->isDeputy()
                && !$adherent->isSenator()
            ;
```

renanbr (approved):

illusionOfParadise (approved):

src/Security/Voter/Committee/CommitteeCandidacyVoter.php
ottaviano: yes it's an alternative, but I found more readable with IF and without NOT operator

PR #4982, opened 2020-11-03
Create ZoneChoiceType

renanbr:


illusionOfParadise (approved):

src/Form/ZoneChoiceType.php
```diff
@@ -0,0 +1,60 @@
...
+class ZoneChoiceType extends AbstractType
+{
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
+
+    public function __construct(Security $security, SessionInterface $session, TranslatorInterface $translator)
+    {
+        parent::__construct($security, $session);
```
illusionOfParadise, line 24: There is no constructor in the parent class.

PR #4980, opened 2020-11-02
Filter elected via geo zone in "espace référent" and LRE

renanbr:
Depends on:

- [x] #4962

PR #4988, opened 2020-11-04
Use new zones in backoffice elected area

renanbr:
Depends on:

- [ ] #4980


src/Entity/Geo/Zone.php
```diff
@@ -42,6 +43,45 @@ class Zone implements GeoInterface
         self::REGION,
     ];
 
+    public const TYPES_BY_MANDATE = [
```
ottaviano, line 46: what about moving this into Mandate class instead of keeping it in generic Zone class?

ottaviano (approved):

src/Entity/Geo/Zone.php
renanbr: done 60510da008b453f73c17bb7c3b18e8577d48e4e8

PR #5005, opened 2020-11-10
[CommitteeMerge] move committee candidacies on merge

ottaviano:


src/Committee/CommitteeMergeCommandHandler.php
```diff
@@ -201,4 +208,32 @@ private function createCommitteeMergeHistory(
     ): CommitteeMergeHistory {
         return new CommitteeMergeHistory($sourceCommittee, $destinationCommittee, $mergedMemberships, $administrator);
     }
+
+    private function transferCandidatures(?Committee $sourceCommittee, ?Committee $destinationCommittee)
+    {
+        $sourceElection = $sourceCommittee->getCurrentElection();
+        $destinationElection = $destinationCommittee->getCurrentElection();
+
+        // Move all candidacy from the source committee to new committee if both have the same election type
```
illusionOfParadise, line 217:
```suggestion
        // Move all candidacies from the source committee to new committee, if both have the same election type
```

illusionOfParadise (approved):

src/Committee/CommitteeMergeCommandHandler.php
ottaviano: 👍

PR #5094, opened 2020-11-24
Validate zone before filtering users

renanbr:
This PR fix also a bug in the messaging area

https://github.com/EnMarche/en-marche.fr/pull/5094/files#diff-a229ac7f462f745c27baf5e382c75805deb1f33e94fd5bc8941634d55a1dd218

### DIFF
--- a/src/Validator/ManagedZone.php
+++ b/src/Validator/ManagedZone.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace App\Validator;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ * @Target({"PROPERTY", "METHOD", "ANNOTATION"})
+ */
+class ManagedZone extends Constraint
+{
+    public $message = "Oups, vous n'avez pas accès à cette zone !";
+
+    /**
+     * @var string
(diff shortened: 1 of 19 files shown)


src/Validator/ManagedZone.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @var string
+     */
+    public $spaceType;
+
+    /**
+     * @var Adherent|null
+     */
+    public $user;
```
ottaviano, line 24: could you remove it?

src/Validator/ManagedZoneValidator.php
```diff
@@ -0,0 +1,110 @@
...
+     * @var ManagedZoneProvider
+     */
+    private $managedZoneProvider;
+
+    /**
+     * @var TokenStorageInterface
+     */
+    private $tokenStorage;
+
+    public function __construct(ManagedZoneProvider $managedZoneProvider, TokenStorageInterface $tokenStorage)
```
ottaviano, line 25: waht about using Security service instead of TokenStorageInterface?

ottaviano (approved):

PR #5122, opened 2020-11-26
Add Department endpoint - Remove Region endpoint

Remg:


src/Entity/Jecoute/Region.php
```diff
@@ -225,10 +129,21 @@ class Region
      */
     private $logoFile;
 
+    /**
+     * @var GeoRegion|null
+     *
+     * @ApiProperty(identifier=true)
```
ottaviano, line 135: could you remove it please?

ottaviano (approved):

PR #5128, opened 2020-11-28
On the road to sf 4.4

ottaviano:


ottaviano:
[ch12734]

mlabarrere:
\ ^--^ /

src/AdherentProfile/AdherentProfile.php
```diff
@@ -39,6 +39,7 @@ class AdherentProfile implements MembershipInterface
      * @Assert\Length(
      *     min=2,
      *     max=50,
+     *     allowEmptyString=true,
```
illusionOfParadise, line 42: Why add this, if the field has `@Assert\NotBlank`?

src/AdherentProfile/AdherentProfile.php
```diff
@@ -52,6 +53,7 @@ class AdherentProfile implements MembershipInterface
...
```
illusionOfParadise, line 56: Why add this, if the field has `@Assert\NotBlank`?

illusionOfParadise (approved):
🥇

src/AdherentProfile/AdherentProfile.php
ottaviano: image
because it creates another error in addition of `NotBlank`

renanbr (approved):

PR #5152, opened 2020-12-02
Link geo entities to geo_data

renanbr:


src/Entity/Geo/GeoTrait.php
```diff
@@ -36,6 +37,14 @@ trait GeoTrait
      */
     private $name;
 
+    /**
+     * @var GeoData|null
+     *
+     * @ORM\OneToOne(targetEntity="App\Entity\GeoData")
+     * @ORM\JoinColumn(nullable=true)
```
ottaviano, line 44: we can avoid this annotation as this is nullable by default

src/Repository/Geo/ZoneRepository.php
```diff
@@ -50,6 +50,7 @@ public function zoneableAsZone(ZoneableInterface $zoneable): Zone
 
         $zone->activate($zoneable->isActive());
         $zone->setName($zoneable->getName());
+        $zone->setGeoData($zoneable->getGeoData());
```
ottaviano, line 53: It seems that `ZoneableInterface` has not `getGeoData`, same for others methods getName, getCode, isActive 🤔

renanbr: `ZoneableInterface` extends `GeoInterface`, thus they are part of the contract

ottaviano: oh yes, my bad, sorry

ottaviano (approved):

ottaviano (approved):

PR #5157, opened 2020-12-04
[Jecoute][Candidate] Add Jecoute to candidate space + rename "J'écoute" to "Questionnaires"

illusionOfParadise:


illusionOfParadise:
[ch12858]

src/Controller/EnMarche/Jecoute/AbstractJecouteController.php
```diff
@@ -69,6 +70,11 @@ public function jecouteSurveyCreateAction(
         SuggestedQuestionRepository $suggestedQuestionRepository,
         UserInterface $user
     ): Response {
+        if (JecouteSpaceEnum::CANDIDATE_SPACE === $this->getSpaceName()
```
ottaviano, line 73: what do you think about creating a function `checkCreateAccess` and use it in this generic createAction so we can override it in `JecouteCandidateController` for doing this specific job?
```
$this->checkCreateAccess();
```

templates/jecoute/surveys_list.html.twig
```diff
@@ -55,14 +58,17 @@
...
                         {% set row = {
                             id: survey.id,
                             name: survey.name,
                             createdAt: survey.createdAt|date('d/m/Y'),
                             questionsCount: data.questions_count,
                             responsesCount: data.responses_count,
-                            zone: survey is instanceof('App\\Entity\\Jecoute\\LocalSurvey') ? survey.zone|trans : null,
+                            zone: survey is instanceof('App\\Entity\\Jecoute\\LocalSurvey')
+                                ? '<span class="adherent-dashboard-tag">'~('geo_zone.'~zone.type)|trans~'</span><br>'
+                                    ~"<div class=\"b__nudge--top-10\">#{zone.name} (#{zone.code})</span>",
```
ottaviano, line 71: what will happen if the zone variable is null? is this possible?

ottaviano (approved):

templates/jecoute/surveys_list.html.twig
illusionOfParadise: For `LocalSurvey`, no.

ottaviano: so this template displays only LocalSurvey? I ask for this code: `#{zone.name} (#{zone.code})` which does not work if zone is null

src/Controller/EnMarche/Jecoute/AbstractJecouteController.php
ottaviano: @illusionOfParadise, could you override the `checkCreateAccess` in `JecouteCandidateController`, please?

PR #5170, opened 2020-12-08
[Jecoute] Some fixes and wording changes

illusionOfParadise:


templates/jecoute/surveys_list.html.twig
```diff
@@ -76,7 +76,7 @@
                                 '<span class="status status__2">Non publié</span>',
                         } %}
 
-                        {% if is_granted('CAN_EDIT_SURVEY', survey) %}
+                        {% if is_granted('IS_AUTHOR_OF', survey) or is_granted('CAN_EDIT_SURVEY', survey) %}
```
ottaviano, line 79: what about let `CAN_EDIT_SURVEY` voter to do the check if the current user is the author of Survey?

ottaviano (approved):

PR #5209, opened 2020-12-17
Link events to zones

renanbr:

PR #5222, opened 2020-12-21
Add zone filters in admin

renanbr:


renanbr:
Depends on #5209 

The commit concerned by this PR: eda461e87cd114d19c28a7faa960f865b94e81de

src/Admin/AdherentAdmin.php
```diff
@@ -578,6 +579,48 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                 'label' => 'Dernière connexion',
                 'field_type' => DateRangePickerType::class,
             ])
+            ->add('zones', CallbackFilter::class, [
```
illusionOfParadise, line 582: Seems that this code is repeated in multiple admin classes.
What about create a zone filter type class and use it?

renanbr: done in c4455466ca64271af9d3c18e4c98d4c9ae97c73f

illusionOfParadise (approved):

PR #5314, opened 2021-02-01
[Admin][Committee] Close a committee while merge

illusionOfParadise:


src/Entity/Committee.php
```diff
@@ -568,6 +579,20 @@ public function getSupervisors(bool $isProvisional = null): array
         }, $this->findSupervisorMandates(null, $isProvisional)->toArray());
     }
 
+    /**
+     * Marks this committee as closed.
+     */
+    public function close(string $timestamp = 'now')
```
ottaviano, line 585: 1/ `: void`
2/ could you use `\DateTime` as argument, please?

ottaviano (approved):

PR #5350, opened 2021-02-17
[Coalition] Causes

Remg:


src/Entity/Coalition/Cause.php
```diff
@@ -0,0 +1,202 @@
...
+     */
+    private $coalition;
+
+    /**
+     * @var Adherent|null
+     *
+     * @ORM\ManyToOne(targetEntity="App\Entity\Adherent")
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private $author;
```
ottaviano, line 100: what about using `AuthoredTrait` + `AuthoredInterface` here?

src/Entity/Coalition/Cause.php
```diff
@@ -0,0 +1,202 @@
...
+    protected $image;
+
+    /**
+     * @var string|null
+     *
+     * @ORM\Column
+     *
+     * @SymfonySerializer\Groups({"cause_read"})
+     */
+    private $name;
```
ottaviano, line 75: should we use `EntityNameSlugTrait` to be able to use `slug` field for the URL?

Remg: we will handle slugs in another PR

illusionOfParadise (approved):

ottaviano (approved):

illusionOfParadise (approved):

PR #5386, opened 2021-02-25
[Committee] Remove canary mode for phone number in committee for AL + show phone of all members

illusionOfParadise:


templates/committee_manager/list_members.html.twig
```diff
@@ -182,13 +182,11 @@
...
+                        {% if is_supervisor %}
                             <td class="member-phone">
-                                {% if membership.isVotingCommittee %}
-                                    <a href="tel:{{ member.phone|phone_number_format }}">
-                                        {{ member.phone|phone_number_format }}
-                                    </a>
-                                {% endif %}
+                                <a href="tel:{{ member.phone|phone_number_format }}">
+                                    {{ member.phone|phone_number_format }}
+                                </a>
```
ottaviano, line 189: could you add a check if member.phone is not empty, please? otherwise it will crash

ottaviano (approved):

PR #5544, opened 2021-04-09
[API] add a proxy endpoint for internal API application

elvino29:


src/DataFixtures/ORM/LoadInternalApiApplicationData.php
```diff
@@ -0,0 +1,43 @@
...
+            'application-3' => new InternalApiApplication(
+                'app-test',
+                'http://test.enmarche.code',
+                Uuid::fromString(self::INTERNAL_API_APPLICATION_03_UUID)
+            ),
+        ];
+
+        foreach ($data as $key => $application) {
+            $manager->persist($application);
+            $this->addReference($key, $application);
```
ottaviano, line 38: I think we can avoid to call `addReference` as we don't use it

src/Entity/InternalApiApplication.php
```diff
@@ -0,0 +1,67 @@
...
+     * @Assert\Length(max=200, maxMessage="internal_api_application.validation.application_name_length")
+     */
+    private $applicationName;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(length=200)
+     *
+     * @Assert\NotBlank
```
ottaviano, line 37: could you add `@Assert\Url` please?

ottaviano (approved):

PR #5627, opened 2021-04-26
[Coalitions] Add causes export

illusionOfParadise:


src/Entity/Geo/City.php
```diff
@@ -84,6 +84,11 @@ public function getPostalCode(): array
         return $this->postalCode ?: [];
     }
 
+    public function getPostalCodeAsString(): ?string
```
ottaviano, line 87: `getPostalCodeAsString(): string`

src/Entity/Geo/City.php
```diff
@@ -84,6 +84,11 @@ public function getPostalCode(): array
...
+    {
+        return implode(', ', $this->postalCode);
```
ottaviano, line 89: what happens if postalCode is null and not an array?

src/Exporter/CausesExporter.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    private $coalitionsHost;
+    private $exporter;
+    private $repository;
+    private $cityRepository;
+    private $urlGenerator;
+    private $translator;
+
+    public function __construct(
+        string $coalitionsHost,
```
ottaviano, line 27: could you use `CoalitionUrlGenerator` instead please?

PR #5674, opened 2021-04-30
[EM-736] Add endpoint to terminate membership

Remg:


src/Repository/OAuth/RefreshTokenRepository.php
```diff
@@ -46,12 +47,39 @@ public function findAllRefreshTokensByClient(Client $client): array
...
 
+    /**
+     * @return RefreshToken[]
+     */
+    public function findAllRefreshTokensByUser(Adherent $user): array
+    {
+        return $this
+            ->createQueryBuilder('rt')
+            ->join('rt.accessToken', 'at')
+            ->where('at.user = :user')
```
ottaviano, line 58: what about adding this check `AND rt.revokedAt IS NULL` to avoid to take already revoked tokens? same for accessToken query

src/Repository/OAuth/RefreshTokenRepository.php
```diff
@@ -46,12 +47,39 @@ public function findAllRefreshTokensByClient(Client $client): array
...
+            $this->revokeToken($refreshToken);
+        }
+    }
+
+    public function revokeUserTokens(Adherent $user): void
+    {
+        foreach ($this->findAllRefreshTokensByUser($user) as $refreshToken) {
+            $this->revokeToken($refreshToken);
+        }
+    }
```
ottaviano, line 77: `flush()` is missing?

PR #5693, opened 2021-05-05
[CoalitionsMessage] allow update message filter via API

ottaviano:


src/AdherentMessage/Filter/FilterFactory.php
```diff
@@ -142,4 +145,13 @@ private static function createCandidateJecouteFilter(Adherent $user): JecouteFil
 
         return new JecouteFilter($user->getCandidateManagedArea()->getZone());
     }
+
+    private static function createCoalitionsFilter(Adherent $user): CoalitionsFilter
+    {
+        if (!$user->isApprovedCauseAuthor()) {
+            throw new \InvalidArgumentException('[AdherentMessage] Adherent should be a cause author');
```
illusionOfParadise, line 152: Maybe?
```suggestion
            throw new \InvalidArgumentException('[AdherentMessage] Adherent should be an approved cause author');
```

illusionOfParadise (approved):

PR #5783, opened 2021-05-31
[API][Events] Get events where user has a registration

illusionOfParadise:


src/Controller/Api/EventsController.php
```diff
@@ -86,4 +89,24 @@ public function eventsCountInReferentManagedArea(
             'participants_as_adherent' => $eventRegistrationRepository->countEventParticipantsAsAdherentInReferentManagedArea($referent),
         ]);
     }
+
+    /**
+     * @Route("/v3/events/with-registration", name="api_events_with_registration", methods={"POST"})
```
ottaviano, line 94: we already have the same route for cause: 
```php
@Route("/v3/causes/followed", name="api_causes_followed", methods={"POST"})
```
what about keep the same forme for URL and route name?

ottaviano (approved):

PR #5801, opened 2021-06-02
[Event/API] add event anonymous subscribe endpoint

ottaviano:


tests/Controller/Api/Event/SubscribeAsAnonymousControllerTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $this->assertStatusCode(400, $this->client);
+
+        $this->client->request(Request::METHOD_POST, sprintf('/api/events/%s/subscribe', LoadCoalitionEventData::EVENT_1_UUID), [], [], [], json_encode([
+            'first_name' => 'Joe',
+            'last_name' => 'Hey',
+            'email_address' => 'j.hey@en-marche-dev.fr',
+        ]));
+
+        $this->isSuccessful($this->client->getResponse());
```
illusionOfParadise, line 34: Could you also check that all information has been saved in the registration?

illusionOfParadise (approved):

tests/Controller/Api/Event/SubscribeAsAnonymousControllerTest.php
ottaviano: done, good point 👍

PR #5812, opened 2021-06-04
[EM-963] Add endpoints to track push token with adherents & devices

Remg:


src/Normalizer/PushTokenDenormalizer.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function supportsDenormalization($data, $type, $format = null, array $context = [])
+    {
+        // Make sure we're not called twice
+        if (isset($context[self::ALREADY_CALLED])) {
+            return false;
+        }
+
+        return is_a($type, PushToken::class, true)
+            && ($this->isAdherentUser() || $this->isDeviceUser())
```
ottaviano, line 54: what about checking just `return ... && $this->security->getUser();` here?
will it be enough?

Remg: yes. validation will fail afterwards if it's another type :ok_hand:

ottaviano (approved):

PR #5927, opened 2021-06-18
[API][Events] Coalition events route returns Coalition and Cause events

illusionOfParadise:


src/Controller/Api/Coalition/CoalitionEventsController.php
```diff
@@ -0,0 +1,38 @@
...
+        BaseEventRepository $baseEventRepository,
+        SerializerInterface $serializer
+    ): JsonResponse {
+        $events = $baseEventRepository->findEventsForCoalition(
+            $coalition,
+            $request->query->getInt('page', 1),
+            $request->query->getInt('page_size', 30)
+        );
+
+        return new JsonResponse(
```
ottaviano, line 31: what about using Sf helper: `$this->json()`?

ottaviano (approved):

PR #6223, opened 2021-09-17
[API][Riposte] Add stats fields

illusionOfParadise:


src/Controller/Api/Jecoute/IncrementRiposteStatsCounterController.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+class IncrementRiposteStatsCounterController extends AbstractController
+{
+    public function __invoke(Riposte $riposte, string $action, EntityManagerInterface $entityManager): JsonResponse
+    {
+        if (!\in_array($action, Riposte::ACTIONS)) {
+            return $this->json([
```
ottaviano, line 16: we can use `default` statement of switch case for doing this return too

ottaviano (approved):

Remg (approved):

PR #6268, opened 2021-09-29
Introduce CMS Blocks

Remg:


src/Admin/CmsBlockAdmin.php
```diff
@@ -0,0 +1,79 @@
...
+
+    protected function configureListFields(ListMapper $listMapper)
+    {
+        $listMapper
+            ->addIdentifier('name', null, [
+                'label' => 'Nom',
+            ])
+            ->add('description', null, [
+                'label' => 'Description',
+            ])
```
ottaviano, line 66: what about adding creator info to the list?

src/CmsBlock/CmsBlockManager.php
```diff
@@ -0,0 +1,20 @@
...
+    private $cmsBlockRepository;
+
+    public function __construct(CmsBlockRepository $cmsBlockRepository)
+    {
+        $this->cmsBlockRepository = $cmsBlockRepository;
+    }
+
+    public function getContent(string $cmsBlockName): ?string
+    {
+        return $this->cmsBlockRepository->getContentByName($cmsBlockName);
```
ottaviano, line 18: could you add `$logger->error()` when block is null?

ottaviano (approved):

PR #6523, opened 2021-11-15
[EM-1246] add a get endpoint for campaign histories list with some fi…

elvino29:
…lters


src/Entity/Phoning/CampaignHistory.php
```diff
@@ -44,6 +59,14 @@
  *         },
  *     },
  * )
+ *
+ * @ApiFilter(SearchFilter::class, properties={
+ *     "campaign.title": "partial",
+ *     "status": "exact",
+ * })
+ * @ApiFilter(PhoningCampaignCallerFilter::class)
```
ottaviano, line 67: Two filters seem to be similar, what about doing something like that:
```suggestion
 * @ApiFilter(AdherentIdentityFilter::class, properties={
 *     "adherent", 
 *.    "caller",
 * })
```

ottaviano (approved):

PR #6592, opened 2021-11-26
[EM-1671] add author postal code on data survey entity

elvino29:


src/Entity/Jecoute/DataSurvey.php
```diff
@@ -28,6 +28,11 @@ class DataSurvey implements AuthorInterface
      */
     private $author;
 
+    /**
+     * @ORM\Column(type="text", nullable=true)
```
ottaviano, line 32:
```suggestion
     * @ORM\Column(nullable=true)
```

ottaviano (approved):

PR #6607, opened 2021-12-01
[API][PAP] Change get campaign endpoint for DC

illusionOfParadise:
[EM-1730]

[EM-1730]: https://em-tech-innovation.atlassian.net/browse/EM-1730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Entity/Pap/Campaign.php
+++ b/src/Entity/Pap/Campaign.php
@@ -44,6 +44,12 @@ use Symfony\Component\Validator\Constraints as Assert;
  *             "method": "GET",
  *             "path": "/v3/pap_campaigns",
  *         },
+ *         "get_paginated": {
+ *             "method": "GET",
+ *             "path": "/v3/pap_campaigns/paginated",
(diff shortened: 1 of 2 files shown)


src/Entity/Pap/Campaign.php
```diff
@@ -25,6 +25,7 @@
  *     attributes={
  *         "order": {"createdAt": "DESC"},
  *         "pagination_enabled": false,
```
elvino29, line 27: what do you thing about removing `"pagination_enable": false` ?
Api platform will handle the pagination for collection operation. so need to use `"pagination_client_enabled": true`

illusionOfParadise: Done.

ottaviano (approved):
💯

PR #6620, opened 2021-12-02
Add EntityFromUuid denormalizer

ottaviano:


src/Normalizer/EntityFromUuidDenormalizer.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->entityManager = $entityManager;
+    }
+
+    public function denormalize($data, $type, $format = null, array $context = [])
+    {
+        if ($object = $this->entityManager->getRepository($type)->findOneBy(['uuid' => $data])) {
+            return $object;
+        }
+
+        throw new ItemNotFoundException();
```
elvino29, line 25: Can you add an exception message. Then, we will have more details when the exception is thrown

elvino29 (approved):

src/Normalizer/EntityFromUuidDenormalizer.php
ottaviano: fixed, thanks

PR #6631, opened 2021-12-06
[API][PAP] Add doors and survey number to floor info

illusionOfParadise:


src/Entity/Pap/FloorStatistics.php
```diff
@@ -30,6 +30,16 @@ class FloorStatistics implements CampaignStatisticsInterface
      */
     private Campaign $campaign;
 
+    /**
+     * @ORM\Column(type="simple_array", nullable=true)
+     */
+    private array $doors = [];
```
ottaviano, line 36:
```suggestion
    private array $visitedDoors = [];
```

ottaviano (approved):

PR #6696, opened 2021-12-24
[API][Events] Send email after modification or cancellation of a default event

illusionOfParadise:


src/Event/EventMessageNotifier.php
```diff
@@ -74,7 +74,7 @@ public function onEventCreated(EventEvent $event): void
 
     public function onEventCancelled(EventEvent $event): void
     {
-        if (!$event->isCoalitionsEvent() && (!$event instanceof CommitteeEventEvent || !$event->getCommittee())) {
+        if ((!$event instanceof CommitteeEventEvent || !$event->getCommittee()) && !$event->needSendMessage()) {
```
ottaviano, line 77: it seems that `needNotifyForCancellation` is not used, maybe you should use it here?

ottaviano (approved):

PR #6733, opened 2022-01-10
[API][PAP] Add address stats

illusionOfParadise:


src/Normalizer/Pap/AppendCampaignStatsNormalizer.php
```diff
@@ -43,6 +47,8 @@ public function normalize($object, $format = null, array $context = [])
 
         $campaign['nb_surveys'] = $object->getCampaignHistoriesWithDataSurvey()->count();
         $campaign['nb_visited_doors'] = $this->campaignHistoryRepository->countVisitedDoors($object);
+        $campaign['nb_addresses'] = $object->getNbAddresses();
```
ottaviano, line 50: we can use `Group("pap_campaign_read")` on getter for do the same job

src/Normalizer/Pap/AppendCampaignStatsNormalizer.php
```diff
@@ -43,6 +47,8 @@ public function normalize($object, $format = null, array $context = [])
...
+        $campaign['nb_voters'] = $object->getNbVoters();
```
elvino29, line 51: you can add the serialization group directly in the Campaign entity and remove those two line in the normalizer

ottaviano (approved):

src/Normalizer/Pap/AppendCampaignStatsNormalizer.php
illusionOfParadise: We need to return this info only for Jemengage Admin, but using a group `pap_campaign_read` will return this info also for app mobile.

src/Repository/Pap/AddressRepository.php
```diff
@@ -76,4 +77,22 @@ public function findNear(float $latitude, float $longitude, int $zoom, int $limi
 
         return $qb->getQuery()->getResult();
     }
+
+    public function countByPapCampaign(Campaign $campaign): int
```
elvino29, line 81: parameter seems unused

src/Repository/Pap/AddressRepository.php
```diff
@@ -76,4 +77,22 @@ public function findNear(float $latitude, float $longitude, int $zoom, int $limi
...
+    {
+        return (int) $this->createQueryBuilder('address')
+            ->select('COUNT(1)')
+            ->getQuery()
+            ->getSingleScalarResult()
+        ;
+    }
+
+    public function countVotersByPapCampaign(Campaign $campaign): int
```
elvino29, line 90: same here

PR #6768, opened 2022-01-17
[EM-1995] add some address information

elvino29:


features/api/pap_address.feature
```diff
@@ -163,6 +163,7 @@ Feature:
     ]
     """
 
+  @debug
```
illusionOfParadise, line 166: To remove.

PR #6771, opened 2022-01-17
[EM-1986] add enabled field on procuration request

elvino29:


src/Entity/ProcurationRequest.php
```diff
@@ -352,6 +352,13 @@ class ProcurationRequest
      */
     public $reachable = false;
 
+    /**
+     * @ORM\Column(type="boolean", options={"default": false})
+     *
+     * @Assert\Type("bool")
+     */
+    private bool $enabled = false;
```
ottaviano, line 360: with typehint `bool $enabled` we can avoid using `Assert\Type`, wdyt?

```suggestion
    /**
     * @ORM\Column(type="boolean", options={"default": false})
     */
    private bool $enabled = false;
```

ottaviano (approved):

PR #6776, opened 2022-01-18
[EM-2010] add endpoint to retrieve cgu and data policy static page

elvino29:


src/Entity/Page.php
```diff
@@ -2,12 +2,25 @@
...
+use ApiPlatform\Core\Annotation\ApiResource;
 use Doctrine\ORM\Mapping as ORM;
 use Gedmo\Mapping\Annotation as Gedmo;
 use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Component\Validator\Constraints as Assert;
 
 /**
+ * @ApiResource(
+ *     attributes={
+ *         "order": {"createdAt": "DESC"},
```
ottaviano, line 15: why do you use `order` if you don't have a list operation?

ottaviano (approved):

src/Entity/Page.php
elvino29: For later if we implement the Operations collections but i can remove it

PR #6781, opened 2022-01-19
[EM-2017] News scope visibility

Remg:


elvino29 (approved):

src/Entity/EntityScopeVisibilityTrait.php
```diff
@@ -22,7 +22,8 @@ trait EntityScopeVisibilityTrait
      *     "pap_campaign_read",
      *     "pap_campaign_read_after_write",
      *     "phoning_campaign_read",
-     *     "phoning_campaign_list"
+     *     "phoning_campaign_list",
+     *     "jecoute_news_read_dc"
```
illusionOfParadise, line 26: You can add `,` in the end, and if we add a new group we will see only one line as modified.
```suggestion
     *     "jecoute_news_read_dc",
```

src/Entity/EntityScopeVisibilityTrait.php
```diff
@@ -40,7 +41,9 @@ trait EntityScopeVisibilityTrait
      *     "pap_campaign_read_after_write",
      *     "phoning_campaign_read",
      *     "phoning_campaign_list",
-     *     "phoning_campaign_write"
+     *     "phoning_campaign_write",
+     *     "jecoute_news_read_dc",
+     *     "jecoute_news_write"
```
illusionOfParadise, line 46: Same here.

illusionOfParadise (approved):

PR #6814, opened 2022-01-27
fix migration

elvino29:


Remg (changes requested):
because it won't be executed again, you can not modify a migration.
create a new one instead.

arniverd (commented):
Ok pour moi

PR #6831, opened 2022-01-31
Remove delegated accesses from delegated users when delegator has no longer a delegator role

illusionOfParadise:


src/MyTeam/Listener/AdherentChangeListener.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $adherent = $event->getUser();
+
+        $teams = $this->myTeamRepository->findBy(['owner' => $adherent]);
+
+        foreach ($teams as $team) {
+            try {
+                $this->scopeGenerator->getGenerator($team->getScope(), $adherent);
+            } catch (NotFoundScopeGeneratorException $e) {
+                $delegatedAccesses = $this->delegatedAccessRepository->findBy([
```
ottaviano, line 50: what about using `removeFromDelegator` method here?

ottaviano (approved):

PR #6853, opened 2022-02-02
[LPN-120] add surveys to correspondent scope

elvino29:


src/Scope/ScopeEnum.php
```diff
@@ -39,6 +39,11 @@ class ScopeEnum extends Enum
         self::PHONING_NATIONAL_MANAGER,
     ];
 
+    public const LOCAL_SCOPES = [
```
illusionOfParadise, line 42: The name is confusing, because there are another local scopes not present in the constant.
What about `SURVEY_SCOPES`? It corresponds better to its use.

src/Api/Filter/JeMengageSurveyScopeFilter.php
```diff
@@ -49,6 +49,18 @@ protected function applyFilter(
                     ->setParameter('zones', $this->zoneRepository->findForJecouteByReferentTags($user->getManagedArea()->getTags()->toArray()))
                 ;
                 break;
+            case ScopeEnum::CORRESPONDENT:
+                $queryBuilder
+                    ->leftJoin(LocalSurvey::class, 'ls', Join::WITH, sprintf('ls.id = %s.id', $alias))
+                    ->orWhere((new Orx())
+                        ->add(sprintf('%s INSTANCE OF :national', $alias))
+                        ->add(sprintf('%s INSTANCE OF :local AND ls.zone = :zone', $alias))
```
illusionOfParadise, line 57: A correspondent managing Paris (as department) could not see surveys, because they are linked to a borough.
Should we check also children zones of managed zone?

src/Scope/ScopeEnum.php
elvino29: It is not just for the surveys that I added this constant. But yes, i can rename it to LOCAL_SURVEY_SCOPES if you want, but I think we'll need it very soon

illusionOfParadise: Not sure. We have `NATIONAL_SCOPES` and at the moment the scope that is not national is local. And any check for local/national scope can be done with.

ottaviano (approved):

PR #6854, opened 2022-02-02
[API][MyTeam] Take into account delegated access in spaces (AdherentMessages, News, Events, Teams, Phoning, PAP, Surveys)

illusionOfParadise:


src/Scope/FeatureEnum.php
```diff
@@ -45,25 +45,28 @@ class FeatureEnum extends Enum
         DelegatedAccess::ACCESS_EVENTS => self::EVENTS,
         DelegatedAccess::ACCESS_JECOUTE => self::MOBILE_APP,
         DelegatedAccess::ACCESS_DASHBOARD => self::DASHBOARD,
+        DelegatedAccess::ACCESS_MOBILE_APP => self::MOBILE_APP,
         DelegatedAccess::ACCESS_NEWS => self::NEWS,
         DelegatedAccess::ACCESS_ELECTIONS => self::ELECTIONS,
-        DelegatedAccess::ACCESS_RIPOSTES => self::RIPOSTES,
-        DelegatedAccess::ACCESS_PHONING => self::PHONING,
```
ottaviano, line 51: Should we remove these features from DB too?

src/Validator/Scope/ScopeVisibilityValidator.php
```diff
@@ -77,7 +77,9 @@ public function validate($value, Constraint $constraint): void
             return;
         }
 
-        if (!$this->managedZoneProvider->isManagerOfZone($this->getAdherent(), $scopeGenerator->getCode(), $value->getZone())) {
+        if (!$this->managedZoneProvider->zoneBelongsToSome($value->getZone(), array_map(
```
ottaviano, line 80: What about creating `zoneBelongsToSomeZomes($value->getZone(), $scope->getZones())`
and doing `array_map` on $zones inside? We will avoid duplication of `static function (Zone $zone) { return $zone->getId(); }, $scope->getZones())`

ottaviano (approved):

PR #6895, opened 2022-02-10
[API][PAP] Add a filter and order to CampaignHistory list + fix national campaign KPI for a local role

illusionOfParadise:


src/Entity/Pap/CampaignHistory.php
```diff
@@ -67,9 +68,11 @@
  *
  * @ApiFilter(SearchFilter::class, properties={
  *     "status": "exact",
+ *     "campaign.uuid": "exact"
  * })
  * @ApiFilter(AdherentIdentityFilter::class, properties={"questioner"})
  * @ApiFilter(DateFilter::class, properties={"createdAt", "beginAt"})
+ * @ApiFilter(OrderFilter::class, properties={"createdAt"})
```
elvino29, line 75: Why not already return a list sorted by creation date. That way there's no need to inject a filter.

illusionOfParadise: The order filter allow to change the order in the front, not only to set the order.

elvino29: ok

ottaviano (approved):

PR #6903, opened 2022-02-11
[Procuration] Update proxy form

ottaviano:


src/Entity/ProcurationProxy.php
```diff
@@ -77,9 +79,17 @@ class ProcurationProxy
     /**
      * @var string|null
      *
-     * @ORM\Column(length=6)
+     * @ORM\Column(nullable=true)
+     *
+     * @Assert\Length(max=255, groups={"front"})
+     */
+    private $voterNumber;
```
elvino29, line 86:
```suggestion
    private ?string $voterNumber = null;
```

same for the others

elvino29 (approved):

PR #6955, opened 2022-02-22
Store & compute last login group

Remg:


src/OAuth/AuthorizationServerFactory.php
```diff
@@ -50,6 +53,7 @@ public function __construct(
         $this->deviceRepository = $deviceRepository;
         $this->clientRepository = $clientRepository;
         $this->scopeRepository = $scopeRepository;
+        $this->entityManager = $entityManager;
```
ottaviano, line 56: seems to be unused

ottaviano (approved):

PR #6957, opened 2022-02-22
[Admin][News] Add 'pinned' and 'enriched' fields and make some other modifications

illusionOfParadise:
SQL to execute:
```sql
UPDATE jecoute_news 
SET link_label = 'Voir'
WHERE external_link IS NOT NULL
;
```

### DIFF
--- a/src/Admin/Jecoute/NewsAdmin.php
+++ b/src/Admin/Jecoute/NewsAdmin.php
@@ -3,6 +3,7 @@
 namespace App\Admin\Jecoute;
 
 use App\Admin\AbstractAdmin;
+use App\Admin\Filter\ZoneAutocompleteFilter;
 use App\Entity\Administrator;
 use App\Entity\Geo\Zone;
 use App\Entity\Jecoute\News;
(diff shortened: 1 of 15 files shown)


Remg (approved):

templates/admin/layout.html.twig
```diff
@@ -304,6 +304,105 @@
...
+                         title: "Gras"
+                     },
+                     {
+                         name: "italic",
+                         action: SimpleMDE.toggleItalic,
+                         className: "fa fa-italic",
+                         title: "Italique"
+                     },
+                     {
+                         name: "heading-1	",
```
ottaviano, line 345:
```suggestion
                         name: "heading-1",
```

ottaviano (approved):

src/Admin/Jecoute/NewsAdmin.php
```diff
@@ -205,6 +301,7 @@ public function prePersist($object)
     public function postPersist($object)
     {
         $this->newsHandler->handleNotification($object);
+        $this->newsHandler->changePinned($object);
```
ottaviano, line 304: I think we should call this on `postUpdate` too

PR #6960, opened 2022-02-23
Synchronize nullable zones

Remg:


src/Mailchimp/Synchronisation/RequestBuilder.php
```diff
@@ -458,18 +458,10 @@ private function buildMergeFields(): array
             $mergeFields[MemberRequest::MERGE_FIELD_MUNICIPAL_TEAM] = (string) $this->takenForCity;
         }
 
-        if ($this->zoneBorough) {
-            $mergeFields[MemberRequest::getMergeCodeFieldFromZone($this->zoneBorough)] = (string) $this->zoneBorough;
-        }
-
         if ($this->zoneCity) {
```
ottaviano, line 461: why don't we do the same for zoneCity and zoneDepartment fields?

ottaviano (approved):

PR #7054, opened 2022-03-15
Add DynamicLinks generator

ottaviano:


src/Entity/Pap/Campaign.php
```diff
@@ -541,4 +547,19 @@ public function setAssociated(bool $value): void
     {
         $this->associated = $value;
     }
+
+    public function getDynamicLinkPath(): string
+    {
+        return '/pap-campagnes/'.$this->uuid;
```
elvino29, line 553:
```suggestion
        return '/pap-campaigns/'.$this->uuid;
```

Remg (approved):

elvino29 (approved):

src/Entity/Pap/Campaign.php
ottaviano: yes, you are absolutely right. Fixed 👍

PR #7134, opened 2022-04-05
[Assessors][LPN-610] update FranceCitiesBundle

elvino29:


ottaviano (approved):
we should update user data (events, committees, etc...) after deployment for being coherent with these new codes.

PR #7259, opened 2022-05-02
[Procuration] Disable/enable a procuration request and list disabled requests

illusionOfParadise:


src/Procuration/Filter/ProcurationRequestFilters.php
```diff
@@ -34,16 +40,31 @@ public function apply(QueryBuilder $qb, string $alias): void
...
+
+                break;
+            case self::UNPROCESSED:
+                $qb
+                    ->andWhere("$alias.processed = :flag AND $alias.processedAt IS NULL")
+                    ->setParameter('enabled', true)
+                    ->setParameter('flag', 0)
+                ;
+
+                break;
```
ottaviano:
```suggestion
        $qb
            ->andWhere("$alias.enabled = :enabled")
            ->setParameter('enabled', true)
        ;

        switch ($this->getStatus()) {
            case self::PROCESSED:
                $qb
                    ->andWhere("$alias.processed = :flag AND $alias.processedAt IS NOT NULL")
                    ->setParameter('flag', 1)
                ;

                break;
            case self::UNPROCESSED:
                $qb
                    ->andWhere("$alias.processed = :flag AND $alias.processedAt IS NULL")
                    ->setParameter('flag', 0)
                ;

                break;
```

ottaviano (approved):

PR #7279, opened 2022-05-10
[Procuration] Apply proxies count for every election round and change calculation of availability

illusionOfParadise:
After merge execute `app:procuration:process-availabilities` command.


ottaviano (approved):
👍

PR #7357, opened 2022-06-16
[VotingPlatform] Add new poll type

ottaviano:


translations/messages.fr.yml
```diff
@@ -909,6 +910,7 @@ voting_platform.vote.majority_vote_mention.good: Bien
 voting_platform.vote.majority_vote_mention.fair: Passable
 voting_platform.vote.majority_vote_mention.insufficient: Insuffisant
 voting_platform.pool_title.committee_supervisor: Animateurs Locaux
+voting_platform.pool_title.poll:
```
elvino29, line 913: No translation here?

ottaviano: No, it should be empty

PR #7364, opened 2022-06-24
[VoteStatuses] update style of voting platform

ottaviano:


templates/voting_platform/finish.html.twig
```diff
@@ -35,6 +35,8 @@
             <h1 class="text--larger b__nudge--bottom-large">
                 {% if designation.isExecutiveOfficeType() %}
                     Félicitations, votre bulletin est dans l'urne !
+                {% elseif designation.isPollType() %}
+                    Félicitations, votre vote est bien enregistré !
                 {% else %}
                     Félicitations, vos bulletins sont dans l'urne !
```
elvino29, line 41: Same text with first condition

elvino29 (approved):

templates/voting_platform/finish.html.twig
ottaviano: not exactly :D 
```
Félicitations, votre bulletin est dans l'urne !
Félicitations, votre vote est bien enregistré !
```

PR #7493, opened 2022-09-07
[Renaissance] allow connected user to fill adhesion form

ottaviano:


src/Form/Renaissance/Adhesion/MembershipRequestMentionsType.php
```diff
@@ -15,7 +15,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
             ->add('conditions', CheckboxType::class, [
-                'required' => false,
+                'required' => true,
```
elvino29, line 18: Default value is true, no need to add required

elvino29 (approved):

PR #7510, opened 2022-09-13
Add Additional informations step in funnel

Remg:


src/Entity/Adherent.php
```diff
@@ -671,6 +671,11 @@ class Adherent implements UserInterface, UserEntityInterface, GeoPointInterface,
      */
     private $certifiedAt;
 
+    /**
+     * @var
+     */
+    private $socioProfessionalCategory;
```
elvino29, line 677: seems never used

templates/renaissance/adhesion/additional_informations.html.twig
```diff
@@ -0,0 +1,37 @@
+{% extends 'renaissance/adhesion/_adhesion_layout.html.twig' %}
+
+{% form_theme form 'renaissance/form_theme.html.twig' %}
+
+{% block final_javascripts %}
```
elvino29, line 5: why the captcha js block here? add the active step block instead

elvino29 (changes requested):

elvino29 (approved):

PR #7761, opened 2022-11-14
Remove privacy card

mckenziearts:


ottaviano (commented):
As discussed, could you please rollback entirely this file `templates/renaissance/adherent/_layout.html.twig` and keep only `templates/renaissance/adherent/profile/form.html.twig`

PR #7762, opened 2022-11-14
[Profile] Security & RE emails

Remg:
- +2 RE users in fixtures
- field "mandates" in RE profile type is now expanded
- new RE messages for CertificationRequests & change email validation
- new voter for profil pages (RE users can not edit profile on EM & viceversa)

### DIFF
--- a/src/Mailer/Message/Renaissance/RenaissanceAdherentChangeEmailMessage.php
+++ b/src/Mailer/Message/Renaissance/RenaissanceAdherentChangeEmailMessage.php
@@ -3,10 +3,9 @@
 namespace App\Mailer\Message\Renaissance;
 
 use App\Entity\Adherent;
-use App\Mailer\Message\Message;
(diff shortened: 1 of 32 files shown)


src/Mailer/Message/Renaissance/RenaissanceAdherentChangeEmailMessage.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace App\Mailer\Message\Renaissance;
+
+use App\Entity\Adherent;
+use App\Mailer\Message\Message;
+use Ramsey\Uuid\Uuid;
+
+final class RenaissanceAdherentChangeEmailMessage extends Message
```
ottaviano, line 9: should be extend `AbstractRenaissanceMessage` too?

ottaviano (approved):
GG

PR #7763, opened 2022-11-14
[Profile] remove captcha and fix JS

ottaviano:


templates/renaissance/adherent/_layout.html.twig
```diff
@@ -17,7 +17,7 @@
                         </button>
                         <!-- Logo section -->
                         <div class="flex items-center px-4 lg:px-0">
-                            <a href="{{ path('app_user_edit', { app_domain: renaissance_host }) }}" class="shrink-0">
+                            <a href="{{ path('app_renaissance_homepage') }}" class="shrink-0">
```
Remg, line 20: you need to remove the following link then https://github.com/EnMarche/en-marche.fr/blob/a053e12963904a73697fd1eec2139848b7dcab95/templates/renaissance/adherent/_layout.html.twig#L51

ottaviano: I rollback the change on logo link

PR #7833, opened 2022-11-24
Add account validation funnel after RE adherent creation

Remg:


src/Entity/Adherent.php
```diff
@@ -901,6 +901,7 @@ public static function createBlank(
         $adherent->territoireProgresMembership = $territoiresProgresMembership;
         $adherent->agirMembership = $agirMembership;
         $adherent->registeredAt = new \DateTime('now');
+        $adherent->referentTags = new ArrayCollection();
```
ottaviano, line 904: what about moving this into construct?

templates/renaissance/adherent/creation_confirmation.html.twig
```diff
@@ -0,0 +1,19 @@
+{% extends 'renaissance/double_columns_layout.html.twig' %}
+
+{% block title 'Connectez-vous avec vos identifiants' %}
+
+{% block back_btn '' %}
```
ottaviano, line 5: it seems to be unused

ottaviano (approved):

PR #7905, opened 2022-12-06
[LPN-1287] update article preview

elvino29:


templates/admin/article/list_preview.html.twig
```diff
@@ -1,15 +1,13 @@
...
-        </a>
-    {% else %}
-        <a href="{{ path('article_view', {categorySlug: object.category.slug, articleSlug: object.slug}) }}" class="btn btn-sm btn-default view_link"
-           title="Prévisualiser" target="_blank" rel="noopener noreferrer">
-            <i class="fa fa-search-plus" aria-hidden="true"></i>
-            Prévisualiser
-        </a>
-    {% endif %}
+{% if object.isForRenaissance %}
+    <a href="{{ path('article_view', {categorySlug: object.category.slug, articleSlug: object.slug, app_domain: renaissance_host, preview: article_preview_admin_key}) }}" class="btn btn-sm btn-default view_link"
```
ottaviano, line 2: preview key should be set only for unpublished articles

ottaviano (approved):

templates/admin/article/list_preview.html.twig
elvino29: ok

PR #7943, opened 2022-12-09
Remove "La République Ensemble" link in footer

tony-tran:
Hi,

The "La République Ensemble" link is KO then i propose to remove it.


ottaviano (approved):
Nice, thank you, Tony !

PR #7961, opened 2022-12-14
[LPN-1307][LocalElection] add some filters and add new panel on edit form

elvino29:


src/Admin/LocalElection/LocalElectionAdmin.php
```diff
@@ -73,6 +81,15 @@ protected function configureFormFields(FormMapper $form): void
                 ])
             ->end()
         ;
+
+        if (!$this->isCurrentRoute('create')) {
+            $form
+                ->with('Listes', ['class' => 'col-md-6'])
+                    ->add('candidaciesGroups', null, [
+                        'label' => false,
+                    ])
```
ottaviano, line 90: `->end()` call is missing here for `with()`

PR #8135, opened 2023-01-24
[1378] create elected representative endpoints for JME

elvino29:


ottaviano (approved):

ottaviano:
IMO we should keep SurveyQuestionFromIdDenormalizer and make custom `PoliticalFunctionFromIdDenormalizer`

PR #8179, opened 2023-02-10
Add admin documents

Remg:


src/Documents/DocumentHandler.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        if (!$file = $document->getFile()) {
+            return;
+        }
+
+        $this->removeFile($document);
+
+        $document->setFilePath($path = sprintf(
+            '%s/%s.%s',
+            'files/documents/admin',
```
ottaviano, line 25: IMO we don't need this detail in the file path, if we open this feature to simple user (not admin) we don't have to change this path.

```suggestion
            'files/documents',
```

src/Entity/Document.php
```diff
@@ -0,0 +1,170 @@
...
+ *             "controller": "App\Controller\Api\DocumentDownloadFileController",
+ *             "requirements": {"uuid": "%pattern_uuid%"},
+ *             "security": "is_granted('ROLE_OAUTH_SCOPE_JEMENGAGE_ADMIN') and is_granted('IS_FEATURE_GRANTED', 'documents')",
+ *         }
+ *     }
+ * )
+ *
+ * @ApiFilter(SearchFilter::class, properties={
+ *     "title": "partial",
+ *     "visibility": "exact",
```
ottaviano, line 50: it seems that the property `visibility` is missing?

src/Entity/Document.php
```diff
@@ -0,0 +1,170 @@
...
+ * })
+ *
+ * @ORM\Entity(repositoryClass="App\Repository\DocumentRepository");
+ * @ORM\Table(name="document")
```
ottaviano, line 54: we can avoid `name` attribut

```suggestion
 * @ORM\Table()
```

src/Entity/Document.php
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * @UniqueEntity(fields={"title"}, message="document.title.unique_entity")
```
ottaviano, line 56: we can add `unique` db constraint too

src/Entity/Document.php
```diff
@@ -0,0 +1,170 @@
...
+
+    public function setFile(?UploadedFile $file): void
+    {
+        $this->file = $file;
+    }
+
+    public function getFilePath(): ?string
+    {
+        return $this->filePath;
+    }
```
ottaviano, line 159: what about removing all this getter/setter and make public properties with old good typehint ? I know it's crazy idea, but ...

src/Normalizer/DocumentNormalizer.php
```diff
@@ -0,0 +1,70 @@
...
+            $parameters['scope'] = $scope->getCode();
+        }
+
+        return $this->urlGenerator->generate('api_documents_get_file_item', $parameters, UrlGeneratorInterface::ABSOLUTE_URL);
+    }
+
+    private function getCurrentScope(): ?ScopeGeneratorInterface
+    {
+        if (!$this->currentScope) {
+            $this->currentScope = $this->scopeGeneratorResolver->resolve();
```
ottaviano, line 65: I think we can improve this local cache system. What about implement it into ScopeGeneratorResolver instead, into `resolve` method?

ottaviano (approved):

src/Scope/FeatureEnum.php
```diff
@@ -46,6 +47,7 @@ class FeatureEnum extends Enum
         self::ADHERENT_FORMATIONS,
         self::COMMITTEE,
         self::GENERAL_MEETING_REPORTS,
+        self::DOCUMENTS,
```
ottaviano, line 50: could you add a translation for this new feature [here](https://github.com/EnMarche/en-marche.fr/blob/9aa3d1f3465f19974ee8f14578b04f57740e2d54/translations/messages+intl-icu.fr.yml#L1356), please?

ottaviano (approved):

PR #8255, opened 2023-03-02
[Admin] Fix mandate geo zone autocomplete

Remg:


src/Admin/ElectedRepresentative/MandateAdmin.php
```diff
@@ -72,9 +73,11 @@ protected function configureFormFields(FormMapper $formMapper): void
                 'error_bubbling' => true,
                 'attr' => ['class' => 'width-140'],
             ])
-            ->add('geoZone', ZoneType::class, [
```
ottaviano, line 75: could you remove the class ZoneType, it seems to be not used anymore

ottaviano (approved):

PR #8306, opened 2023-03-13
[LPN-1381] add new menu for adherent creation

elvino29:


src/Renaissance/Membership/Admin/AdherentCreateCommand.php
```diff
@@ -152,6 +149,9 @@ class AdherentCreateCommand implements MembershipInterface
      */
     public \DateTime $cotisationDate;
 
+    private ?int $adherentId = null;
```
ottaviano, line 152: it seems to be unused

ottaviano (approved):

PR #8411, opened 2023-03-30
[LPN-1618] adherent readhesion process

elvino29:


src/Donation/TransactionSubscriber.php
```diff
@@ -81,7 +81,13 @@ public function onPayboxIpnResponse(PayboxResponseEvent $event): void
...
             if ($donation->isMembership()) {
-                if ($adherent = $donation->getDonator()->getAdherent()) {
+                $donator = $donation->getDonator();
+                if ($adherent = $donator->getAdherent()) {
+                    $membershipDonations = $donator->getMembershipDonations();
+                    if ($membershipDonations->count() > 1) {
+                        $this->membershipRequestHandler->finishRenaissanceReAdhesion($adherent);
+                    }
+
                     $this->membershipRequestHandler->finishRenaissanceAdhesion($adherent);
```
ottaviano, line 91: the adherent will receive two emails here, 1 from finishRenaissanceReAdhesion and another from finishRenaissanceAdhesion, normal?

elvino29: no, missing `else` condition

PR #8673, opened 2023-05-25
[RE-1749] add new adherent tag on projection entity

elvino29:


src/Entity/Projection/ManagedUser.php
```diff
@@ -315,6 +315,15 @@ class ManagedUser
...
     private ?UuidInterface $committeeUuid;
 
+    /**
+     * @var string[]|null
+     *
+     * @ORM\Column(type="simple_array", nullable=true)
+     *
+     * @Groups({"managed_user_read"})
+     */
+    private $tags;
```
ottaviano, line 325:
```suggestion
    private ?array $tags = null;
```

ottaviano (approved):

PR #8893, opened 2023-07-13
[RE-1952] change the incrementation of adherent formation print

elvino29:


src/Entity/AdherentFormation/Formation.php
```diff
@@ -221,9 +224,18 @@ class Formation implements EntityScopeVisibilityWithZoneInterface, EntityAdheren
      */
     private bool $valid = false;
 
+    /**
+     * @var Collection|Adherent[]
+     *
+     * @ORM\ManyToMany(targetEntity="App\Entity\Adherent")
+     * @ORM\JoinTable(name="adherent_formation_print_by_adherents")
```
ottaviano, line 231: could you add `onDelete="CASCADE"` + `fetch="EXTRA_LAZY"`, please?

src/Entity/AdherentFormation/Formation.php
```diff
@@ -335,4 +347,36 @@ public function incrementPrintCount(): void
...
+    {
+        $this->printByAdherents = $printByAdherents;
+    }
+
+    public function addPrintByAdherent(Adherent $adherent): void
+    {
+        if (!$this->printByAdherents->contains($adherent)) {
+            $this->printByAdherents->add($adherent);
+        }
+    }
```
ottaviano, line 369: we can increment the count on `addPrintByAdherent` method and return `true` or `false` for flushing
```suggestion
    public function addPrintByAdherent(Adherent $adherent): bool
    {
        if (!$this->printByAdherents->contains($adherent)) {
            $this->printByAdherents->add($adherent);
            $this->incrementPrintCount();
            
            return true;
        }
        
        return false;
    }
```

src/Entity/AdherentFormation/Formation.php
```diff
@@ -335,4 +347,36 @@ public function incrementPrintCount(): void
...
+
+    public function removePrintByAdherent(Adherent $adherent): void
+    {
+        if ($this->printByAdherents->contains($adherent)) {
+            $this->printByAdherents->removeElement($adherent);
+        }
+    }
```
ottaviano, line 376: I think we can remove this method as not used

src/Controller/Renaissance/Formation/ContentController.php
```diff
@@ -28,8 +29,12 @@ public function __construct(
 
     public function __invoke(Formation $formation): Response
     {
-        $formation->incrementPrintCount();
-        $this->entityManager->flush();
+        /** @var Adherent $adherent */
+        $adherent = $this->getUser();
+
+        if (!$formation->hasPrintByAdherent($adherent) && $formation->addPrintByAdherent($adherent)) {
```
ottaviano, line 35: we can simplify this by removing `$formation->hasPrintByAdherent` call as we already do this check in `addPrintByAdherent` method.

```php
if ($formation->addPrintByAdherent($adherent)) {
```

PR #8896, opened 2023-07-13
[RE-1975] remove LRE space

elvino29:


src/Entity/AdherentCharter/LreCharter.php
```diff
@@ -1,12 +0,0 @@
-<?php
-
-namespace App\Entity\AdherentCharter;
-
-use Doctrine\ORM\Mapping as ORM;
-
-/**
- * @ORM\Entity
- */
-class LreCharter extends AbstractAdherentCharter
```
ottaviano, line 10: could you add a SQL query into migration for removing all rows for this type of charter from `adherent_charter` table?

same for `LreManagerElectedRepresentativeFilter` and `LreManagerElectedRepresentativeMessage` entities.

ottaviano (approved):

src/Entity/AdherentCharter/LreCharter.php
elvino29: ok

PR #9656, opened 2024-01-22
Chatbot v2

Remg:


src/Telegram/WebhookRequestHandler.php
```diff
@@ -0,0 +1,58 @@
...
+
+        try {
+            $data = BotApi::jsonValidate($content, true);
+        } catch (InvalidJsonException $e) {
+            throw new BadRequestHttpException('The request content is not a valid JSON payload.');
+        }
+
+        try {
+            $update = Update::fromResponse($data);
+        } catch (InvalidArgumentException $e) {
```
ottaviano, line 48: IMO all this stuff can be moved into TelegramBotController because it's too related to the HTTP request. Actually I can't use this handler from Sf Command but I can use UpdateHandler. Wdyt?

translations/messages+intl-icu.fr.yml
```diff
@@ -1480,3 +1480,9 @@ administrator:
             communautes_thematiques: Communautés thématiques
             elections_departementales: Élections départementales
             archives: Archives
+
+chatbot:
+    type:
+        telegram: Telegram
+    assistant_type:
+        openai: OpenAI
```
ottaviano, line 1488: what about making it in two lines ?

```
chatbot.type.telegram:
chatbot.assistant_type.openai:
```

it's easier to search when it's in one line

PR #9705, opened 2024-02-01
[Admin] Refactor adherents filter

Remg:


src/Admin/Filter/AdherentTagFilter.php
```diff
@@ -0,0 +1,40 @@
...
+                $orX = $qb->expr()->orX();
+
+                $condition = match ($value->getType()) {
+                    ContainsOperatorType::TYPE_NOT_CONTAINS => 'NOT LIKE',
+                    default => 'LIKE',
+                };
+
+                foreach ($value->getValue() as $index => $choice) {
+                    $orX->add($alias.'.tags '.$condition.' :tag_'.$index);
+                    $qb->setParameter('tag_'.$index, '%'.$choice.'%');
```
ottaviano, line 31: as we can use this filter many times on the same Admin we need customize `:tag_1` query params to be unique in the query

ottaviano (approved):

PR #10150, opened 2024-04-27
[BDE] create donation form page

ottaviano:


victorcohenfr:
Task linked: [RE-1898 Mettre en place le funnel de don  BDE](https://app.clickup.com/t/2643737/RE-1898)

PR #11089, opened 2024-11-28
[Event images] Retrieve old images informations

Remg:


src/Entity/AdvancedImageTrait.php
```diff
@@ -12,10 +12,10 @@ trait AdvancedImageTrait
     }
 
     #[ORM\Column(type: 'bigint', nullable: true)]
-    protected $imageSize;
+    protected ?int $imageSize;
```
ottaviano, line 15:
```suggestion
    protected ?int $imageSize = null;
```

src/Entity/AdvancedImageTrait.php
```diff
@@ -12,10 +12,10 @@ trait AdvancedImageTrait
...
 
     #[ORM\Column(length: 50, nullable: true)]
-    protected $imageMimeType;
+    protected ?string $imageMimeType;
```
ottaviano, line 18:
```suggestion
    protected ?string $imageMimeType = null;
```

ottaviano (approved):

PR #11915, opened 2025-06-02
[Donation] Send email for first successful transaction only

Remg:


src/Entity/Donation.php
```diff
@@ -636,4 +636,18 @@ public function isSuccess(): bool
...
+        if (!$transaction->isSuccessful()) {
+            return false;
+        }
+
+        $successfulTransactions = array_filter(
+            $this->transactions->toArray(),
+            fn (Transaction $t) => $t->isSuccessful()
+        );
+
+        return 1 === \count($successfulTransactions) && $this->transactions->first() === $transaction;
```
ottaviano, line 651: we should compare with the first transaction of `$successfulTransactions`, no ?

```suggestion
        return 1 === \count($successfulTransactions) && current($successfulTransactions) === $transaction;
```

